PR #4, opened 2025-06-12
fix [setup]: build with `--no-build-isolation` to find and use the correct pytorch version

ivanium:
To solve the problem of  ```ImportError: /data/yifanqiao/code/kvcached/kvcached/vmm_ops.cpython-311-x86_64-linux-gnu.so: undefined symbol: _ZNK2at10TensorBase4nameB5cxx11Ev```.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,6 @@ This will compile and install kvcached. If you have the right versions of vLLM a
 
 NOTE: `--no-build-isolation` is required for kvcached to find the right PyTorch in the current virtual environment.
 
-
 ## Testing
 
 To verify the successful installation and benchmark the performance of vLLM/SGLang with kvcached, run:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -28,24 +28,12 @@ This script will download the specified versions of vLLM and SGLang, create sepa
...
-```txt
-ImportError: /data/yifanqiao/code/kvcached/kvcached/vmm_ops.cpython-311-x86_64-linux-gnu.so: undefined symbol: _ZNK2at10TensorBase4nameB5cxx11Ev
-```
-
-To resolve this issue, one can rebuild the `vmm_ops` library locally by running:
-
-```python
-python setup.py build_ext --inplace
-```
+NOTE: `--no-build-isolation` is required for kvcached to find the right PyTorch in the current virtual environment.
```
jiarong0907, line 37: We might still want to keep the method of manual build in case some people need it. We can remove the error related information.

jiarong0907 (approved):
LGTM

Issue #9, opened 2025-06-19
Minor issue regarding a warning logging


nanomaoli:
Relevant line: [line 42 in kvcached/kvcached/controller/cli.py](https://github.com/ovg-project/kvcached/blob/a403172ee33a1138b2fdb284c40b720bad39f79d/kvcached/controller/cli.py#L42)

```python
        with RwLockedShm(get_ipc_name(ipc_name),
                         np.int64().itemsize * 2, RwLockedShm.WLOCK) as mm:
            mem_info = np.ndarray((2, ), dtype=np.int64, buffer=mm)
            delta = kv_cache_limit - mem_info[0]
            if delta < 0:
                if mem_info[1] + delta < 0:
                    logger.warning(
                        f"kv_cache_limit is less than in_use for {ipc_name}")
            mem_info[0] = kv_cache_limit
            return mem_info
```

### Issue description
This block appears to check whether the new kv_cache_limit is smaller than the currently in-use KV cache (mem_info[1]). However, the condition:
```python
if mem_info[1] + delta < 0
```
compares the *change in limit* (delta) against the *current in_use memory* (mem_info[1]), rather than directly checking if kv_cache_limit < mem_info[1].

This may not cause any functional errors, but could lead to incorrect logging behavior.

### Suggestion
If the intention is to emit a warning when the new kv_cache_limit is below the in-use memory, it may be:
```python
if kv_cache_limit < mem_info[1]:
```

jiarong0907:
Thanks for identifying this issue. I think it wants to check whether there is a room to decrease. If the left space is smaller than the space we want to decrease, it will raise a warning.

```Python
if mem_info[0] - mem_info[1] + delta < 0:
    logger.warning(f"kv_cache_limit is less than the free space for {ipc_name}")
```

Does this make sense?

jiarong0907:
@nanomaoli can you send a PR to fix it?

PR #10, opened 2025-06-20
[Bugfix] fix a warning condition for update_kv_cache_limit()

nanomaoli:
- Fixes a logging condition that previously was `mem_info[1] + delta < 0`.
- Replaced with a more accurate condition: `mem_info[0] - mem_info[1] + delta < 0`.
- This ensures that the warning is logged only when the decreased kv_cache_limit is smaller than current kv cache volume.


jiarong0907:
Resolves #9

PR #13, opened 2025-07-08
[WIP][Feat] TP support for kvcached

nanomaoli:
Dear all,

To use this proposed tensor parallel support, please add a small modification below.
At `kvcached/engine_integration/vllm-v0.8.4/vllm/v1/worker/gpu_model_runner.py`, Line **280** if you did not modify nearby code, or specifically, the init_kvcached() call, the original code is:
```
 kvcached_ops.init_kvcached()
```
Here to pass in to kvcached the tensor_parallel_size value, I modified it as:
```
 kvcached_ops.init_kvcached(parallel_config.tensor_parallel_size)
```
This is a minor edit, and I found this may be more convenient for your use, so I did not update the patch.

**tests and known issues**
I had test runs with our ./start_server.sh and ./start_client.sh scripts on both vLLM and SGLang, with tensor parallel size 4, 2, and 1(no tp).
All above runs finished successfully. However, I saw a warning message when running SGLang on all the 3 tensor parallel sizes.
The message looks like:

> /users/8/li003703/projects/ovg-project/kvcached/engine_integration/sglang-v0.4.6.post2/python/sglang/srt/managers/scheduler.py:1245: UserWarning: token_to_kv_pool_allocator memory leak detected! available_size=464896.0, protected_size=0, self.max_total_num_tokens=500925                                                                                                                                                                  
self.token_to_kv_pool_allocator.available_size()=464896.0                                                                                                                                                               
self.tree_cache.evictable_size()=0

but the GPU memory utilization, either right after server is ready or after running a ./start_client job, are similar to those of vLLM cases. (Maybe irrelevant but I wanted to mention that, for both vLLM and SGLang, I can see an increase in GPU memory usage after finishing a ./start_client job. Not sure if this is some caching behavior or not.)

Thanks
Hao

### DIFF
--- a/engine_integration/scripts/kvcached-vllm-v0.9.2.patch
+++ b/engine_integration/scripts/kvcached-vllm-v0.9.2.patch
@@ -1,29 +1,12 @@
 diff --git a/vllm/v1/core/block_pool.py b/vllm/v1/core/block_pool.py
-index d21f94727..cdb7abb95 100644
+index d21f94727..173f67963 100644
 --- a/vllm/v1/core/block_pool.py
 +++ b/vllm/v1/core/block_pool.py
-@@ -72,7 +72,7 @@ class BlockPool:
-     def get_cached_block(
-             self, block_hash: BlockHash,
-             kv_cache_group_ids: list[int]) -> Optional[list[KVCacheBlock]]:
--        """Get the cached block by the block hash for each group in 
-+        """Get the cached block by the block hash for each group in
-         `kv_cache_group_ids`, or None if cache miss for any group.
-         If there are duplicated blocks, we return the first block in the cache.
- 
-@@ -338,7 +338,7 @@ class BlockPool:
- 
-     def take_events(self) -> list[KVCacheEvent]:
-         """Atomically takes all events and clears the queue.
--        
-+
-         Returns:
-             A list of KV cache events.
-         """
-@@ -347,3 +347,105 @@ class BlockPool:
+@@ -347,3 +347,112 @@ class BlockPool:
          events = self.kv_event_queue
          self.kv_event_queue = []
          return events
++
 +
 +class ElasticBlockPool(BlockPool):
 +    """ElasticBlockPool that manages KVCacheBlocks.
@@ -31,8 +14,13 @@ index d21f94727..cdb7abb95 100644
 +    elastic KV cachememory management.
 +    """
 +
-+    def __init__(self, num_gpu_blocks: int, block_size: int, cell_size: int,
-+                 num_layers: int, enable_caching: bool, enable_kv_cache_events: bool = False):
++    def __init__(self,
++                 num_gpu_blocks: int,
++                 block_size: int,
++                 cell_size: int,
++                 num_layers: int,
++                 enable_caching: bool,
++                 enable_kv_cache_events: bool = False):
 +        assert isinstance(num_gpu_blocks, int) and num_gpu_blocks > 0
 +        assert not enable_caching, (
 +            "Caching is not supported in ElasticBlockPool")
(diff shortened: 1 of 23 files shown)


jiarong0907:
Hi @nanomaoli , did you make any changes to the vllm code? If so, please also commit the updated vllm patch.

engine_integration/scripts/kvcached-vllm-v0.9.2.patch
```diff
@@ -249,10 +182,20 @@ index e2fdf6f8a..778b56598 100644
...
 --- a/vllm/v1/worker/gpu_model_runner.py
 +++ b/vllm/v1/worker/gpu_model_runner.py
-@@ -311,6 +311,24 @@ class GPUModelRunner(LoRAModelRunnerMixin):
+@@ -51,7 +51,8 @@ from vllm.v1.attention.backends.utils import (AttentionMetadataBuilder,
+                                               CommonAttentionMetadata)
+ from vllm.v1.core.encoder_cache_manager import compute_encoder_budget
+ from vllm.v1.kv_cache_interface import (AttentionSpec, FullAttentionSpec,
+-                                        KVCacheConfig, KVCacheSpec, MambaSpec,
++                                        KVCacheConfig, KVCacheSpec,
++                                        KVCacheTensor, MambaSpec,
```
ivanium, line 194: Can we import KVCacheTensor inside the function body or shall we just annotate the type from `KVCacheTensor` to `"KVCacheTensor"`?

ivanium (commented):
Left a minor comment on import. Otherwise LGTM

kvcached/kv_cache_manager.py
```diff
@@ -152,6 +157,24 @@ def __init__(self, total_mem_size: int, page_size: int):
             # Start preallocation thread
             self._start_prealloc_thread()
 
+        self.tp_size = tp_size
+
+    def call_map_to_kv_tensors(self, offsets: list[int]) -> None:
+        """Call the mapping function to map pages to physical memory, considering tensor parallelism."""
+        if self.tp_size is not None and self.tp_size > 1:
```
ivanium: Do we still need to check `self.tp_size is not None`? If we do have to, we may need to update the type annotation to `Optional[int]`

nanomaoli: I think having self.tp_size > 1 is good enough now, we could just let the default value be 1 at PageAllocator's __init__().

nanomaoli:
I tested running these 16 settings: 

- {vLLM, SGLang} x (tp_size = 1, 2, 4, 8) x {ENABLE_KVCACHED = true, false} 

I used 8 A40 GPUs.

For each setting above, I run with NUM_REQUEST=1000, REQUEST_RATE = {10, 100, 1000}. All these runs have all 1000 requests successful.

For each ENABLE_KVCACHED=true setting above, I run with NUM_REQUEST=50, REQUEST_RATE = 10, and dump the model responses. I compare the reponses with ENABLE_KVCACHED = false settings, which shows that the responses are correct.

nanomaoli:
Some changes in the vLLM patch can be cleaned, and start_client.sh script needs to set the version accordingly. ------ these are already fixed in two recent commits "using latest vllm in client.sh" and "fix_patch".

In kvcached/kv_cache_manager.py's PageAllocator, now I think we don't need to check `self.tp_size is not None` ------ Dr. Yifan already made a comment.

engine_integration/scripts/kvcached-vllm-v0.9.2.patch
jiarong0907: Tried this but it cannot pass the pre-commit type check.

kvcached/kv_cache_manager.py
jiarong0907: Good catch. Let me fix.

jiarong0907:
> There is a warning, when using kvcached with SGLang. It is like:
> 
> > warnings.warn(msg)
> > /users/8/li003703/projects/ovg-project/kvcached/engine_integration/sglang-v0.4.6.post2/python/sglang/srt/managers/scheduler.py:1245: UserWarning: token_to_kv_pool_allocator memory leak detected!
> > available_size=2179072.0, protected_size=0, self.max_total_num_tokens=2198282
> > self.token_to_kv_pool_allocator.available_size()=2179072.0
> > self.tree_cache.evictable_size()=0
> 
> If not enabling kvcached, this warning doesn't exist. It doesn't fail the program, and seems to not cause any significant difference.

I remember in the SGLang PR, shan or Yifan has some way to disable this warning. I will add it to our patch in a different PR.

PR #14, opened 2025-07-09
[Chore] Change print to logger

shanyu-sys:


jiarong0907:
Thanks for enhancing this! As I tested, the logger only works with sglang but not vllm. Probabaly because vLLM doesn't configure the root logger, so kvcached logger inherits the default Python logging level (WARNING), which filters out INFO messages. 

Probably considering define our own kvcached logger?

jiarong0907 (approved):
Added customized logger for kvcached. The logging_config can be merged into util after code refactoring.

PR #18, opened 2025-07-11
chore [engine_integration]: separate installation for vllm and sglang

ivanium:
This PR added an option to the engine integration setup script to selectively install VLLM, SGLang, or both.


jiarong0907 (approved):
LGTM!

PR #20, opened 2025-07-13
fix: misc minor fixes after replacing prints with python logging module

ivanium:
1. Fix some mismatches after replacing prints with the Python logging module. 
2. Add a util function to allow setting the debug level with an env var.


jiarong0907 (approved):
LGTM. Thanks!

PR #22, opened 2025-07-13
fix [vllm-patch]: add init_kvcached() in vllm scheduler

### LINKED ISSUE #21: [Bug] vLLM will fail because of the check in get_kv_cache_manager
vLLM will crash during initialization because of the failure of this check.

init_kvcached will only be called by worker processes. The main process will not call it and set the `_kvcached_initialized` flag, causing the failure of this check. It works for TP=1 because the worker share the same process with main.

```
def get_kv_cache_manager(num_blocks: int, block_size: int, cell_size: int,
                         num_layers: int) -> KVCacheManager:
    if not _kvcached_initialized:
        raise RuntimeError(
            "kvcached is not initialized. Please call init_kvcached() first.")

    return KVCacheManager(num_blocks, block_size, cell_size, num_layers)
```

ivanium:
To ensure the VLLM scheduler can successfully get_kv_cache_manager().

This is not happening now with TP=1, but will become a problem after we support TP for vLLM.

Fix #21

### DIFF
--- a/engine_integration/scripts/kvcached-vllm-v0.8.4.patch
+++ b/engine_integration/scripts/kvcached-vllm-v0.8.4.patch
@@ -1,8 +1,8 @@
 diff --git a/vllm/v1/core/block_pool.py b/vllm/v1/core/block_pool.py
-index 74f3f7852..83911ad01 100644
+index 74f3f7852..2ac1bcb01 100644
 --- a/vllm/v1/core/block_pool.py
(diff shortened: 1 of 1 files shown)


ivanium:
discussed with Jiarong offline and we'll need to revisit this after TP support is finished

jiarong0907:
This issue has been handled by https://github.com/ovg-project/kvcached/pull/13.

PR #23, opened 2025-07-15
[Doc] Update README

jiarong0907:


README.md
```diff
@@ -1,27 +1,29 @@
 # kvcached
 
-kvcached is a KV cache management system for serving multiple Large Language Models (LLMs) on a single GPU. It manages and coordinates the use of KV cache resources among co-located LLM instances, enabling them to share GPU memory. kvcached is compatible with popular LLM serving engines, including vLLM and SGLang.
+kvcached is a new KV cache management system that supports on-demand KV cache allocation. It implements the concept of GPU virtual memory, allowing applications to reserve virtual address space without immediately committing physical memory. Physical memory is then automatically allocated and mapped as needed at runtime. This capability allows multiple LLMs to run concurrently on a single GPU and flexibly share its memory, significantly improving GPU utilization and reducing memory fragmentation.
```
shanyu-sys, line 3: Since we are about to support the TP case, shall we say "allows multiple LLMs to run concurrently on the same GPUs“  in stead of "on a single GPU"?

jiarong0907: Good point. Let me fix it.

PR #27, opened 2025-07-15
[Chore] Disable mem leak in SGLang when kvcached is enabled

jiarong0907:


nanomaoli (approved):
This looks good and will suppress the warning when using SGLang.

PR #28, opened 2025-07-16
[Feat] Memory usage control and monitoring

jiarong0907:
Add memory usage control and monitoring for running kvcached instances.

To use:
```
python controller/kvctl.py
```

Supported functions:

```
kvcached> help
Available commands:
  list [ipc ...]               List IPC segments and usage
  limit <ipc> <size>           Set absolute limit (e.g. 512M, 2G)
  limit-percent <ipc> <pct>    Set limit as percentage of total GPU RAM
  watch [-n sec] [ipc ...]     Continuously display usage table
  kvtop [ipc ...] [--refresh r]  Launch curses kvtop UI (q to quit)
  !<shell cmd>                 Run command in system shell
  help                         Show this help message
  exit | quit                  Exit the shell

kvcached>
```


controller/__init__.py
ivanium, line 1: I actually feel we can leave this controller as a submodule of kvcached and expose a cli tool for it when packaging kvcached. Happy to discuss more offline

ivanium (commented):
Take a quick look and the code looks good and the tools are very cool! Not sure about the potential overhead to read/write shm so I'll try the code too.

PR #52, opened 2025-07-27
[Chore] Support sglang v0.4.9

jiarong0907:


ivanium (approved):
LGTM. Pushed a small fix to the index type when freeing the KV cache. Tested and it worked on my machine.

PR #53, opened 2025-07-27
Add Dockerfile

Thunderbeee:


Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh
+
+WORKDIR /workspace
+RUN git clone https://github.com/ovg-project/kvcached.git
+WORKDIR /workspace/kvcached
+RUN pip install . --no-build-isolation
+
+WORKDIR /workspace/kvcached/engine_integration/scripts
+RUN chmod +x setup.sh && ./setup.sh all
```
JustinTong0323: Do we need to set env for vllm? Maybe for sgl only is better.

JustinTong0323: like split into 2 docker files

JustinTong0323:
Don't forget to run `pre-commit run --all-files` before commit

Dockerfile
jiarong0907: Sounds good. I solved it.

PR #55, opened 2025-07-27
fix [integration/sglang-v0.4.6]: correct num_layers calculation

ivanium:
Should use `layer_num` rather than `end_layer - start_layer + 1`. The latter one counts one more layer.


ivanium:
No I saw it is fixed in v0.4.9 and the PR to SGLang. And then I realized there was an issue in our previous patch

PR #58, opened 2025-07-28
benchmark tp ipc overhead

nanomaoli:
A standalone benchmark for the performance of Inter-Process Communication involved in tensor-parallel inference (of vLLM, currently).


jiarong0907:
Thanks a lot! Could you all all your benchmark results? The current one seems to only have the sequential implementation.

nanomaoli:
> Thanks a lot! Could you all all your benchmark results? The current one seems to only have the sequential implementation.

Hi Dr. Xing, I put into README.md the results for 3 implementations from my test runs, as well as example commands. Also fixed a minor bug.

nanomaoli:
Submit the parallel mapping benchmark script + Minor edits

PR #54, opened 2025-07-27
[Fix]: Make available_size() and available blocks count correctly

ivanium:
This PR:
1. Correctly count allocated blocks with consideration for partially allocated pages and reserved blocks
2. Correct available_size() with consideration for reserved pages
3. Extracted embedded shared memory update logic from KVCacheManager into a separate MemInfoTracker class and moved it to PageAllocator to improve code organization and eliminate coupling.

Memory tracking now accurately reflects physical memory allocation/deallocation.
There are no functional changes to external APIs.

PR #60, opened 2025-07-31
[Feat] Configurable page size

ivanium:
This PR makes the page size configurable so that users can use `KVCACHED_PAGE_SIZE_MB` to set a larger page size and reduce memory mapping/unmapping overheads.

Need to be merged after PR #54 merged.


jiarong0907 (approved):
The code looks good to me. I tested it with SGLang and vLLM---both work well.

PR #61, opened 2025-07-31
[Feat] Contiguous KV tensor layout

ivanium:
This PR revises the FTensorAllocator to support a contiguous layout for all KV cache tensors. For each token, the new layout will put the K/V slot of all layers into a contiguous space, and hence they share the same page and only need to be mapped/unmapped once. To support various KV cache layouts needed by the upper-level engine, the C++ extension will view and stride the contiguous tensor in a correct way and return a set of tensor views.

This PR needs to be merged after PR #54 and PR #60


ivanium:
close this PR as it will be addressed in PR #62

PR #63, opened 2025-08-06
modify Dockerfile

Thunderbeee:


docker/Dockerfile.sglang
```diff
@@ -1,38 +1,29 @@
-ARG CUDA_VERSION=12.6.1
-FROM nvidia/cuda:${CUDA_VERSION}-devel-ubuntu22.04
+FROM pytorch/pytorch:2.6.0-cuda12.6-cudnn9-devel
```
jiarong0907, line 1: Can we avoid using specific pytorch version? vllm and sglang later might rely on updated version of pytorch, and I believe their setup scripts will automatically install pytorch.

docker/Dockerfile.sglang
```diff
@@ -1,38 +1,29 @@
...
 WORKDIR /workspace/kvcached
-RUN uv pip install --system -r requirements.txt
+RUN pip install . --no-build-isolation
 
-WORKDIR /workspace
-RUN git clone -b v0.4.9 https://github.com/sgl-project/sglang.git sglang-v0.4.9
-WORKDIR /workspace/sglang-v0.4.9
-RUN uv pip install --system -e "python[all]"
-RUN git apply /workspace/kvcached/engine_integration/scripts/kvcached-sglang-v0.4.9.patch
+WORKDIR /workspace/kvcached/engine_integration/scripts
```
jiarong0907, line 23: Can we avoid using the setup script directly? It will still using uv and install the engine inside kvcached folder. If you look at my current version, I want to install the engine source code directly under /workspace without using uv.

docker/Dockerfile.sglang
```diff
@@ -1,38 +1,29 @@
...
+RUN chmod +x setup.sh && ./setup.sh sglang
 
+EXPOSE 8080 30000 30001
```
jiarong0907, line 26: vllm and sglang are using different ports. Please check.

docker/Dockerfile.vllm
```diff
@@ -1,38 +1,29 @@
...
+EXPOSE 8080 30000 30001
 
 WORKDIR /workspace/kvcached
-RUN uv pip install --system -e . --no-build-isolation
-
-EXPOSE 8080 12346 30000 30001
-
-CMD ["bash"]
\ No newline at end of file
+CMD ["bash"]
```
jiarong0907, line 29: Same comments as the sglang setup.

jiarong0907 (changes requested):
See my comments above.

docker/Dockerfile.sglang
Thunderbeee: thanks Jiarong!

docker/Dockerfile.sglang
Thunderbeee: I try to fix based on previous (old) Dockerfile, but always give me some error when loading model. Let me take a look and I will try to pull the latest docker image to see if it works already. Thanks!

docker/Dockerfile.sglang
Thunderbeee: thanks! I will first try to pull the lates docker image again

jiarong0907:
Replaced by https://github.com/ovg-project/kvcached/pull/68.

PR #75, opened 2025-08-23
Update mem info tracker test

nanomaoli:
This pull request will update stale tests under `/kvcached/test/`

This first commit adds a test file to replace the stale one `test_shm_client.py`.
It covers below cases:

1. modify the `used_size` entry
2. modify the `prealloc_size` entry
3. coordinate racing processes with lock
4. call `update_memory_usage()`
5. call `check_and_get_resize_target()`


jiarong0907 (approved):
This needs to run with the kvcached installed python environment.

```
pytest test_shm_info_tracker.py
```
Might need to install pytest first

```
python -m pip install pytest
```

PR #79, opened 2025-08-25
[Feat]: Model traffic statistics, sleep mode, and activation

cui36:
- Add traffic monitor (controller/traffic_monitor.py) and sleep and wake up function (controller/sleep_manager.py)
    - Implement request rate tracking per model
    - Add new API endpoints for traffic statistics
- Add test file (test/test_traffic_monitor.py and test/test_sleep_manager.py)
    - Add idle time monitoring for sleep mode detection
    - Test the whole traffic monitoring functionality
- Update README documentation (controller/README.md)


cui36:
@jiarong0907 This is a feature branch and Task 1 will be updated incrementally. I'll add a comment when all TODO items are completed and the PR is ready for final review and merge.

pyproject.toml
```diff
@@ -40,6 +40,7 @@ classifiers = [
 dependencies = [
     "numpy>=1.26.0",
     "posix_ipc>=1.0",
+    "aiohttp>=3.8.0",
```
jiarong0907: sglang and vllm already have aiohttp. Not sure whether we should install our own one as well. Should we run kvcached within vllm or sglang's virtual environment?

engine_integration/benchmark/start_client.sh
```diff
@@ -11,7 +11,7 @@ DEFAULT_VLLM_PORT=12346
 DEFAULT_SGL_PORT=30000
 DEFAULT_MODE="dev"
 
-NUM_PROMPTS=1000
+NUM_PROMPTS=10
```
jiarong0907: Let's do not commit these local temporal changes.

controller/README.md
```diff
@@ -84,3 +84,247 @@ python benchmark.py --config example-config.yaml
 ```
 
 A separate *tmux* session (`benchmark-<name>`) is created for every model so you can watch latency/throughput side by side.
+
+## Testing Scripts
```
jiarong0907: The README here is to describe the functionality of the controller. You should describe the sleep model and traffic monitoring here, could be with some examples to test it, but not the entire test suite. All the tests should be moved to the test folder under project root.

controller/example-config.yaml
```diff
@@ -7,37 +7,58 @@ kvcached: # kvcached environment variables for all instances
...
+  #   venv_path: ../engine_integration/sglang-v0.4.6.post2/.venv
+  #   kvcached_env:
+  #     - "ENABLE_KVCACHED=true"
+  #     - "KVCACHED_IPC_NAME=SGLANG"
+  #   engine_args:
+  #     - "--disable-radix-cache"
+  #     - "--trust-remote-code"
+  #     - "--mem-fraction-static 0.5"
+  #     - "--host=localhost"
+  #     - "--port=30000"
```
jiarong0907: Why there is a comment block.

controller/frontend.py
```diff
@@ -35,6 +37,22 @@ def configure_endpoints(self):
...
+        self.app.router.add_get('/traffic/idle', self.handle_idle_models)
+        self.app.router.add_get('/traffic/active', self.handle_active_models)
+
+        # Sleep management endpoints
+        self.app.router.add_get('/sleep/status', self.handle_sleep_status)
+        self.app.router.add_post('/sleep/model/{model_name}',
+                                 self.handle_sleep_model)
+        self.app.router.add_post('/wake/model/{model_name}',
+                                 self.handle_wake_model)
+        self.app.router.add_get('/sleep/candidates',
```
jiarong0907: For all these handles, better to name them starting from /model

controller/sleep_manager.py
```diff
@@ -0,0 +1,361 @@
+import asyncio
```
jiarong0907, line 1: The file name sleep_manager is not good because this file does both mointoring and sleep/wake. I suggest split it into two models. One does traffic monitoring; the other does model status management.

jiarong0907 (changes requested):

controller/sleep_manager.py
cui36: Done, right now we have two files: one is named "traffic_monitor.py" and the other is named "sleep_manager.py"

controller/frontend.py
cui36: Done

controller/example-config.yaml
cui36: As I test all the instances from vLLM, but will restore the former content.

engine_integration/benchmark/start_client.sh
cui36: Done

pyproject.toml
cui36: Delete that. Yes, we currently are running kvcached in vllm and sgl's venv.

jiarong0907:
@cui36 Are the following done?

> TODO
Add idle time monitoring for sleep mode detection
Test the whole traffic monitoring functionality

cui36:
> @cui36 Are the following done?
> 
> > TODO
> > Add idle time monitoring for sleep mode detection
> > Test the whole traffic monitoring functionality

Yes, all done.

controller/sleep_manager.py
```diff
@@ -0,0 +1,473 @@
...
+                        )
+                        return True
+                    else:
+                        logger.error(
+                            f"SGLang resume API returned status {response.status}: {await response.text()}"
+                        )
+                        return False
+        except Exception as e:
+            logger.error(f"Error calling SGLang resume API at {url}: {e}")
+            return False
```
jiarong0907: SGLang needs more steps, like `engine.update_weights_from_tensor(list(hf_model_new.named_parameters()))`?

controller/sleep_manager.py
```diff
@@ -0,0 +1,473 @@
...
+        str]] = None  # model_name -> {"host": "localhost", "port": "8000"}
+
+
+class SleepManager:
+    """Manages sleep mode for idle models to save resources"""
+
+    def __init__(self,
+                 config: Optional[SleepConfig] = None,
+                 traffic_monitor: Optional[TrafficMonitor] = None):
+        self.config = config or SleepConfig()
```
jiarong0907, line 43: Better to support parsing the .yaml file to get these configs. I mean when using launch.py to launch everything.

jiarong0907 (commented):
I chanaged some code. Please rerun the tests.

controller/sleep_manager.py
cui36: I see

controller/sleep_manager.py
cui36: Yes right now in our code, we implement the most necessary APIs. 
- We did not add the additional because the current kvcached has no explicit clear operation implemented yet.
- Regarding SGL, their future plan is to transition to an **offload-based** sleep/wakeup mechanism instead of deletion, so there is no need for update_weights_from_tensor().

So do you think we need to add those APIs right now?

controller/example-config.yaml
cui36, line 1: @jiarong0907 
`venv_path:  /home/jrxing/kvcached/engine_integration/vllm-v0.9.2/.venv`
Do we need to keep the absolute paths, or should we switch to safer and more flexible relative paths?
And also sgl version. According to the latest Docker file, the SGL version should be sglang-v0.4.9.

jiarong0907: relatively path could be better.

controller/example-config.yaml
```diff
@@ -5,33 +5,44 @@ kvcached: # kvcached environment variables for all instances
...
     using_venv: true
-    venv_path:  /home/jrxing/kvcached/engine_integration/vllm-v0.9.2/.venv
+    venv_path:  ../engine_integration/vllm-v0.9.2/.venv
     kvcached_env: # kvcached environment variables for this instance
       - "ENABLE_KVCACHED=true"
       - "KVCACHED_IPC_NAME=VLLM"
     engine_env:
       - "VLLM_USE_V1=1"
       - "VLLM_ATTENTION_BACKEND=FLASH_ATTN"
+      - "VLLM_SERVER_DEV_MODE=1"
```
jiarong0907: What does `- "VLLM_SERVER_DEV_MODE=1"` do?

controller/launch.py
```diff
@@ -267,6 +373,30 @@ def main() -> None:
         logger.error("Invalid configuration: %s", e)
         sys.exit(1)
 
+    # Parse sleep manager configuration
+    sleep_manager_config = _parse_sleep_manager_config(raw_cfg)
+
+    # Extract model configurations for sleep manager
+    vllm_models_config, sglang_models_config = _extract_model_configs_for_sleep_manager(
+        instances_cfg)
```
jiarong0907: Is this overlapping with `instances_cfg = _parse_cfg(raw_cfg, cfg_path.parent)`? I think we don't need to parse the .yaml again; instead, directly get the config from `instances_cfg`?

controller/launch.py
```diff
@@ -267,6 +373,30 @@ def main() -> None:
...
+
+    # Update sleep manager config with extracted model configs
+    sleep_manager_config["vllm_models_config"] = vllm_models_config
+    sleep_manager_config["sglang_models_config"] = sglang_models_config
+
+    # Pass sleep manager configuration directly through environment variables
+    global_kvcached_env["SLEEP_MANAGER_CONFIG_PATH"] = str(cfg_path)
```
jiarong0907: Is this path being used later?

controller/launch.py
```diff
@@ -267,6 +373,30 @@ def main() -> None:
...
+        sleep_manager_config["auto_sleep_enabled"])
+    global_kvcached_env["SLEEP_MANAGER_IDLE_THRESHOLD_SECONDS"] = str(
+        sleep_manager_config["idle_threshold_seconds"])
+    global_kvcached_env["SLEEP_MANAGER_CHECK_INTERVAL_SECONDS"] = str(
+        sleep_manager_config["check_interval_seconds"])
+    global_kvcached_env["SLEEP_MANAGER_WAKEUP_ON_REQUEST"] = str(
+        sleep_manager_config["wakeup_on_request"])
+    global_kvcached_env["SLEEP_MANAGER_MIN_SLEEP_DURATION"] = str(
+        sleep_manager_config["min_sleep_duration"])
+
```
jiarong0907: How are these global env used later?

controller/sleep_manager.py
```diff
@@ -0,0 +1,642 @@
...
+    check_interval_seconds: int = 60  # Check every minute
+    auto_sleep_enabled: bool = False  # Whether to automatically put models to sleep
+    wakeup_on_request: bool = True  # Whether to automatically wake models on request
+    min_sleep_duration: int = 60  # Minimum time to keep model asleep (seconds)
+    vllm_models_config: Dict[str, Dict[
+        str,
+        str]] = None  # model_name -> {"host": "localhost", "port": "8000"}
+    sglang_models_config: Dict[str, Dict[
+        str,
+        str]] = None  # model_name -> {"host": "localhost", "port": "8000"} for memory occupation control and full model recovery
```
jiarong0907, line 27: We have default config here again. Seems to be repeating launch.py. Maybe we should them here and remove the default values in luanch.py?

requirements.txt
```diff
@@ -2,4 +2,5 @@
 setuptools>=64
 wheel
 pytest
-posix_ipc>=1.0
\ No newline at end of file
+posix_ipc>=1.0
+pyyaml>=6.0
```
jiarong0907: Check whether this conflicts with vllm and sglang's requirements.

controller/launch.py
cui36: Yes, it is used. But I think it is useless. So delete.

controller/sleep_manager.py
cui36: In the sleep_manager.py, the SleepConfig part is a constructor. So I think keep an initial value is ok.

cui36: @jiarong0907 Follow-up question: do we want to set different sleep patterns for different models (instances), or keep the value static?
This is our system design right now (in the view of sleepconfig rather than model):
SleepConfig 
├── auto_sleep_enabled: bool          
├── idle_threshold_seconds: int       
├── check_interval_seconds: int       
├── wakeup_on_request: bool           
├── min_sleep_duration: int           
├── vllm_models_config: Dict          
│   ├── "model1": {"host": "...", "port": "..."}
│   └── "model2": {"host": "...", "port": "..."}
└── sglang_models_config: Dict        
    ├── "model3": {"host": "...", "port": "..."}
    └── "model4": {"host": "...", "port": "..."}

controller/launch.py
cui36: Good point. Combined together

requirements.txt
cui36: Yes, the version is satisfied. But these two libs have already been installed on the venv env. So maybe delete them could be better.

requirements.txt
```diff
@@ -1,5 +1,4 @@
 # Should be mirrored in pyproject.toml
 setuptools>=64
 wheel
-pytest
-posix_ipc>=1.0
```
jiarong0907: Why posix_ipc is deleted?

controller/sleep_manager.py
jiarong0907: We don't need to set different sleep patterns for now.

For the default values, two copies will be very confusing. As a user, she might not know which one to change.

requirements.txt
cui36: Oh I've checked the initial pip list, and posix_ipc is already included.

cui36: Well I think this file can keep the same with the main branch

PR #85, opened 2025-08-27
async ipc, update ipc benchmark

nanomaoli:
This PR
1. changes to async ipc for tp
2. updates previous ipc benchmarks to work with contiguous tensor layout
3. fixes init logging on total_mem_size

### DIFF
--- a/kvcached/tp_ipc_util.py
+++ b/kvcached/tp_ipc_util.py
@@ -108,8 +108,7 @@ def start_worker_listener_thread(rank: int):
     t.start()
 
 
-async def send_and_receive_message_async(rank: int,
-                                         message: Message) -> Message:
+async def _send_and_receive_message(rank: int, message: Message) -> Message:
     """
     Send a message to the worker and receive a response asynchronously.
     """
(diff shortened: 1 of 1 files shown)


benchmarks/bench_map_parallelism/kvcached_map_parallel_benchmark.py
```diff
@@ -331,8 +338,21 @@ def main():
...
     args = ap.parse_args()
 
+    os.environ["KVCACHED_CONTIGUOUS_LAYOUT"] = (
+        "false" if args.not_contiguous else "true"
+    )
+    contiguous_layout = (
+        os.getenv("KVCACHED_CONTIGUOUS_LAYOUT", "true").lower() == "true"
+    )
+    print(f"Using layout contiguous={contiguous_layout}")
+
```
jiarong0907, line 355: This is a bit confused to me. If the os env will be overwritten by the input arguement, we don't need care about it? Just use the input arguement.

jiarong0907: Oh, IC. Do you do this because the rest of kvcached needs to use the os env? If so, you are correct. Ignore my comments.

jiarong0907:
I changed some function names. Please retest whether it can run.

benchmarks/bench_map_parallelism/kvcached_map_parallel_benchmark.py
nanomaoli: Yes, this is to emulate kvcached's initialization. The environment variable is used in `kvcached/utils.py`.

jiarong0907 (approved):

PR #97, opened 2025-09-01
[Bench] vLLM-kvcached latency benefit benchmark framework

jiarong0907:


controller/example-config.yaml
jiarong0907, line 1: Let's do not change this example config file for benchmark purpose. If we need a new config file, we should put it into the corresponding benchmark folders.

controller/example-config.yaml
```diff
@@ -36,20 +36,41 @@ instances: # instances configuration
       - "--no-enable-prefix-caching"
       - "--host=localhost"
       - "--port=12346"
-      - "--gpu-memory-utilization 0.5"
+      - "--gpu-memory-utilization 0.75"
```
jiarong0907: Why is this only 0.75?

benchmarks/common/vllm/datasets.py
```diff
@@ -0,0 +1,102 @@
...
+
+from transformers import PreTrainedTokenizerBase
+
+# Pull the canonical base classes / helpers from the installed vLLM package.
+from vllm.benchmarks.datasets import BenchmarkDataset, SampleRequest, is_valid_sequence
+
+__all__ = ["MyCustomDataset"]
+
+
+class MyCustomDataset(BenchmarkDataset):
```
jiarong0907: Better classname?

benchmarks/bench_latency_benefit/create_custom_dataset.py
```diff
@@ -0,0 +1,129 @@
...
+    if seed is not None:
+        random.seed(seed)
+    
+    # Generate random text using letters and spaces
+    chars = string.ascii_letters + " "
+    text = ''.join(random.choice(chars) for _ in range(length))
+    return text
+
+
+def generate_completion(prompt: str, completion_length: int, seed: Optional[int] = None) -> str:
```
jiarong0907: Do you must have completion text? The output text is generated by the model inference, right? Doesn't make a lot of sense to generate ours. If this is required by the vllm benchmark suite, we can keep this.

benchmarks/bench_latency_benefit/create_custom_dataset.py
```diff
@@ -0,0 +1,129 @@
...
+
+import argparse
+import json
+import random
+import string
+from pathlib import Path
+from typing import Optional
+
+
+def generate_random_text(length: int, seed: Optional[int] = None) -> str:
```
jiarong0907: Is it matter if the input text is random text or just `hello`*N times.

benchmarks/bench_latency_benefit/create_custom_dataset.py
```diff
@@ -0,0 +1,129 @@
...
+    )
+    
+    args = parser.parse_args()
+    
+    create_dataset(
+        output_path=args.output,
+        num_samples=args.num_samples,
+        prompt_length=args.prompt_length,
+        completion_length=args.completion_length,
+        seed=args.seed
```
jiarong0907: If this dataset only has random input and output texts. Not sure whether we need to create the file to store it as I mentioned above both input and output content may not be important. 

If the dataset has timing information, that's a different story.

benchmarks/bench_latency_benefit/plot_request_rate.py
```diff
@@ -0,0 +1,279 @@
...
+    results = load_benchmark_results(args.result_files)
+    
+    if not results:
+        print("Error: No valid benchmark results found")
+        return 1
+    
+    print(f"Loaded {len(results)} benchmark results")
+    
+    # Ensure output directory exists
+    import os
```
jiarong0907: Why import here? Same for the Gemini's comment.

benchmarks/bench_latency_benefit/results/metrics/vllm-Qwen-Qwen3-14B-qps_100-prompt_1024-completion_128-1.json
jiarong0907, line 1: I think in our final PR. We only include the result related to the figure drawn.

benchmarks/common/vllm/datasets.py
```diff
@@ -0,0 +1,102 @@
...
+            requests.append(
+                SampleRequest(
+                    prompt=prompt,
+                    prompt_len=prompt_len,
+                    expected_output_len=exp_output_len,
+                )
+            )
+
+        # If we have fewer than needed, oversample to pad.
+        self.maybe_oversample_requests(requests, num_requests)
```
jiarong0907: I am curious where the timestamp of the request is determined.

jiarong0907: Never mind. I found it in the benchmark suite.

controller/example-config.yaml
cui36: Got it.

controller/example-config.yaml
cui36: Yes, the latest version has changed to 0.5.

benchmarks/common/vllm/datasets.py
cui36: Can we change this file?

benchmarks/bench_latency_benefit/plot_request_rate.py
cui36: Solve

benchmarks/bench_latency_benefit/results/metrics/vllm-Qwen-Qwen3-14B-qps_100-prompt_1024-completion_128-1.json
cui36: Got it.

benchmarks/bench_latency_benefit/create_custom_dataset.py
cui36: Yes we could delete them in the benchmark.

benchmarks/bench_latency_benefit/create_custom_dataset.py
cui36: You are right. I find there already exists "--random-input-len" option in vllm. So we don't need to create our own dataset (even "hello" * N times).

benchmarks/common/vllm/datasets.py
jiarong0907: I mean the class name `MyCustomDataSet` to a more descriptive name?

benchmarks/bench_latency_benefit/bench_kvcached_vllm.py
```diff
@@ -0,0 +1,1416 @@
...
+    RandomDataset,
+    SampleRequest,
+    ShareGPTDataset,
+    SonnetDataset,
+    VisionArenaDataset,
+)
+from benchmark_utils import convert_to_pytorch_benchmark_format, write_to_json
+from benchmarks.common.vllm.datasets import (  # Added custom dataset
+    MyCustomDataset,
+)
```
jiarong0907: Delete this? No error here?

PR #98, opened 2025-09-01
[Setup] Remove yapf, and fix isort-ruff conflicts

jiarong0907:


.pre-commit-config.yaml
```diff
@@ -6,20 +6,13 @@ default_stages:
   - manual # Run in CI
 exclude: 'engine_integration/vllm-v0.8.4|engine_integration/sglang-v0.4.6.post2'
```
ivanium: Occasionally spotted this. Can we make the `exclude` more general to different vllm/sglang versions?

ivanium (approved):
LGTM in general. Approve first. I left a minor comment on excluding vllm/sglang source code

.pre-commit-config.yaml
jiarong0907: good catch

jiarong0907: fixed

PR #100, opened 2025-09-03
Update tests

nanomaoli:
### Changes Made
`test_simulator.py`: This is **obsolete** after we have real kvcached integration with SGLang and vLLM, so I **removed** it.

`test_shm.py`: This basically tests shm read, which is now **done with** MemInfoStruct and RwLockedShm, and **covered in** test_shm_info_tracker, so I also **removed** it.

`test_multi_processing_read_mem_usage.py`: This file benchmarks shm read/write overhead, which is less meaningful to fail or pass as unit tests. **_Maybe have a simple benchmark file in benchmarks?_**

`test_kvcached.py`: This is also more like a benchmark, and it is measuring the map_to_kv_tensors method, which is **covered in** an existing benchmark bench_map_parallelism, so I **removed** this.

### Files Left Unchanged For Now, Need Confirmation:
`test_kv_cache_manager.py`: Currently a print-based exploratory script. I think of **extracting unit tests** and crafting a test file **based on my understanding** of KVCacheManager's methods. **_Is this good?_**

`bench_page_prealloc.py`: This is a benchmark for page allocation performance. **_Maybe polish and move it to benchmarks?_**


jiarong0907 (approved):

jiarong0907:
@nanomaoli Please check my comments above. Thanks!

nanomaoli:
Message received.

Todo:
Adding new unit tests for the uncovered areas — kv cache manager, page allocators, controllers, and any other components that don’t yet have coverage.

jiarong0907:
> Message received.
> 
> Todo: Adding new unit tests for the uncovered areas — kv cache manager, page allocators, controllers, and any other components that don’t yet have coverage.

Yes, we can open another PR for this.

PR #108, opened 2025-09-07
[Chore] Remove mem-util limit

jiarong0907:
- Remote static memory utilization limit in benchmark scripts
- Disable vLLM requested and available memory size check

### DIFF
--- a/benchmarks/simple_bench/start_server.sh
+++ b/benchmarks/simple_bench/start_server.sh
@@ -139,7 +139,6 @@ if [ "$engine" == "vllm" ]; then
     vllm serve "$MODEL" \
     --disable-log-requests \
     --no-enable-prefix-caching \
-    --gpu-memory-utilization 0.5 \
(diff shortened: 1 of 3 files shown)


ivanium:
> > LGTM in general.
> > To make sure I understand, so this PR disables vLLM's free-memory check when kvcached is enabled, but it will still preserve user-specified values for the kv cache memory usage if memory is sufficient. For example, if the GPU has enough memory and the user sets vllm `gpu-memory-utilization 0.5`, then vLLM will still allocate at most half of GPU memory.
> > Is this correct? The behavior sounds good to me, just wanted to double check.
> 
> Yes, this is correct. It only disables the check where user sets a large limit, but the current memory is not enough for that limit.

Sounds great. In SGLang we have already done that and now this PR adds the missing piece for vLLM, which is great!

ivanium (approved):

Issue #106, opened 2025-09-07
Unsupported shape '[182376, 40, 2, 40, 128]'


jiarong0907:
Model: meta-llama/Llama-2-13b-hf

```
[2025-09-07 04:15:26] Load weight end. type=LlamaForCausalLM, dtype=torch.float16, avail mem=114.90 GB, mem usage=24.30 GB.
[kvcached][INFO][2025-09-07 04:15:26][page_allocator.py:149] Init kvcached KV cache allocator: num_layers=40, mem_size_per_layer=1312MB, total_mem_size=104961MB, page_size=2MB, tp_size=1, async_sched=True, contiguous_layout=True, enable_prealloc=True
[2025-09-07 04:15:27] Scheduler hit an exception: Traceback (most recent call last):
  File "/workspace/kvcached/engine_integration/sglang-v0.4.9/python/sglang/srt/managers/scheduler.py", line 2723, in run_scheduler_process
    scheduler = Scheduler(server_args, port_args, gpu_id, tp_rank, pp_rank, dp_rank)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/workspace/kvcached/engine_integration/sglang-v0.4.9/python/sglang/srt/managers/scheduler.py", line 338, in __init__
    self.tp_worker = TpWorkerClass(
                     ^^^^^^^^^^^^^^
  File "/workspace/kvcached/engine_integration/sglang-v0.4.9/python/sglang/srt/managers/tp_worker_overlap_thread.py", line 66, in __init__
    self.worker = TpModelWorker(
                  ^^^^^^^^^^^^^^
  File "/workspace/kvcached/engine_integration/sglang-v0.4.9/python/sglang/srt/managers/tp_worker.py", line 81, in __init__
    self.model_runner = ModelRunner(
                        ^^^^^^^^^^^^
  File "/workspace/kvcached/engine_integration/sglang-v0.4.9/python/sglang/srt/model_executor/model_runner.py", line 237, in __init__
    self.initialize(min_per_gpu_memory)
  File "/workspace/kvcached/engine_integration/sglang-v0.4.9/python/sglang/srt/model_executor/model_runner.py", line 310, in initialize
    self.init_memory_pool(
[...]

PR #122, opened 2025-09-15
Fix Llama-2-13b-hf unsupported shape issue on sglang (issue #106)

qinganrice:
Hi community maintainers,

I’ve made some changes to address issue #106. The root cause of the shape mismatch is that the virtual memory size computed in `interface.py (/kvcached/kvcached/integration/sglang) ` does not match the actual allocated memory size after page boundary alignment in `allocator.cpp (/kvcached/csrc)`. This alignment can produce a total number of elements that isn’t perfectly divisible by the model’s dimensions.

Please let me know if you have any questions. It’s a minor contribution, but I hope it helps!


jiarong0907:
@qinganrice Thank you! Can you fix the pre-commit problem and remove unnecesary code changes?

kvcached/integration/sglang/interfaces.py
```diff
@@ -1,19 +1,14 @@
-# SPDX-FileCopyrightText: Copyright contributors to the kvcached project
-# SPDX-License-Identifier: Apache-2.0
-
```
jiarong0907, line 3: Do not delete these copyright lines.

kvcached/integration/sglang/interfaces.py
```diff
@@ -1,19 +1,14 @@
...
 import math
 from typing import List, Optional, Tuple
 
 import torch
 
 from kvcached.kv_cache_manager import KVCacheManager
-from kvcached.tp_ipc_util import start_worker_listener_thread
+from kvcached.tp_ipc_util import start_worker_listerner_thread
```
jiarong0907: Typo here? listener. We should have fixed this problem.

kvcached/integration/sglang/interfaces.py
```diff
@@ -1,19 +1,14 @@
...
 from kvcached.utils import CONTIGUOUS_LAYOUT, PAGE_SIZE, get_kvcached_logger
-from kvcached.vmm_ops import (
-    create_kv_tensors,
-    init_kvcached as _init_kvcached_impl,
-    shutdown_kvcached as _shutdown_kvcached_impl,
-)
+from kvcached.vmm_ops import create_kv_tensors
+from kvcached.vmm_ops import init_kvcached as _init_kvcached_impl
+from kvcached.vmm_ops import shutdown_kvcached as _shutdown_kvcached_impl
```
jiarong0907: This seems to be the order version of code. You might want to merge main into your branch.

kvcached/integration/sglang/interfaces.py
```diff
@@ -35,6 +30,7 @@ def init_kvcached(
 
     if device is None:
         device = f"cuda:{torch.cuda.current_device()}"
+        # device = "cuda:1"
```
jiarong0907: Delete unrelated code.

kvcached/integration/sglang/interfaces.py
```diff
@@ -43,7 +39,7 @@ def init_kvcached(
 
     if tp_size > 1:
         # start the listener thread for tensor parallel kv cache management
-        start_worker_listener_thread(torch.cuda.current_device())
+        start_worker_listerner_thread(torch.cuda.current_device())
```
jiarong0907: same typo

jiarong0907:
Would be helpful to run this before commit.

We use pre-commit to ensure a consistent coding style. You can set it up by

```
pip install pre-commit
pre-commit install
```

Before pushing your code, please run the following check and make sure your code passes all checks.

```
pre-commit run --all-files
```

jiarong0907:
It works for me for sglang with `_contiguous_layout`. Could you also fix the same problem for non `_contiguous_layout` and for vLLM? Thanks!

ivanium:
Thanks for the PR! Yes I am taking a look at this PR and making some minor changes. Will take it over.

qinganrice:
> I tested the code, and it worked for me. @ivanium It would be good to take a look because it changes the core code.
> 
> One quick question is why we don't need alignment in c++ when contigous is enabled.

Thanks for your review and comments. My understanding is that alignment of contiguous layout is already done by [line](https://github.com/ovg-project/kvcached/blob/c99dfb3b416505bf3bc046eb60b6e37d415373c3/csrc/allocator.cpp#L224C3-L225C65). But alignment of non-contiguous (size of each layer) is missed before tensor creation.  v_base_offset also has the similar issue. Feel free to correct me if it's wrong!

ivanium (approved):

ivanium:
LGTM now. Unified size alignment logic. Resolved Gemini review issues. Close #106

Issue #91, opened 2025-08-31
[TODO] Add an example of vLLM semantics router?


jiarong0907:
Can we add an exmple to demonstrate kvcached with vLLM semantics router? https://vllm-semantic-router.com/

We can run multiple models on one GPU for the router to choose, including the sleep and wakeup functionality based on the traffic monitoring status.

PR #135, opened 2025-09-28
[Example] Add vllm semantic router example

ztang2370:
Issue: https://github.com/ovg-project/kvcached/issues/91

This is WIP, a basic working example.

TODO:
1. Include the sleep and wakeup functionality based on the traffic monitoring status.
2. As vllm semantic router doesn't have a release for now, the patches need to be updated later after they have a release.

### DIFF
--- a/examples/03_model_router_sleep/README.md
+++ b/examples/03_model_router_sleep/README.md
@@ -1,0 +1,250 @@
+# KVCached with vLLM Semantic Router Example
+
+This example demonstrates how to integrate KVCached with the [vLLM Semantic Router](https://github.com/vllm-project/semantic-router) for intelligent model routing based on query content. The semantic router automatically routes requests to the most appropriate model endpoint based on semantic analysis of the input using BERT-based classification.
+
+The vLLM Semantic Router is an intelligent Mixture-of-Models (MoM) router that directs OpenAI API requests to the most suitable models from a defined pool based on semantic understanding of the request's intent, complexity, and task type.
+
+## Overview
+
+The setup consists of:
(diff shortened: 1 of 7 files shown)


cui36:
Great! I think the sleep and wakeup functionality has already been merged, right? @jiarong0907

jiarong0907:
@ztang2370 @cui36 Having vllm semantic router is great, but I would suggest we add it as a feature later. 

For the example, we can just use the current router we have. The example just needs to show the features of the router and sleeping. This will be the `03_model_router_sleep`.

ztang2370:
> @ztang2370 @cui36 Having vllm semantic router is great, but I would suggest we add it as a feature later.
> 
> For the example, we can just use the current router we have. The example just needs to show the features of the router and sleeping. This will be the `03_model_router_sleep`.

But my understanding is the features of routing and sleeping are already shown in controller, isn't it?

jiarong0907:
> > @ztang2370 @cui36 Having vllm semantic router is great, but I would suggest we add it as a feature later.
> > For the example, we can just use the current router we have. The example just needs to show the features of the router and sleeping. This will be the `03_model_router_sleep`.
> 
> But my understanding is the features of routing and sleeping are already shown in controller, isn't it?

Yes, but we need an end to end example to tell users how this works and can be used.

cui36:
> > > @ztang2370 @cui36 Having vllm semantic router is great, but I would suggest we add it as a feature later.
> > > For the example, we can just use the current router we have. The example just needs to show the features of the router and sleeping. This will be the `03_model_router_sleep`.
> > 
> > 
> > But my understanding is the features of routing and sleeping are already shown in controller, isn't it?
> 
> Yes, but we need an end to end example to tell users how this works and can be used.

Oh I see. Will update some info there today.

PR #137, opened 2025-09-29
[Example] Add multi agent example

shanyu-sys:


shanyu-sys:
> @shanyu-sys Thanks for providing this example! Can we improve this example by replacing the frontend with popular agent programming framework, e.g. LangChain?

Updated.

PR #143, opened 2025-09-30
update: assets ttft figures

cui36:


jiarong0907:
@cui36 Thanks for the updates. Could we keep the fine-grained xtics on the y-axis? Also I think the title on top of each figure is OK, we can just keep it.

jiarong0907 (approved):

Issue #211, opened 2025-11-06
Crash when calling release_memory_occupation() with SGLang backend


hsuchifeng:
<img width="1318" height="650" alt="Image" src="https://github.com/user-attachments/assets/27cc4e9c-338f-4104-beb0-bfe99b145a92" />

ivanium:
Hi @hsuchifeng, thanks for trying out kvcached! 

This is a known issue #95 because the current kvcached has supported `clear()` to clean up all KV caches. We will fix this quickly and let you know when it's done.

PR #213, opened 2025-11-07
fix Issue #211  Crash when calling release_memory_occupation() with SGLang backend

jiahe7ay:
image

Fixed issue #211 and merged PR #150; now both the sleep and wakeup functions in sglang and vllm are working properly.  
Currently, kvcached does not use the prefix cache, so we force reset_prefix_cache to return true in order to adapt to the sleep feature.

### DIFF
--- a/kvcached/kv_cache_manager.py
+++ b/kvcached/kv_cache_manager.py
@@ -116,19 +116,11 @@ class KVCacheManager:
                 if total_wait >= KV_TENSOR_WAIT_TIMEOUT:
                     raise TimeoutError("KV tensors not created after "
                                        f"{KV_TENSOR_WAIT_TIMEOUT} seconds")
-
                 time.sleep(0.001)  # 1ms
                 total_wait += 0.001
-
-            # Reserve the first block as null block for padding tokens
-            if self.reserve_null_block:
-                # Skip the wait to avoid dead-lock with the event.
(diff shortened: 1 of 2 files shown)


kvcached/integration/vllm/patches.py
```diff
@@ -104,9 +104,13 @@ def free_blocks(
                 ]
                 if len(block_ids) > 0:
                     self.kv_cache_manager.free(block_ids)
-
+            """
+            Currently, kvcached does not use the prefix cache, so we force reset_prefix_cache to return true in order to adapt to the sleep feature.
+            Once prefix cache support is added in the future, it will still need to be implemented.
+            """
```
ivanium: I like the comments here 👍

kvcached/kv_cache_manager.py
ivanium, line 1: Is this the same as #150 but just removed the `@synchronized` decorator? Is that intentional or neglect?

jiahe7ay: neglect

jiahe7ay:
> I fixed #150 and merged it. You can rebase to the current main and I can take a second look. The major delta here will be the comments for ElasticCachePool's prefix-cache-related API but I'd be happy to merge those.

I will rebase again to the current main."

PR #234, opened 2026-01-26
[Fix] Fix SGLang tensor parallel rank bug and TP IPC broadcast logic

rishabhsinha17:
## Summary

This PR fixes critical bugs in SGLang's tensor parallel support and the core TP IPC broadcast logic.

### Bug Fixes

1. **SGLang TP rank bug**: Fix `start_worker_listener_thread()` to use `tp_rank` instead of `torch.cuda.current_device()`. The socket path is based on rank, not device ID, causing IPC failures in multi-GPU setups.

2. **Broadcast functions sending to rank 0**: The `broadcast_*` functions were iterating `range(tp_size)` instead of `range(1, tp_size)`, sending messages to rank 0 which has no listener thread.

3. **Local + remote check for kv_tensors_created**: In TP mode, the coordinator now checks locally first, then broadcasts to workers. Previously it only checked remote workers.

4. **Map/unmap always runs locally first**: The page allocator now always calls `map_to_kv_tensors`/`unmap_from_kv_tensors` locally on rank 0, then broadcasts to workers if `tp_size > 1`.

### Improvements

- Add `is_worker` parameter to SGLang's `init_kvcached()` to match vLLM's behavior
- Add `_tp_size` global state to SGLang interfaces for proper TP coordination
- Update SGLang patches to detect `tp_rank` and `tp_size` from `get_tp_group()`
- Add debug logging when SGLang TP detection falls back to single-GPU mode
- Reset `_tp_size` in `shutdown_kvcached()` for proper cleanup
- Add type hints to global module state variables
- Replace bare `except` clauses with specific exception handling

## Test Plan

- [ ] Verify Python syntax is valid
- [ ] Test with SGLang single-GPU inference
- [ ] Test with SGLang tensor parallel (multi-GPU) inference  
- [ ] Test with vLLM to ensure no regressions

Supersedes #226

### DIFF
--- a/kvcached/autopatch.py
+++ b/kvcached/autopatch.py
@@ -1,19 +1,41 @@
 # SPDX-FileCopyrightText: Copyright contributors to the kvcached project
 # SPDX-License-Identifier: Apache-2.0
 
+"""Automatic patching for kvcached integration with LLM serving engines.
+
+This module registers import hooks for vLLM and SGLang to enable kvcached's
+elastic KV cache management when these libraries are loaded.
+"""
+
+import logging
 from importlib import import_module
+
+logger = logging.getLogger(__name__)
 
 
 def autopatch_all() -> None:
+    """Register import hooks for all supported serving engines.
+
+    This function attempts to import the autopatch modules for vLLM and SGLang,
+    which register their respective import hooks. Failures are logged at debug
+    level since not all engines may be installed.
+    """
     # Importing these modules registers their when_imported hooks
     try:
         import_module("kvcached.integration.vllm.autopatch")
-    except Exception:
-        pass
+        logger.debug("Registered vLLM autopatch hooks")
+    except ImportError:
+        logger.debug("vLLM autopatch not available (vLLM may not be installed)")
+    except Exception as e:
+        logger.warning("Failed to register vLLM autopatch hooks: %s", e)
+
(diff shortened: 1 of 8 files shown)


rishabhsinha17:
Closing this. Rechecked each fix against current main; the TP/IPC path has moved enough that most of it has landed and the rest would now be wrong.

- Fix 1 (`start_worker_listener_thread` keyed on `tp_rank` rather than `torch.cuda.current_device()`) is in: `kvcached/integration/sglang/interfaces.py` calls `start_worker_listener_thread(tp_rank, pp_rank)`.
- Fix 2 (broadcast over `range(1, tp_size)` to skip rank 0) is no longer correct. It assumed rank 0 has no listener. On main every rank starts one when `world_size > 1`, and the vLLM path starts one per worker via `is_worker=True`, so skipping rank 0 would leave it without map/unmap.
- Fix 4 (always map locally on rank 0, then broadcast) rests on the same assumption and conflicts with the current design where rank 0 is a listener like any other rank.
- `is_worker` on the SGLang `init_kvcached()` is in as well.

Rebasing what remains onto six months of IPC changes (#313, #379, #411, #422) is not worth it against a branch this old. If any of the broadcast semantics here are still wrong on main I would rather file a fresh issue with a current repro.

Issue #236, opened 2026-02-02
ElasticTokenToKVPoolAllocator requires elastic MHA pool


wdx727:
When I start the service, the following error is thrown:
`File "srt/managers/scheduler.py", line 3357, in run_scheduler_process
  File "srt/managers/scheduler.py", line 360, in __init__
  File "srt/managers/tp_worker.py", line 245, in __init__
  File "srt/model_executor/model_runner.py", line 362, in __init__
  File "srt/model_executor/model_runner.py", line 510, in initialize
  File "srt/model_executor/model_runner.py", line 2019, in init_memory_pool
  File "/usr/local/lib/python3.12/dist-packages/kvcached/integration/sglang/patches.py", line 58, in __init__
    raise ValueError("ElasticTokenToKVPoolAllocator requires elastic MHA pool")
ValueError: ElasticTokenToKVPoolAllocator requires elastic MHA pool`

wdx727:
> Is this the same issue as issue [#235](https://github.com/ovg-project/kvcached/issues/235) ? Also SGLang 0.5.6?

Yes.

cui36:
Hi @wdx727, we have already supported the MLA model and multi page-size. Welcome to have a try!

PR #237, opened 2026-02-08
feat: support ElasticPagedTokenToKVPoolAllocator for page-size>1

cui36:
Have tested on sglang-v0.5.3 and sglang-v0.5.6, with Qwen/Qwen2-1.5B and --page-size 4, passed.


cui36:
### What is missing so far

The KVCacheManager (`self.kvcached_allocator`) is created but never called. Nobody calls `.alloc()` or `.free()` on it.

cui36:
Somehow know how to trigger the issue https://github.com/ovg-project/kvcached/issues/236. But I think that part is an independent PR.

Issue #245, opened 2026-02-21
TimeoutError during startup with Pipeline Parallelism (PP) on Qwen3-8B


iMmAseu:
I am testing Qwen3-8B using vLLM with kvcached on a 2-GPU node. When Pipeline Parallelism (PP) is enabled (e.g., PP=2, TP=1), the engine fails to initialize, raising a TimeoutError. The KVCacheManager fails to detect KV tensor creation within the 10-second threshold.

<img width="2388" height="393" alt="Image" src="https://github.com/user-attachments/assets/9226b214-8d9e-4410-9d63-7dca2a9e6ad3" />

PR #246, opened 2026-02-21
[Issue #245] Add pipeline parallelism support

iMmAseu:
## Summary

This PR fixes a TimeoutError during vLLM initialization when Pipeline Parallelism is enabled.

Fixes #243

---

## Root Cause

`KVCacheManager` only used `tensor_parallel_size`, which caused incorrect process detection in PP setups.

---

## Changes

- Fix IPC synchronization using global `world_size`
- Compute consistent `global_rank`
- Rename `tp_size` → `world_size`
- Update listener rank detection

---

## Test Plan

- [x] Syntax check
- [x] Multi GPU inference (on Qwen3 8B)
- [x] PP inference
- [x] sgl and support for other version vllm
- [x] tp=2 and pp=2

### DIFF
--- a/benchmarks/simple_bench/start_server.sh
+++ b/benchmarks/simple_bench/start_server.sh
@@ -154,7 +154,7 @@ elif [ "$engine" == "sgl" -o "$engine" == "sglang" ]; then
         export TORCHDYNAMO_DISABLE=1
         SGL_L4_ARGS="--attention-backend torch_native"
     fi
-    
+
     # SGLang uses python requests to check localhost:30000 at startup.
     # Exclude localhost from proxy so it doesn't try to route through Squid.
     export no_proxy="localhost,127.0.0.1,::1"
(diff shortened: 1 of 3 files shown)


ivanium:
Just curious does this PR support enabling both TP and PP, say TP=2 and PP=2? In that case, how do we get the consistent `global_rank` here?

cui36:
Hi @iMmAseu, thanks for the update! Let's split it into two parts, and I will go through the vllm patch.

benchmarks/simple_bench/start_client.sh
```diff
@@ -23,14 +23,15 @@ venv_path=""
 
 usage() {
     cat <<EOF
-Usage: $0 <engine> [--venv-path PATH] [--port PORT] [--model MODEL_ID]
+Usage: $0 <engine> [--venv-path PATH] [--port PORT] [--model MODEL_ID] [--dataset-path PATH]
```
cui36: Let's not change this file if unnecessary

kvcached/integration/vllm/interfaces.py
```diff
@@ -20,18 +20,19 @@
 _kvcached_initialized: bool = False
 _kvcached_device = None
 _async_sched = False
-_tp_size: int = 1
+_world_size: int = 1
```
cui36, line 23: I will take the vllm patch part.

PR #249, opened 2026-02-22
feat: support GPT-OSS in SGLang

cui36:


cui36:
@ztang2370 Thanks for the update! And for SGLang it will use the original MHA Pool, which has already worked. We can focus on the left vLLM later.We can confirm whether the Pool logic will be similar.

PR #252, opened 2026-02-24
feat: support Prefix Caching for vLLM

qinganrice:
# Add Prefix Caching Support for vLLM

## Summary

Implements to support prefix caching for vLLM, enabling KV cache block reuse across requests that share common prefixes. Includes LRU eviction with proper refcount synchronization to prevent memory leaks.

### Core Changes

**New Files:**
- `kvcached/prefix_cache_manager.py` - PrefixCacheManager with LRU eviction logic
- `benchmarks/bench_prefix_cache/` - Comprehensive test suite (3 tests, helpers, documentation)

**Modified Files:**
- `kvcached/kv_cache_manager.py` - Added refcount synchronization between physical blocks and cache entries
- `kvcached/integration/vllm/patches.py` - Integrated prefix cache into vLLM request processing
- `kvcached/integration/vllm/interfaces.py` - Pass `PREFIX_CACHE_MAX_SIZE` to KVCacheManager
- `kvcached/utils.py` - Added `PREFIX_CACHE_MAX_SIZE` configuration constant
- `benchmarks/simple_bench/start_server.sh` - Enabled prefix caching flag

## Testing

**Model:** meta-llama/Llama-3.2-1B

### Test 1: Basic Prefix Cache
- **Config:** Default cache (1000 blocks), ~500 token prefix
- **Result:** 1.08x speedup on cached requests

### Test 2: Long Prefix Speedup
- **Config:** Default cache (1000 blocks), ~2000 token prefix with 40+ examples
- **Result:** 1.2-1.5x speedup (higher with larger models)

### Test 3: LRU Eviction
- **Config:** Small cache (10 blocks), 4-5 distinct prefixes
- **Result:** 20+ successful evictions, 51 expected failures (blocks in use), eviction working correctly

## Configuration

Set before starting server:
- export KVCACHED_PREFIX_CACHE_MAX_SIZE=10  # Default: 1000
- export KVCACHED_LOG_LEVEL=DEBUG           # Optional: see cache operations

## Documentation

Complete test suite with:
- 3 test scripts (basic, Long Prefix, LRU Eviction)
- 3 automation scripts
- README with usage, troubleshooting, examples

## To Do List:

- SGLang Support: Extend prefix cache implementation to SGLang framework
- Unified Cache Manager: Merge kv_cache_manager.py and prefix_cache_manager.py into a single file that supports both vLLM and SGLang frameworks

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -19,10 +19,12 @@ if TYPE_CHECKING:
     try:
         from vllm.v1.core.block_pool import KVCacheBlock  # type: ignore[import-untyped]
         from vllm.v1.core.block_pool import KVCacheEvent  # type: ignore[import-untyped]
+        from vllm.v1.core.scheduler import Request  # type: ignore[import-untyped]
     except ImportError:
         # Fallback if vLLM is not available during type checking
         KVCacheBlock = Any  # type: ignore[misc,assignment]
         KVCacheEvent = Any  # type: ignore[misc,assignment]
+        Request = Any  # type: ignore[misc,assignment]
 
 # Version ranges for vLLM support
 VLLM_V8_RANGE = ">=0.8.4,<0.9.0"  # vLLM 0.8.x versions, need to cover 0.8.5.post1
@@ -60,10 +62,6 @@ class ElasticBlockPoolPatch(VersionAwarePatch, BasePatch):
 
         logger = self.logger  # Capture logger in closure
 
-        _PREFIX_CACHE_WARNING_MSG = (
-            "kvcached does not support prefix cache yet, "
-            "double check vLLM is not launched with prefix cache enabled.")
-
         class ElasticBlockPool(BlockPool):  # type: ignore
             """ElasticBlockPool that manages KVCacheBlocks using kvcached."""
 
@@ -77,7 +75,10 @@ class ElasticBlockPoolPatch(VersionAwarePatch, BasePatch):
                 enable_kv_cache_events: bool = False,
             ) -> None:
                 assert isinstance(num_gpu_blocks, int) and num_gpu_blocks > 0
-                assert not enable_caching, "Caching is not supported in ElasticBlockPool"
+                self.enable_prefix_cache = enable_caching
+                if enable_caching:
+                    logger.info("Prefix caching enabled for ElasticBlockPool")
+
                 assert not enable_kv_cache_events, (
                     "KV cache events are not supported in ElasticBlockPool")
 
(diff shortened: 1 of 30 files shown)


kvcached/integration/vllm/patches.py
```diff
@@ -88,20 +90,95 @@ def __init__(
...
+                Returns:
+                    List containing a single KVCacheBlock if found, None otherwise
+                """
+                if not self.enable_prefix_cache:
+                    return None
+
+                # For now, kvcached only supports single group (group_id=0)
+                # Ignore kv_cache_group_ids parameter
+                if len(kv_cache_group_ids) > 1:
+                    logger.warning(f"ElasticBlockPool only supports single KV cache group, "
```
lianghao208, line 108: why can't it suppot multiple kv cache groups?

qinganrice: Hi @lianghao208, since we only recently added support for the hybrid attention in the latest release, the prefix caching implementation in this PR only focuses on the single-group case. Support for multi-group could be added later. However, prefix caching will prevent KVCached from releasing memory after a request finishes, which conflicts with KVCached’s core design. We are currently investigating this and looking for a good balance to support both features. If you have any ideas or comments, please feel free to share. Thanks!

lianghao208: Thanks for the reply, if we simply store prefix cache by BlockHashWithGroupId, just like the implementation in vllm, is that viable or there are other concerns?
```
# vllm/v1/core/kv_cache_utils.py
# `BlockHashWithGroupId` combines a `BlockHash` with its KV cache group ID.
# It is represented as raw bytes for compactness and efficiency. The helper
# functions below pack/unpack the `BlockHash` and group id into/from the key.
BlockHashWithGroupId = NewType("BlockHashWithGroupId", bytes)
```

qinganrice: Hi @lianghao208, if you store prefix cache by BlockHashWithGroupid, KVCached will raise an error. I will sync this with the team and get back to you soon. Thanks for your patience.

lianghao208: Thanks, that makes sense, I just noticed the patches.py already filter out the multiple kv cache groups condition in kvcache initialization.
`We are currently investigating this and looking for a good balance to support both features` that's our main concerns as well.

kvcached/kv_cache_manager.py
```diff
@@ -214,15 +235,58 @@ def free(self, indices: List[int]):
...
 
+        # If prefix caching is enabled, check refcounts before freeing
+        blocks_to_free = []
+        if self.enable_prefix_cache and self.prefix_cache_manager is not None:
+            for idx in indices:
+                # First, decrement PrefixCacheManager refcount if block is cached
+                can_free_from_cache = True
+                if idx in self.prefix_cache_manager.block_to_hash:
+                    block_hash = self.prefix_cache_manager.block_to_hash[idx]
+                    can_free_from_cache = self.prefix_cache_manager.decrement_refcount(block_hash)
```
lianghao208: If can_free_from_cache is true, it means the refcount in this prefix cache block entry is already 0, and the block will be freed immediately. Is the _evict_lru() still necessary?

kvcached/kv_cache_manager.py
```diff
@@ -205,6 +221,11 @@ def _alloc(self,
             self.num_avail_blocks -= num_from_page
             remaining_need -= num_from_page
 
+        # Initialize refcount for newly allocated blocks
+        if self.enable_prefix_cache:
+            for idx in ret_index:
+                self.block_refcounts[idx] = self.block_refcounts.get(idx, 0) + 1
```
lianghao208: I think a newly allocated block's refcount has already been increased in block_pool.cache_full_blocks() --> kv_cache_manager.cache_blocks().
The whole process:
```
kvcache_manager.allocate_slots()-->coordinator.allocate_new_blocks()-->block_pool.get_new_blocks()-->coordinator.cache_blocks()-->block_pool.cache_full_blocks()-->refcount++
```
Is this refcount increment duplicated?

cui36: Hi @lianghao208 thanks for the great advice! And could we have a further discussion on our slack channel (https://join.slack.com/t/ovg-project/shared_invite/zt-3s9e1i5ad-Jpvl~1TfoxDbuTTVEftg5A) if that works for you?

PR #263, opened 2026-03-03
feat: support gpt-oss in sglang-0.5.9

ztang2370:
Support gpt-oss in sglang-0.5.9. `GptOssForCausalLM` follows `hybrid_swa_model` path https://github.com/ztang2370/sglang/blob/38ee749dd90cdab572d393bb856430ef92c981d2/python/sglang/srt/configs/model_config.py#L1320.

### DIFF
--- a/csrc/inc/allocator.hpp
+++ b/csrc/inc/allocator.hpp
@@ -25,13 +25,19 @@ public:
   ~FTensorAllocator();
 
   // KV cache interfaces.
+  // group_id allows multiple independent KV cache pools (e.g., for hybrid
+  // attention models with separate full-attention and sliding-window groups).
+  // Default group_id=0 preserves backward compatibility.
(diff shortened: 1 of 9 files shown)


csrc/inc/allocator.hpp
```diff
@@ -63,17 +84,11 @@ class FTensorAllocator {
...
-  size_t kv_tensor_size_per_layer_;
 
   mutable std::mutex mtx_;
-  // For per-layer layout: one tensor per layer
-  std::unordered_map<std::string, std::unique_ptr<FTensor>> ftensors_;
-  // For contiguous layout: single tensor containing all layers
-  std::unique_ptr<FTensor> contiguous_kv_tensor_;
-  std::shared_ptr<Page> zero_page_;
+  // Map from group_id to per-group KV cache state.
+  std::unordered_map<int64_t, KVGroup> kv_groups_;
```
ivanium, line 91: nits: if its group_id, we can just use a vector IIUC

PR #269, opened 2026-03-08
feat: support KV cache CPU offloading in vLLM

qinganrice:
**Summary:**
This PR adds support for using vLLM CPU offloading with KVCached. KV cache blocks are moved between GPU and CPU so that repeated prompts can be served from CPU cache instead of recomputing.

**Changes:**
KVConnectorMixinPatch
- Patches `KVConnectorModelRunnerMixin.use_uniform_kv_cache() `so it returns False when kvcached is enabled.
- Reason: With OffloadingConnector, vLLM’s `prefer_cross_layer_blocks=True` makes `use_uniform_kv_cache()` return True, so vLLM uses `allocate_uniform_kv_caches()` and a single cross-layer torch.zeros tensor. That skips kvcached’s `_allocate_kv_cache_tensors` path and the VMM elastic pool.
- By forcing `use_uniform_kv_cache()` to False, allocation goes through `_allocate_kv_cache_tensors`, so kvcached’s VMM-backed tensors are used and CPU offloading works.
- Applied only for vLLM >=0.12.0 via @version_range(VLLM_V12_RANGE).

**Test:**

- Runs successfully with vLLM 0.16.0. 
- GPU→CPU offload, CPU→GPU fetch (cache hit)  and CPU eviction (LRU & ARC) are tested.

### DIFF
--- a/kvcached/integration/vllm/autopatch.py
+++ b/kvcached/integration/vllm/autopatch.py
@@ -11,12 +11,14 @@ from kvcached.integration.vllm.patches import (
     VLLM_ALL_RANGE,
     VLLM_V8_RANGE,
     VLLM_V9_PLUS_RANGE,
+    VLLM_V12_RANGE,
     ElasticBlockPoolPatch,
     EngineCorePatch,
     GPUModelRunnerPatch,
     GPUWorkerPatch,
     KVCacheCoordinatorPatch,
     KVCacheManagerPatch,
+    KVConnectorMixinPatch,
 )
 from kvcached.utils import get_kvcached_logger
 
@@ -44,6 +46,7 @@ def _patch_vllm(_vllm: types.ModuleType) -> None:
             (GPUWorkerPatch(), VLLM_ALL_RANGE),
             (KVCacheCoordinatorPatch(), VLLM_V9_PLUS_RANGE),
             (KVCacheManagerPatch(), VLLM_V8_RANGE),
+            (KVConnectorMixinPatch(), VLLM_V12_RANGE),
         ]
     )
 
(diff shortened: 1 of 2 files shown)


qinganrice:
For SGLang's support, KVCached requires `--disable-radix-cache` but HiCache (extended RadixCache for CPU offloading) requires radix cache to track GPU↔CPU load/store per prefix tree node and SGLang forbids using both flags together. Therefore, we cannot support SGLang CPU offloading so far.

jiarong0907:
> For SGLang's support, KVCached requires `--disable-radix-cache` but HiCache (extended RadixCache for CPU offloading) requires radix cache to track GPU↔CPU load/store per prefix tree node and SGLang forbids using both flags together. Therefore, we cannot support SGLang CPU offloading so far.

After support prefix cache, kvcached should no long require `--disable-radix-cache`.

shipiyouniao:
Linking #361 here so we can coordinate the vLLM CPU-offloading work. This PR provides the per-layer route that keeps allocation inside KVCached's VMM path; #361 aims to support VMM-backed cross-layer allocation as well.

My preference would be to complete the combined implementation in #361: use the cross-layer layout where supported, while preserving this PR's per-layer approach as a compatibility fallback. That would avoid landing two overlapping patches to the uniform-cache decision and maintaining them separately.

The offload, cache-hit restore, and LRU/ARC eviction scenarios reported here would be a useful baseline for that consolidation. We should also verify the supported version/backend coverage and fallback behavior before saying #361 supersedes this PR. No need to close this one before that is established. Would you be happy to coordinate the remaining work in #361?

PR #300, opened 2026-04-08
Reuse KVCacheBlockClass Object for Performance Optimization

lianghao208:
We introduce a reusable kv_block_pool to prevent constant `KVCacheBlockClass `object creation, which can significantly reduce object creation overhead.


lianghao208:
@cui36 This PR needs approval for git workflows. Thanks.

PR #301, opened 2026-04-08
fix(vllm): prevent EngineCore crash when shared physical KV pool is exhausted (fix #262)

### LINKED ISSUE #262: EngineCore crashes when physical KV pool is exhausted under multi-instance load
When running multiple kvcached instances under heavy concurrent load, one instance can crash with a fatal AssertionError in  
  ElasticBlockPool.get_new_blocks().

  The root cause is that available_size() (used by vllm's scheduler to check if blocks are available) and the actual alloc()
  call are not atomic with respect to the shared physical pool. Another instance can consume physical pages between the check
  and the allocation, causing alloc() to return None and the subsequent assert block_ids is not None to fire, killing the
  EngineCore.

  **Environment**

  - GPU: AMD MI300X (192 GB)
  - vLLM: 0.14.0
  - kvcached: (repo main)
  - Setup: 6× Qwen2.5-7B-Instruct instances, kvcached_gpu_utilization=0.90
  
  **Steps to Reproduce**

  1. Launch 6 vllm instances sharing a kvcached physical pool
  2. Run a staggered load sweep with long completions (e.g. completion_len=2048, peak_rps=20) across all instances
  3. When multiple instances are simultaneously draining heavy backlogs (high KV cache usage), one instance crashes

  **Logs (see attached):**

  [kvcached][WARNING] kv_cache_manager.py:174 available_size()=71 < need_size=76
  ERROR core.py:938 EngineCore encountered a fatal error.
    ...
    scheduler.schedule()
    → kv_cache_manager.allocate_slots()
    → coordinator.allocate_new_blocks()
    → block_pool.get_new_blocks(num_new_blocks)   ← AssertionError

  **Expected behavior:**

  When alloc() returns None, the engine should handle it gracefully rather than crash.

  Notes:

  The reserved_page_list mechanism partially mitigates this race for pre-mapped pages, but does not cover cases where the
  needed allocation exceeds the reservation buffer.

<img width="1858" height="1450" alt="Image" src="https://github.com/user-attachments/assets/a0ee7e60-d944-491f-9072-3f488349a400" />

mahendrarathore1742:
## Summary

This PR fixes a crash in `ElasticBlockPool.get_new_blocks()` under multi-instance load when the shared physical KV pool is exhausted between scheduler availability check and allocation.

Previously, `kv_cache_manager.alloc()` could return `None`, and the code asserted on that result, causing a fatal `AssertionError` and crashing `EngineCore`.

## Root Cause

`available_size()` and `alloc()` are not atomic across multiple instances sharing the same physical pool.  
A race can occur where one instance consumes physical pages after another instance checks capacity but before it calls `alloc()`.

## What Changed

- Updated `ElasticBlockPool.get_new_blocks()` in `kvcached/integration/vllm/patches.py`:
  - Replaced fatal assert with graceful handling when `alloc()` returns `None`.
  - Added one retry path after opportunistic eviction from `_evictable_blocks` (when prefix cache is enabled).
  - If allocation still fails, raise `ValueError` with clear context (`requested`, `available`) instead of crashing.
  - Added a defensive check for unexpected partial allocation length.

## Why This Is Safe

- Keeps existing allocation flow and prefix-cache behavior intact.
- Only changes failure-mode handling from hard crash to controlled error.
- Retry path is bounded (single retry), so no retry loops.

## Tests

Added regression tests in `tests/test_prefix_cache.py`:

- `test_get_new_blocks_raises_when_alloc_returns_none`
- `test_get_new_blocks_retries_after_eviction_when_alloc_returns_none`

### Result

- ✅ Both new regression tests pass (`2 passed`).

## Issue

Closes #262


mahendrarathore1742:
Hi @jiarong0907 @cui36 @ztang2370  — could you review this when you get a chance? Thanks!

cui36:
@mahendrarathore1742  Thanks for the PR. Reviewing soon!

kvcached/integration/vllm/patches.py
```diff
@@ -207,7 +207,12 @@ def _make_cache_key(block_hash: Any, group_id: int) -> bytes:
                 big-endian group_id so the same content hash is distinct across
                 KV cache groups (e.g. full attention vs sliding window).
                 """
-                return bytes(block_hash) + group_id.to_bytes(4, "big", signed=False)
+                block_hash_bytes = (
+                    block_hash.encode("utf-8")
```
cui36: What's the motivation for the **str** branch here? BlockHash is NewType("BlockHash", bytes) in vLLM, so block_hash should always be bytes here?

kvcached/integration/vllm/patches.py
```diff
@@ -286,14 +291,29 @@ def get_new_blocks(
...
+
+                if block_ids is None:
+                    raise ValueError(
+                        "Unable to allocate KV cache blocks from physical pool; "
+                        f"requested={num_blocks}, available={self.kv_cache_manager.available_size()}"
+                    )
 
-                block_ids = self.kv_cache_manager.alloc(num_blocks)
-                assert block_ids is not None and len(block_ids) == num_blocks
+                if len(block_ids) != num_blocks:
```
cui36: kv_cache_manager.alloc() either returns None or a list of exactly need_size items ([kv_cache_manager.py:190-225](https://github.com/ovg-project/kvcached/blob/main/kvcached/kv_cache_manager.py#L190-L225)), so partial returns aren't part of its contract.  I think the suggestion from Copilot is good: either removing it or replacing it with an assert, which means `assert len(block_ids) == num_blocks`, maybe enough.

PR #313, opened 2026-04-21
Add vLLM NIXL PD smoke compatibility

AAbouzeid:
## Summary
- add an eager vLLM `NixlConnector` compatibility patch for kvcached PD disaggregation
- align NIXL registration with kvcached KV tensor block counts and keep kvcached NIXL on the NHD path
- add a hard guard for `KVCACHED_CONTIGUOUS_LAYOUT=true` with NIXL to avoid silent KV corruption
- add a GPU smoke script that compares plain vLLM+NIXL against kvcached+NIXL

## Validation
- `python -m pytest tests/test_vllm_nixl_compat.py -q`
- `bash -n tools/run_vllm_nixl_pd_smoke.sh`
- pod smoke: `INSTALL_VLLM=0 ./tools/run_vllm_nixl_pd_smoke.sh`
- pod semantic smoke prompts: arithmetic retrieval, ticket lookup, checksum extraction; each passed with and without kvcached

### DIFF
--- a/kvcached/integration/vllm/autopatch.py
+++ b/kvcached/integration/vllm/autopatch.py
@@ -7,6 +7,7 @@ import types
 from wrapt.importer import when_imported
 
 from kvcached.integration.patch_base import PatchManager, log_patch_results
+from kvcached.integration.vllm.nixl_compat import NixlConnectorPatch
 from kvcached.integration.vllm.patches import (
     VLLM_ALL_RANGE,
     VLLM_V8_RANGE,
@@ -38,6 +39,7 @@ def _patch_vllm(_vllm: types.ModuleType) -> None:
 
     patch_manager.register_patches_with_versions(
         [
+            (NixlConnectorPatch(), VLLM_ALL_RANGE),
             (ElasticBlockPoolPatch(), VLLM_ALL_RANGE),
             (EngineCorePatch(), VLLM_ALL_RANGE),
             (GPUModelRunnerPatch(), VLLM_ALL_RANGE),
--- a/kvcached/integration/vllm/nixl_compat.py
+++ b/kvcached/integration/vllm/nixl_compat.py
@@ -80,11 +80,21 @@ class NixlConnectorPatch(BasePatch):
             patch.logger.info("kvcached: NixlConnector layout overridden to NHD")
             return None
 
-        if not hasattr(NixlConnector, "_original_get_layout"):
-            NixlConnector._original_get_layout = (
-                NixlConnector.get_required_kvcache_layout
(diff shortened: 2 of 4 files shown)


kvcached/integration/vllm/autopatch.py
```diff
@@ -50,5 +50,64 @@ def _patch_vllm(_vllm: types.ModuleType) -> None:
...
+
+    # Bug 1: force NHD layout
+    @classmethod  # type: ignore[misc]
+    def _kvcached_layout(cls, *args, **kwargs):
+        if not enable_kvcached():
+            return NixlConnector._original_get_layout(*args, **kwargs)
+        logger.info("kvcached: NixlConnector layout overridden to NHD")
+        return None
+
+    NixlConnector._original_get_layout = NixlConnector.get_required_kvcache_layout
```
qinganrice: To avoid potential errors in cases of repeated patching (e.g., module reload during testing), suggest changing it to:
```
if not hasattr(NixlConnector, "_original_get_layout"):
    NixlConnector._original_get_layout = NixlConnector.get_required_kvcache_layout
```
Without this guard, a second invocation would save the already-patched version as _original_get_layout, causing infinite recursion when the kvcached fallback path tries to call the "original" method.

kvcached/integration/vllm/autopatch.py
```diff
@@ -50,5 +50,64 @@ def _patch_vllm(_vllm: types.ModuleType) -> None:
...
+        if not enable_kvcached():
+            return _original_register(self, kv_caches, *args, **kwargs)
+
+        from kvcached.integration.vllm.interfaces import _num_blocks_per_layer
+        if _num_blocks_per_layer > 0 and _num_blocks_per_layer != self.num_blocks:
+            logger.info(
+                "kvcached: NixlConnector num_blocks %d -> %d",
+                self.num_blocks, _num_blocks_per_layer,
+            )
+            self.num_blocks = _num_blocks_per_layer
```
qinganrice: The patch works correctly when `ratio == 1 (i.e., block_size == kernel_block_size)`, but will still trigger the assertion when `ratio > 1`, because vLLM's `self.num_blocks` has already been multiplied by `self._physical_blocks_per_logical_kv_block`, while kvcached's `_num_blocks_per_layer` is the logical (pre-ratio) count.

AAbouzeid: Addressed!

kvcached/integration/vllm/autopatch.py
AAbouzeid: Addressed!

Issue #299, opened 2026-04-08
Performance Optimization Track Issue


lianghao208:
As mentioned, our team (member: @lianghao208  @pokerfaceSad  @will-qq  @xdxd1234-bit ) has developed an elastic memory management system to solve the KV cache over-allocation problem, aligning with the kvcached project's goal. During performance benchmarking, we observed a significant performance gap between our solution and counterparts like vLLM and KVCached. 

GPU: H20 * 1
Model：Llama2 7B
vllm bench + sharegpt with random QPS

| | vLLM | ours | kvcached |
| :--- | :--- | :--- | :--- |
| Mean TTFT (ms) | 80.95 | 80.69 | 267.72 |
| P99 TTFT (ms) | 260.72 | 263.28 | 915.25 |
| Mean TPOT (ms) | 29.56 | 29.71 | 68.22 |
| P99 TPOT (ms) | 65.19 | 65.18 | 218.51 |

Through our analysis, we identified several optimization opportunities in KVCached that could further enhance performance, as detailed below:

- [ ] Redundant Object Creation Optimization #300 
- [ ] Reduce CUDA Call Overhead in available_size
- [ ] Page Allocator Migration from Python to C++
- [ ] Asynchronous Pages Release

# 1. Redundant Object Creation Optimization
Implement object pooling and reuse strategy. (e.g. **KVCacheBlockClass** instances in get_new_blocks())

# 2. Reduce CUDA Call Overhead in available_size

Eliminate expensive get_avail_physical_pages() CUDA calls during block allocation in available_size() method to minimize blocks allocation latency.

# 3. Page Allocator Migration from Python to C++
Rewrite page allocation logic from Python to high-performance C++ implementation to eliminated Python interpreter overhead and GIL contention for memory allocation operations.

# 4. Asynchronous Pages Release
Transform page release process from synchronous to asynchronous operation

This issue tracks the implementation of critical performance optimizations for the system. The current implementation has several areas where overhead can be significantly reduced to improve overall system throughput and reduce latency. These optimizations target key bottlenecks identified in production usage, particularly focusing on reducing CUDA call overhead, Python interpreter overhead, object creation overhead, and synchronous page operations.

cui36:
Big thanks to @lianghao208 and your team for this contribution!

howdareyou-greta:
[...]

PR #319, opened 2026-04-30
Performance Optimization: Page Allocator Migrates from Python to C++

lianghao208:
issue #299


cui36:
Thanks @lianghao208!

cui36:
Hi @lianghao208,

Thanks for the C++ migration. After running it, I want to propose two must-fixes and a design decision.

## Must-fix (pushed a new commit)

**1. `torch_bindings.cpp` is missing `#include <pybind11/functional.h>`** (and `<pybind11/stl.h>` for the new vector / unordered_map returns). Without it, `set_should_use_worker_ipc_callback` rejects any Python callable, and `KVCacheManager.__init__` always raises `TypeError` at `kv_cache_manager.py:111`. All 5 tests in `test_kvcache_manager.py` regress from `PASSED` on main to `ERROR` on this PR. 

**2. The test fixture in `tests/test_kvcache_manager.py` accesses Python-only attributes** that don't exist on the C++ allocator: `enable_page_prealloc`, `min_reserved_pages`, `reserved_page_list`, `num_total_pages`. Replace with `get_num_reserved_pages()` / `get_num_total_pages()`. 

## Design decision: still support `kvctl limit`? (proposed in a new PR: https://github.com/ovg-project/kvcached/pull/323)

The remaining issues all stem from one question: whether the engine should still react to external writers of the shm `total_size`:

1. `kv_cache_manager.py:230-234` comments out the resize poll.
2. C++ derives the IPC name as `kvcached_engine_<pgid>` while Python / `kvctl` use `kvcached_<engine_tag>_<pgid>`, the two sides aren't even looking at the same `/dev/shm` segment, so even if #1 were re-enabled, no signal would cross.

**If `kvctl` stays a published feature** (it has docs in `examples/02_memory_control/`, a console-script entry in `setup.py`, and a working test on `main`), all three need fixing. Cost: ~7 μs/alloc for the shm poll, but the microbench still retains ~83% of the C++ migration's speedup over `main` (alloc+free: `main` 65 μs → PR #319 23 μs → with fixes 30 μs). More optimization details in PR https://github.com/ovg-project/kvcached/pull/323.

A future optimization that keeps `kvctl` working *and* recovers the ~7 μs would be a watcher thread that polls shm at `low frequency` and exposes a single atomic flag to the alloc hot path.

Note that no matter adding kvctl or not, **passive elasticity** (unmap-on-free, physical GPU memory floating between co-located instances via CUDA VMM) is **unaffected either way**. Only the active control plane is at stake here.

lianghao208:
@cui36 Thanks for the review, they all make sense to me. I notice the PR #323 has already fixed all the correspondence problems.

csrc/page_allocator.cpp
```diff
@@ -0,0 +1,719 @@
...
+  update_memory_usage();
+  auto end_time = std::chrono::steady_clock::now();
+  auto duration = std::chrono::duration_cast<std::chrono::microseconds>(
+      end_time - start_time);
+  LOGGER(DEBUG, "alloc 1 page slow path cost %lu us", duration.count());
+
+  return std::make_shared<InternalPage>(page_id, page_size_);
+}
+
+void PageAllocator::free_page(page_id_t page_id) {
```
qinganrice, line 239: `std::lock_guard` is used, but `lock_.unlock() / lock_.lock() `is called manually later. If` unmap_pages()` throws between the manual `unlock() `and `lock()`, the guard's destructor will `unlock()` a mutex that is not locked by this thread.

csrc/page_allocator.cpp
```diff
@@ -0,0 +1,719 @@
...
+  lock_.lock();
+  free_page_list_.push_back(page_id);
+  update_memory_usage();
+  cond_.notify_all();
+}
+
+void PageAllocator::free_pages(const std::vector<page_id_t> &page_ids) {
+  auto start_time = std::chrono::steady_clock::now();
+
+  std::lock_guard<std::mutex> lock(lock_);
```
qinganrice: Same here.

kvcached/kv_cache_manager.py
```diff
@@ -346,11 +388,8 @@ def available_size(self) -> int:
         if self.in_shrink:
             blocks_from_free_pages = 0
         else:
-            virtual_free_pages = self.page_allocator.get_num_free_pages()
-            physical_free_pages = self.page_allocator.get_avail_physical_pages(
-            ) + self.page_allocator.get_num_reserved_pages()
-            free_pages = min(virtual_free_pages, physical_free_pages)
-            blocks_from_free_pages = free_pages * Page.get_num_blocks(
+            free_pages = self.page_allocator.get_num_free_pages()
```
qinganrice: The previous implementation took` min(free_virtual_pages, available_physical_pages).` The new version only counts `virtual free pages + in-page free blocks`, assuming physical memory is always available. This relies on `prealloc_worker` limiting pre-mapping to available physical memory, but slow-path allocations in alloc_page() (free page -> map -> allocate) could still fail if GPU memory is exhausted by other processes?

csrc/page_allocator.cpp
```diff
@@ -0,0 +1,719 @@
...
+    if (!reserved_page_list_.empty()) {
+      page_id = reserved_page_list_.front();
+      reserved_page_list_.pop_front();
+      num_free_pages_--;
+
+      // Trigger preallocation if reserved pool is getting low
+      if (free_page_list_.size() != 0 &&
+          reserved_page_list_.size() <
+              static_cast<size_t>(min_reserved_pages_)) {
+        max_reserved_pages_ =
```
qinganrice: When` reserved_page_list_ `drops below `min_reserved_pages_`, `max_reserved_pages_ ` is enlarged by 4 times (up to `num_total_pages_`). Once grown, freed pages permanently stay mapped in the reserved pool until `trim() `is explicitly called. It is worth noting that after a traffic spike, physical memory usage will not be decreased naturally which prevents it from being shared with other serving engines. This behavior contradicts the elastic design principles of KVCached?

lianghao208: Fixed. Thanks for pointing this out, we enlarge `max_reserved_pages_` by 4 times in out internal version, it works with another thread called `deallocator` to shrink the memory pool, which is not introduced here.

csrc/page_allocator.cpp
lianghao208: Fixed by using split scoped lock guards to keep expensive operations outside critical sections, achieving both zero overhead and exception safety.

csrc/page_allocator.cpp
lianghao208: updated

kvcached/kv_cache_manager.py
lianghao208: Rolled back the changes since they were meant for our internal setup. Because we use preemption to let high-priority jobs 'see' the full GPU capacity, we don't need to check actual physical available memory every time. The original code actually works better for what we need right now.

Issue #320, opened 2026-04-30
Autopatch fails when ENABLE_KVCACHED is set inside a Python script (requires explicit from kvcached import autopatch)


ztang2370:
## Summary

When kvcached is used from a custom Python script, users must add `from kvcached import autopatch` before `import vllm` / `import sglang` for the patches to take effect. Setting `os.environ["ENABLE_KVCACHED"]="1"` inside the script does not work, even though the documentation suggests this env var is the toggle. This is confusing and forces a kvcached-specific source-level import in user code.

Related to issue https://github.com/ovg-project/kvcached/issues/316.

## Reproduction

```
import os
os.environ["ENABLE_KVCACHED"] = "1"
os.environ["KVCACHED_AUTOPATCH"] = "1"

from vllm.engine.arg_utils import AsyncEngineArgs
from vllm.usage.usage_lib import UsageContext
from vllm.v1.engine.async_llm import AsyncLLM

engine_args = AsyncEngineArgs(
    model="tencent/HunyuanOCR",
    trust_remote_code=True,
    gpu_memory_utilization=0.7,
    max_model_len=4096,
    enable_prefix_caching=False,
    max_num_batched_tokens=8192,
    mm_processor_cache_gb=0,
)
vllm_config = engine_args.create_engine_config(usage_context=UsageContext.OPENAI_API_SERVER)

async_llm = AsyncLLM.from_vllm_config(
    vllm_config=vllm_config,
    usage_context=UsageContext.OPENAI_API_SERVER,
    stat_loggers=None,
    enable_log_requests=engine_args.enable_log_requests,
    aggregate_engine_logging=engine_args.aggregate_engine_logging,
    disable_log_stats=engine_args.disable_log_stats,
)
```

Expected: vllm is patched by kvcached.
Actual: vllm runs unpatched. Patching only happens if either:
1. `ENABLE_KVCACHED=1` is exported in the shell before launching Python, or 
2. `from kvcached import autopatch` is added to the script before `import vllm`.

## Root cause

The autopatch entry point is `kvcached_autopatch.pth`:

```
# kvcached_autopatch.pth
import os, importlib, importlib.util; (
    os.environ.setdefault("KVCACHED_AUTOPATCH", "1"),
    getattr(importlib.import_module("kvcached.autopatch"), "autopatch_all", lambda: None)()
) if os.getenv("ENABLE_KVCACHED", "false").lower() in ("true", "1")
[...]

Issue #317, opened 2026-04-28
Is it better to have consistent Env style?


jiarong0907:
Can we make the environmental virable all at the same style for the following and others that are related to kvcached?

```
export ENABLE_KVCACHED=true
export KVCACHED_AUTOPATCH=1
```

to

```
export ENABLE_KVCACHED=1
export KVCACHED_AUTOPATCH=1
``` 

Related to this issue: https://github.com/ovg-project/kvcached/issues/316

ivanium:
I remember we actually have it: https://github.com/ovg-project/kvcached/blob/c5ce8eca3442bd42907a1dbf5de772766e540992/kvcached/integration/patch_base.py#L21

jiarong0907:
> I remember we actually have it:
> 
> [kvcached/kvcached/integration/patch_base.py](https://github.com/ovg-project/kvcached/blob/c5ce8eca3442bd42907a1dbf5de772766e540992/kvcached/integration/patch_base.py#L21)
> 
> Line 21 in [c5ce8ec](/ovg-project/kvcached/commit/c5ce8eca3442bd42907a1dbf5de772766e540992)
> 
>  return os.getenv("ENABLE_KVCACHED", "false").lower() in ("true", "1")

Great! Then, it maybe just a documentation issue.

Issue #316, opened 2026-04-28
The kvcached environment variable set in the python script does not take effect


Dyouvrean:
Our model is deployed with triton. Before the tritonserver starts, environment variables are set. kvcached takes effect, but since multiple vllm models need to be started, environment variables need to be set before the model script initializes the engine. The contents of the involved documents are as follows: 
The following environment variable are set in model.py, and configured "gpu_memory_utilization": 0.2，
`import os
os.environ["ENABLE_KVCACHED"] = "1"
os.environ["KVCACHED_AUTOPATCH"] = "1"`
Tests have found that kvcached does not dynamically and automatically allocate video memory; it still allocates it based on the gpu_memory parameter.
To verify if it was triton's issue, we directly used a python script for testing and found that it was the same.
`import os
os.environ["ENABLE_KVCACHED"] = "1"
os.environ["KVCACHED_AUTOPATCH"] = "1"

from vllm.engine.arg_utils import AsyncEngineArgs
from vllm.usage.usage_lib import UsageContext
from vllm.v1.engine.async_llm import AsyncLLM

vllm_engine_config = {
    "model": "/weights/pretrain/hunyuan-ocr",
    "trust_remote_code": True,
    "gpu_memory_utilization": 0.7,
    "max_model_len": 4096,
    "enable_prefix_caching": False,
    "max_num_batched_tokens": 8192,
    "mm_processor_cache_gb": 0
}

engine_args = AsyncEngineArgs(**vllm_engine_config)

vllm_config = engine_args.create_engine_config(usage_context=UsageContext.OPENAI_API_SERVER)

async_llm = AsyncLLM.from_vllm_config(
            vllm_config=vllm_config,
            usage_context=UsageContext.OPENAI_API_SERVER,
            stat_loggers=None,
            enable_log_requests=engine_args.enable_log_requests,
            aggregate_engine_logging=engine_args.aggregate_engine_logging,
            disable_log_stats=engine_args.disable_log_stats,
        )
print(1)`
It didn't take effect. The video memory was still increased as set. How to solve it? Thanks!

jiarong0907:
Hi @Dyouvrean, thanks for reporting this issue. 

Could you quickly try

```
os.environ["ENABLE_KVCACHED"] = "true"
os.environ["KVCACHED_AUTOPATCH"] = "1"
``` 

instead?

jiarong0907:
[...]

PR #322, opened 2026-05-03
fix: honor ENABLE_KVCACHED in autopatch enable check

cui36:
To solve the issue https://github.com/ovg-project/kvcached/issues/320, https://github.com/ovg-project/kvcached/issues/317 and https://github.com/ovg-project/kvcached/issues/316, by setting kvcached using unified arg `ENABLE_KVCACHED`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,11 +175,18 @@ The documentation covers:
 
 ## Testing
 
-kvcached can be enabled by setting the following environmental variables:
+kvcached can be enabled by setting `ENABLE_KVCACHED` either in the shell:
 
 ```bash
 export ENABLE_KVCACHED=true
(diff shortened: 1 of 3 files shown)


jiarong0907:
```
def _env_enabled() -> bool:
    return os.getenv("KVCACHED_AUTOPATCH", "false").lower() in ("true", "1")
    return (
        os.getenv("ENABLE_KVCACHED", "false").lower() in ("true", "1")
        or os.getenv("KVCACHED_AUTOPATCH", "false").lower() in ("true", "1")
    )
```

Not sure this is a good design because enabling kvcached is separated from enabling autopath. I can imagine someone manually integrates kvcached with vLLM and thus doesn't need our autopatching.

PR #327, opened 2026-05-09
[Do Not Merge] P2pNcclConnector PD disaggregation

AAbouzeid:
# P2P NCCL Debug PR Description

## Summary

This PR adds debug-only instrumentation and experiment scripts for investigating a
`P2pNcclConnector` hang in PD disaggregation.

No production fix is included. This is intentionally observational: the goal is to
identify where the hang occurs and whether it is caused by kvcached or upstream vLLM
P2P NCCL behavior.

## What Changed

- Added P2P NCCL debug logging around:
  - `P2pNcclConnector.build_connector_meta`
  - `P2pNcclConnector.save_kv_layer`
  - `P2pNcclConnector.start_load_kv`
  - `P2pNcclEngine.send_tensor`
  - `P2pNcclEngine.recv_tensor`
  - ZMQ/NCCL send/recv paths
- Added a kvcached P2P debug harness:
  - `experiments/10_p2p_nccl_debug.sh`
- Added a direct upstream vLLM P2P NCCL validation harness:
  - `experiments/11_vllm_p2p_nccl_direct.sh`

## Key Finding

The hang is reproducible with kvcached disabled and also with direct upstream vLLM
P2P NCCL. It is not kvcached-specific.

The strongest signal is that the hang disappears when vLLM request ID randomization
is disabled:

```bash
VLLM_DISABLE_REQUEST_ID_RANDOMIZATION=1
```

## Experiment Matrix

| Run | Setup | Request ID Randomization | Result | Log Dir |
| --- | --- | ---: | --- | --- |
| 1 | kvcached harness, kvcached disabled | on | hangs | `logs_p2p_debug/20260509_123315` |
| 2 | kvcached harness, kvcached disabled | off | passes | `logs_p2p_debug/20260509_124428` |
| 3 | direct upstream vLLM P2P NCCL | on | hangs | `logs_vllm_p2p_direct/20260509_125509` |
| 4 | direct upstream vLLM P2P NCCL | off | passes | `logs_vllm_p2p_direct/20260509_130208` |

## Root Cause Hypothesis

`P2pNcclConnector` keys transferred tensors as:

```text
request_id#layer_name
```

However, prefill and decode run as separate vLLM engines. With request ID
randomization enabled, each engine derives a different internal
`request.request_id` from the same external `X-Request-Id`.

Observed example from the instrumented run:

Producer sends:

```text
...-b5ce4b84#model.layers.0.self_attn.attn
```

Consumer waits for:

```text
...-a08a4814#model.layers.0.self_attn.attn
```

The embedded external transfer prefix matches:

```text
___prefill_addr_<addr>___decode_addr_<addr>_<uuid>
```

but the randomized internal suffix differs, so decode waits forever for a tensor key
that producer never sends.

## Interpretation

The stall is not at:

- proxy registration
- ZMQ connection setup
- ZMQ ack
- NCCL init
- NCCL send

The instrumented kvcached run showed producer successfully sending layer tensors and
decode waiting on a missing tensor key. The direct upstream vLLM runs confirm the
same behavior without kvcached in the path.

## Scope

This PR does not fix the issue.

It intentionally avoids:

- tensor layout changes
- `.contiguous()` changes
- block count rewrites
- timeouts
- exceptions
- production behavior changes

## Next Step

The likely fix is for `P2pNcclConnector` to key P2P tensor transfers by a stable
transfer request ID, derived from the external embedded request ID, rather than the
per-engine randomized internal vLLM request ID.

## Logs attached below
[p2p_nccl_request_id_evidence_20260509.tar.gz](https://github.com/user-attachments/files/27552015/p2p_nccl_request_id_evidence_20260509.tar.gz)

### DIFF
--- a/experiments/.gitignore
+++ b/experiments/.gitignore
@@ -1,0 +1,1 @@
+logs_p2p_debug/
--- a/experiments/10_p2p_nccl_debug.sh
+++ b/experiments/10_p2p_nccl_debug.sh
@@ -14,6 +14,7 @@
 # Useful overrides:
 #   MODEL=Qwen/Qwen2.5-1.5B-Instruct ./experiments/10_p2p_nccl_debug.sh
 #   TIMEOUT_REQUEST=300 KVCACHED_P2P_WAIT_LOG_INTERVAL_S=5 ./experiments/10_p2p_nccl_debug.sh
+#   RUN_WITH_KVCACHED=0 ./experiments/10_p2p_nccl_debug.sh
 #   KEEP_ALIVE_ON_FAIL=1 ./experiments/10_p2p_nccl_debug.sh
 #
 # Logs are saved to experiments/logs_p2p_debug/<timestamp>/.
@@ -51,6 +52,7 @@ P2P_MEM_POOL_SIZE_GB="${P2P_MEM_POOL_SIZE_GB:-32}"
 
 TIMEOUT_STARTUP="${TIMEOUT_STARTUP:-240}"
 TIMEOUT_REQUEST="${TIMEOUT_REQUEST:-90}"
+RUN_WITH_KVCACHED="${RUN_WITH_KVCACHED:-1}"
 INSTALL_DEPS="${INSTALL_DEPS:-1}"
 INSTALL_KVCACHED="${INSTALL_KVCACHED:-1}"
 SKIP_MODEL_DOWNLOAD="${SKIP_MODEL_DOWNLOAD:-0}"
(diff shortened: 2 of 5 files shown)


cui36:
Thanks @AAbouzeid! Since it’s not an issue on our side, I think we can leave it for now and add support later once vLLM has better support for it.

ivanium:
@AAbouzeid feel free to try the Nixl connector in vLLM which is better maintained :)

cui36:
> @AAbouzeid feel free to try the Nixl connector in vLLM which is better maintained :)

Yeah we've already supported in https://github.com/ovg-project/kvcached/pull/313

PR #328, opened 2026-05-09
fix: defer autopatch gate to import time without eager kvcached import

cui36:
The .pth used to read ENABLE_KVCACHED at site-init, so users who set the env var inside their Python script before importing vllm/sglang got no autopatch — the gate had already declined. Move the gate into wrapt's when_imported callbacks so it runs at the moment vllm/sglang is imported, by which time in-script env mutations are already visible.

Avoid importing kvcached (and transitively torch) from .pth itself: only find_spec the package and register lazy callbacks. The actual integration module is loaded inside the callback, after vllm/sglang has pulled in torch on its own — so non-vllm Python processes pay ~16ms instead of ~570ms at startup.

Keeps _env_enabled() unchanged (single KVCACHED_AUTOPATCH variable) and preserves the ENABLE_KVCACHED user-facing name via a bridge inside the callback, so existing shell-export users see no behavior change.


cui36:
@ztang2370 I think we can only change the pth file to solve the bug?

ztang2370:
Yes I think this works.

PR #330, opened 2026-05-13
Fix illegal memory access caused by missing _kernel_block_sizes attribute in some vLLM versions

lianghao208:
In vLLM 0.16, gpu_model_runner does not have the `self._kernel_block_sizes` attribute. As a result, kernel_block_sizes = getattr(self, "_kernel_block_sizes", None) returns None, which causes kvcached to incorrectly assume ratio = 1. This leads to an excessively large stride value being computed, ultimately resulting in an illegal memory access error.

For example:

| Variable | Value |
|----------|-------|
| Model | Qwen3.5-9B hybrid architecture |
| kv_cache_group[0,1,2] | MambaSpec |
| kv_cache_group[3] | FullAttentionSpec, layers=8, block_size=528 |
| num_kv_heads | 4 |
| head_size | 256 |
| page_size_bytes | 2,162,688 > 2MB |
| virtual block size | 528 |
| FA3 kernel_block_size | 16 (determined by get_supported_kernel_block_sizes()) |

| Variable | Incorrect Value | Correct Value |
|----------|----------------|---------------|
| kernel_block_size | 528 | 16 |
| ratio | 1 | 33 (= 528 ÷ 16) |
| kernel_kvcache_shape | [2, **2699**, 528, 4, 256] | [2, **89067**, 16, 4, 256] |
| kvcached's perspective | One FA3 kernel block holds 528 tokens, up to **2,699** blocks | One FA3 kernel block holds 16 tokens, up to **89,067** blocks |
| vllm's perspective | One FA3 kernel block holds 16 tokens, up to **89,067** blocks | One FA3 kernel block holds 16 tokens, up to **89,067** blocks |
| block dimension stride | **1,081,344 illegal memory access!!!** (=528\*4\*256\*2) | 32,768 (=2\*16\*4\*256) |

This PR adds proper handling for the case where _kernel_block_sizes is not defined, ensuring the stride is computed correctly across different vLLM versions.

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -1061,6 +1061,7 @@ class GPUModelRunnerPatch(VersionAwarePatch, BasePatch):
 
         # Capture patch instance for version-aware access
         patch_instance = self
+        logger = self.logger  # Capture logger in closure
 
         def _allocate_kv_cache_from_kvcached(self, kv_cache_config):
             import torch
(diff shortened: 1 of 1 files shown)


cui36:
Hi @lianghao208, The code looks correct, but I don't think kvcached should manually reconstruct the kernel block size from `attn_backend_cls`. vLLM 0.16 already has the source of truth in `self._prepare_kernel_block_sizes(kv_cache_config)`, and newer versions expose `prepare_kernel_block_sizes(kv_cache_config, self.attn_groups)`. Maybe we can reuse those paths?

lianghao208:
@cui36 Agree, reusing `prepare_kernel_block_sizes` reduces the risk of breaking changes during future vLLM upgrades.

PR #333, opened 2026-05-19
fix: handle str block_hash in prefix cache key encoding

daniel620:
### `_make_cache_key` TypeError fix
`_make_cache_key` calls `bytes(block_hash)` which raises `TypeError` when vLLM
passes a string hash (e.g. hex digest). Added an `isinstance(block_hash, str)`
check to encode str to bytes first. Bytes input path is unchanged.

### Tests out of sync with null_block allocation
PR #319 (`c95e1c4`, *Page Allocator Migrates from Python to C++*) introduced a
null_block placeholder allocation in `ElasticBlockPool.__init__`:

```python
_null_ids = self.kv_cache_manager.alloc(1)
self.null_block = self.kv_block_pool[_null_ids[0]]
```

This consumes one block during pool initialization, but the test expectations in
`test_prefix_cache.py` were not updated to account for it. The tests were not
caught because the CI workflow does not run pytest.

Affected tests and their adjustments:
- `pool_factory(N)` → `pool_factory(N+1)` in 7 tests where the pool is fully filled
- `test_get_usage`: adjusted expected values to account for the null_block
- `test_reset_frees_evictable_blocks`: fixed `_block_id_to_hash` (renamed to `_block_id_to_key`)

## Test plan
- [x] `pytest tests/test_kvcache_manager.py` — 4 passed, 1 skipped
- [x] `pytest tests/test_shm_info_tracker.py` — 8 passed
- [x] `pytest tests/test_prefix_cache.py` — 26 passed
- [x] `pre-commit run --all-files` — all 12 hooks passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -349,6 +349,8 @@ class ElasticBlockPoolPatch(VersionAwarePatch, BasePatch):
                 big-endian group_id so the same content hash is distinct across
                 KV cache groups (e.g. full attention vs sliding window).
                 """
+                if isinstance(block_hash, str):
+                    block_hash = block_hash.encode()
                 return bytes(block_hash) + group_id.to_bytes(4, "big", signed=False)
 
             def get_cached_block(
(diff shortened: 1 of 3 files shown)


cui36:
LGTM. And made a small change: Moved `_make_cache_key` from a nested staticmethod to a module-level function and added a GPU-free unit test for it. This lets the str-hash fix be tested directly without vLLM, a GPU, or the compiled extension (it previously had no test), at no performance cost since module-level dispatch is if anything slightly faster than the prior self-bound staticmethod call.

Issue #339, opened 2026-05-21
TP2 vLLM startup requires two kvcached fixes before the server becomes ready


shipiyouniao:
## Summary

When kvcached is enabled on the latest main, a TP2 vLLM startup can hang before the API server becomes ready.

The process stays alive and model loading proceeds, but the server port never starts listening until two separate TP startup defects are addressed.

This does not reproduce when running the same TP2 startup configuration without kvcached.

## Environment

- kvcached: latest main
  - base commit: 0d4d581dc5b17ea8f07a83a9f9cf5345ba24478b
- vLLM: 0.18.1
- GPU: NVIDIA GeForce RTX 4090
- OS: Linux
- Model: Qwen3-8B-FP8
- tensor_parallel_size: 2
- kv cache mode: kvcached autopatch enabled

## Reproduction Setup

This was reproduced on a direct TP2 startup path without relying on controller readiness logic.

Environment variables:

```bash
ENABLE_KVCACHED=true
KVCACHED_AUTOPATCH=1
KVCACHED_IPC_NAME=kvcached_hotfix_probe3
NCCL_CUMEM_HOST_ENABLE=0
VLLM_SERVER_DEV_MODE=1
CUDA_VISIBLE_DEVICES=0,1
```

vLLM startup configuration:

```text
model: /root/offload-lab/local-models/Qwen3-8B-FP8
port: 19113
gpu-memory-utilization: 0.35
kv-cache-memory-bytes: 536870912
max-model-len: 4096
kv-cache-dtype: fp8_e4m3
enable-sleep-mode: true
enable-prefix-caching: false
disable-log-stats: true
tensor-parallel-size: 2
max-num-seqs: 64
```

## Actual Behavior

With kvcached enabled, the TP2 process stays alive but `/health` remains unreachable until both of the following defects are fixed:

1. The coordinator path can initialize kvcached with `world_size=1` even though EngineCore already knows `tensor_parallel_size=2`.
2. After correcting that world size, TP startup can still hang when the background prealloc thread starts too early and races the first null-block allocation on the multi-process map path.

Observed progression during debugging:

- bare TP2 without kvcached starts and `/health` returns 200
- TP2 + kvcached on latest main hangs before port listen
- fixing the coordinator world size moves startup past the earlier `world_size=1` allocator setup point but does not yet make the server ready
- fixing the prealloc timing on top of that makes `/health` return 200 and the server starts normally

### Root cause 1: coordinator reads TP world size too early

[...]

PR #341, opened 2026-05-21
fix: defer tp prealloc until first alloc

shipiyouniao:
## Summary

Defer the background prealloc thread until after the first null-block allocation completes in the TP multi-process startup path.

Related issue: #339

## Root cause

After the TP coordinator world size is corrected, vLLM still performs one early `KVCacheManager.alloc(1)` to reserve its real null block.

In the TP multi-process path, starting the background prealloc thread before that first alloc finishes can send both flows through the same multi-process page-map path during startup.

That race can leave the server alive but never ready, even though worker IPC sockets and direct map/unmap operations are otherwise healthy.

## Changes

- Detect the multi-process TP/worker-IPC startup path where vLLM will reserve the null block from its first normal alloc
- Defer `start_prealloc_thread()` until after that first alloc completes
- Keep eager prealloc startup for the existing single-process / reserve-null-block flows
- Add a regression test that verifies multi-process startup defers prealloc until the first alloc, while single-process startup keeps eager prealloc

## Why this approach

This is narrower than globally disabling preallocation and matches the verified behavior from the TP2 reproduction.

The goal is not to remove preallocation, but to prevent it from racing the first null-block allocation in the one startup path where that ordering matters.

## Validation

- Added unit test: `python -m pytest tests/test_tp_prealloc_startup.py -q`
- Remote TP2 reproduction with both fixes applied reached:
  - `Starting vLLM server on http://0.0.0.0:19113`
  - `Application startup complete`
  - `GET /health -> 200`

## Notes

This PR is intentionally split from the coordinator `world_size` fix. The TP2 startup issue turned out to be a two-defect chain, and this PR addresses the second one.

### DIFF
--- a/kvcached/kv_cache_manager.py
+++ b/kvcached/kv_cache_manager.py
@@ -168,6 +168,16 @@ class KVCacheManager:
             except Exception as e:
                 logger.warning("Failed to set up broadcast callbacks: %s. Falling back to single-process mode.", e)
 
+        # In multi-process mode, vLLM reserves its null block from the first
+        # post-init alloc() call rather than via reserve_null_block=True.
+        # Starting the background prealloc thread before that first alloc can
+        # race the same multi-process map path and stall TP startup.
+        self._defer_prealloc_until_first_alloc = (
+            not self.reserve_null_block
+            and (self.world_size > 1 or use_worker_ipc)
+        )
+        self._prealloc_started = False
+
         self.num_avail_blocks = 0  # Only count free blocks in avail_pages
         self.avail_pages: Dict[int, InternalPage] = {}
         self.full_pages: Dict[int, InternalPage] = {}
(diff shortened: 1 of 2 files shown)


shipiyouniao:
Closing this because #422 has superseded the startup-order workaround in this PR on the current main path.

Hardware validation used current `main` (`666671bd`) plus #340 (`e7fbe7d` merge result), without #341:

- 2x RTX 4090, TP=2
- vLLM 0.22.1, Qwen3-8B-FP8
- kvcached autopatch enabled
- page preallocation remained enabled
- allocator initialized with `world_size=2`
- three independent cold starts reached `/health = 200` in 140s, 85s, and 76s
- every run logged `Application startup complete` with no startup hang

This indicates that #422's GIL-safe preallocation/worker-IPC changes cover the second startup defect without deferring preallocation until the first allocation. #340 remains the separate TP world-size fix.

@RixinLiu, could you please update the roadmap to mark #341 as covered by #422?

RixinLiu:
updated

PR #356, opened 2026-06-08
feat: Triattention support

iMmAseu:
## Summary

This PR adds optional external TriAttention integration for kvcached without vendoring the TriAttention source tree.

TriAttention is loaded from a separately installed checkout and is enabled only when `ENABLE_TRIATTENTION=1` is set. Existing kvcached users are unaffected by default.

## What Changed

- Add an external TriAttention loader under `kvcached.integration.triattention_external`.
- Add optional TriAttention activation paths for both vLLM and SGLang autopatching.
- Fail fast when `ENABLE_TRIATTENTION=1` is requested but TriAttention activation fails.
- Add a kvcached compatibility patch for the upstream TriAttention repository.
- Add benchmark scripts for comparing kvcached-only vs kvcached+TriAttention:
  - `tools/compare_vllm_triattention.py`
  - `tools/compare_sglang_triattention.py`
- Add a SGLang `decode-stress` benchmark profile, since SGLang TriAttention compression is decode-time driven.

## Differences between vLLM and SGLang Workloads 

The vLLM benchmark uses a long-prompt workload because the vLLM integration can reclaim KV blocks through `block_pool.free_blocks(...)`, which is routed through kvcached's `ElasticBlockPool`. As a result, reclaim can directly reduce physical KV memory pressure and show up clearly in peak GPU memory.

SGLang behaves differently. Its TriAttention compression path is decode-time driven, so a very long prompt alone mostly creates a prefill-time memory peak before compression has a chance to help. For SGLang, this PR uses a `decode-stress` workload: short prompt, long forced decode, `min_tokens=max_tokens`, and `ignore_eos=true`. This makes compression and reclaim happen while decode memory is growing, which better exposes the effect in GPU peak memory.

## Benchmark Results

Model: Qwen3-8B  
GPU memory sampled with `nvidia-smi`.

### vLLM Long-Context Workload

Workload:

- prompt repeat: 1200
- max output tokens: 2048
- concurrencies: 4, 8, 16
- KV budgets: 1024, 2048

| Concurrency | KV Budget | kvcached Peak | kvcached+TriAttention Peak | Reduction | Compression Events | Freed Blocks | E2E Change |
|---:|---:|---:|---:|---:|---:|---:|---:|
| 4 | 1024 | 35669 MiB | 23329 MiB | 12340 MiB / 34.6% | 175 | 6157 | -3.6% |
| 4 | 2048 | 35669 MiB | 23621 MiB | 12048 MiB / 33.8% | 175 | 5901 | +3.7% |
| 8 | 1024 | 45893 MiB | 23905 MiB | 21988 MiB / 47.9% | 331 | 11644 | -23.9% |
| 8 | 2048 | 45893 MiB | 25061 MiB | 20832 MiB / 45.4% | 331 | 11132 | -26.0% |
| 16 | 1024 | 45893 MiB | 25633 MiB | 20260 MiB / 44.1% | 641 | 23103 | -31.5% |
| 16 | 2048 | 45893 MiB | 27367 MiB | 18526 MiB / 40.4% | 639 | 22542 | -31.1% |

The vLLM path shows a large peak GPU memory reduction, especially at higher concurrency, because reclaimed KV blocks are returned through kvcached's physical allocator path.

### SGLang Decode-Stress Workload

Workload:

- workload profile: `decode-stress`
- prompt repeat: 64
- prompt tokens: 2010
- max tokens: 8192
- min tokens: 8192
- ignore EOS: true
- concurrency: 4
- KV budget: 1024

| Mode | Success | TTFT Mean | E2E Mean | GPU Peak | Compression Events | Skipped Events | Freed Blocks |
|---|---:|---:|---:|---:|---:|---:|---:|
| kvcached | 4/4 | 230.482 ms | 76296.729 ms | 25570 MiB | 0 | 0 | 0 |
| kvcached + TriAttention | 4/4 | 238.556 ms | 72940.810 ms | 21910 MiB | 16 | 4 | 34316 |

SGLang peak GPU memory was reduced by 3660 MiB, or 14.3%.

These happen when a request has already been released by the time a late compression attempt is observed. The successful compression and reclaim path is confirmed by compression_events=16 and freed_blocks=34316.

## Validation
pre-commit run --all-files
Passed.

## Additional checks:

vLLM benchmark CSV parsed successfully.
SGLang decode-stress benchmark CSV parsed successfully.
SGLang benchmark script syntax check passed.

### DIFF
--- a/tools/compare_vllm_triattention.py
+++ b/tools/compare_vllm_triattention.py
@@ -332,11 +332,15 @@ def stream_chat_request(
 
     end = time.perf_counter()
     ttft = (first_chunk or end) - start
+    # A stream that opens and closes but yields no content tokens is a failure,
+    # not a success: e.g. the engine crashed mid-decode. Counting it as a
+    # success would report a (misleadingly low) GPU peak for a dead run.
     return RequestMetrics(
-        ok=True,
+        ok=chunks > 0,
         ttft_ms=ttft * 1000.0,
         e2e_ms=(end - start) * 1000.0,
         output_chunks=chunks,
+        error="" if chunks > 0 else "stream completed but produced 0 output tokens",
     )
 
 
--- a/tools/compare_vllm_triattention.py
+++ b/tools/compare_vllm_triattention.py
@@ -1,0 +1,908 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright contributors to the kvcached project
+# SPDX-License-Identifier: Apache-2.0
+
+"""Compare long-context vLLM serving with and without TriAttention.
+
+This script starts two kinds of vLLM servers repeatedly:
+
+* kvcached-only baseline: ENABLE_KVCACHED=true, ENABLE_TRIATTENTION=0
+* kvcached + TriAttention: ENABLE_KVCACHED=true, ENABLE_TRIATTENTION=1
+
+For each run it sends a concurrent long-context streaming workload, samples GPU
+memory, parses the server log for TriAttention activity, and writes a CSV row.
+It intentionally uses only the Python standard library so it can run in the same
+venv as vLLM without extra client dependencies.
+"""
+
+from __future__ import annotations
+
+import argparse
+import csv
+import json
+import math
+import os
+import re
+import signal
+import statistics
+import subprocess
+import tempfile
+import threading
+import time
+import urllib.error
+import urllib.request
+from collections import Counter
+from concurrent.futures import ThreadPoolExecutor, as_completed
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+CSV_FIELDS = [
+    "label",
+    "workload",
+    "model",
+    "prompt_repeat",
+    "kv_budget",
+    "divide_length",
+    "max_tokens",
+    "concurrency",
+    "requests",
+    "successes",
+    "failures",
+    "success_rate",
+    "elapsed_secs",
+    "ttft_mean_ms",
+    "ttft_p50_ms",
+    "ttft_p99_ms",
+    "tpot_mean_ms",
+    "e2e_mean_ms",
+    "e2e_p50_ms",
+    "output_chunks",
+    "output_chunks_per_sec",
+    "gpu_mib_before",
+    "gpu_mib_peak",
+    "gpu_mib_avg",
+    "gpu_mib_p95",
+    "gpu_mib_after",
+    "gpu_mib_delta_peak",
+    "compression_events",
+    "compression_skipped_events",
+    "compression_skip_reasons",
+    "free_blocks_events",
+    "freed_blocks",
+    "triattention_activation_seen",
+    "triattention_activation_failures",
+    "error_count",
+    "log_path",
+]
+
+
+PROMPT_UNIT = (
+    "Segment {idx}: kvcached maps KV pages; TriAttention compresses long KV "
+    "cache using sparse stats, budget, window, and block reclaim."
+)
+
+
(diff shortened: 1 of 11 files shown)


shipiyouniao:
My main concern is the maintenance boundary here. I don't think kvcached should take responsibility for managing third-party components or maintaining their engine integration patches. TriAttention can be useful alongside kvcached, but that doesn't necessarily mean its integration belongs in this repository.

I realize this is opt-in rather than a mandatory dependency. Still, the current setup asks users to clone TriAttention, apply a patch shipped by kvcached, and then install it. That leaves this project maintaining compatibility across TriAttention and the inference engines, beyond its KV memory-management responsibilities.

Could you reconsider where this integration should live? I'd prefer the TriAttention-specific patches, setup, and benchmarks to live in TriAttention or a separate adapter. If that integration needs a missing, generally useful allocation/reclamation hook from kvcached, we can discuss that smaller change here. Documentation linking to a validated integration would also be reasonable.

I'd like us to agree on that boundary before spending more time on the implementation details.

tools/compare_vllm_triattention.py
```diff
@@ -0,0 +1,971 @@
...
+    ttft = (first_chunk or end) - start
+    # NOTE: for this latency study we count a request that streamed to completion
+    # (no HTTP/transport error) as ok, so its e2e/throughput is measured even when
+    # it yielded no parseable content (gpt-oss harmony output is corrupted by
+    # compression but the engine still decodes the full token budget). A truly dead
+    # engine raises and is caught above (ok=False). output_chunks still records
+    # whether any content was emitted. ttft is unreliable when chunks==0 (falls back
+    # to e2e); rely on e2e for the latency comparison.
+    return RequestMetrics(
+        ok=True,
```
shipiyouniao, line 356: [P1] Don't count errored or incomplete streams as successful requests

`ok=True` only means that reading the HTTP body did not raise. I exercised this function with (1) only `data: [DONE]`, (2) an SSE `error` object followed by `[DONE]`, and (3) one content delta followed by EOF with no completion marker. All three return `ok=True`; their `output_chunks` are 0, 0, and 1. `build_row()` then includes them in successes and latency statistics, so a broken run can report 100% success.

Could you reject SSE errors, require a valid completion signal, and distinguish empty output from successful generation for this text workload? Please add fixtures for these three cases plus a valid completed stream. Failed-stream timing can still be retained separately for diagnosis. Also, SSE chunks are not necessarily tokens, so token-based throughput/TPOT should use actual token counts or be explicitly labeled as chunk-based.

shipiyouniao (changes requested):
One concrete benchmark issue worth fixing regardless of where the integration ends up living. My earlier comment was about the maintenance boundary, not a conclusion that TriAttention cannot be useful with kvcached. I'll leave the integration direction to that discussion and am happy to help review code quality.

The vLLM benchmark currently counts empty, errored, and truncated streams as successful requests. I reproduced this against the current function with three mocked HTTP streams; this is a CPU parser check, not a GPU/model test. Details below.

The comment also explicitly allows compression-corrupted output into the latency comparison. Keeping raw timing for diagnosis is fine, but please separate it from successful serving results and provide a baseline quality evaluation alongside the memory/latency numbers. Lossy compression need not produce identical text, but lower latency alone cannot establish a useful result.

jiarong0907:
We tried the triattention at that time to see whether it really worked. It turned to be working only at very specific scenarios. Because it is not a general feature, I agree that kvcached should not include triattention for now.

RixinLiu:
convert this pr to draft now

PR #357, opened 2026-06-09
fix map_to_kv_tensors/unmap_from_kv_tensors in MLA non-contiguous layout

lianghao208:
In MLA models with a non-contiguous layout, `map_to_kv_tensors` and `unmap_from_kv_tensors` perform redundant mapping for both K and V individually. Under the MLA architecture, this double-mapping approach results in a 50% waste of memory.

### DIFF
--- a/csrc/allocator.cpp
+++ b/csrc/allocator.cpp
@@ -45,8 +45,9 @@ static inline size_t get_v_base_offset(const torch::Tensor &tensor) {
 
 FTensorAllocator::FTensorAllocator(const torch::Device &device,
                                    bool contiguous_layout)
-    : dev_(device), num_layers_(0), contiguous_layout_(contiguous_layout),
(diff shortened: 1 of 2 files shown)


shipiyouniao:
Thanks for fixing the non-contiguous MLA double mapping. Could we add focused regression coverage for the new C++ branch before merging?

The current tests cover MLA shape/alignment at the Python layer, but none exercises `FTensorAllocator` with `contiguous_layout=false` and `num_kv_buffers=1`. A regression test should map and unmap an offset and verify that MLA performs one physical mapping per layer, while the existing MHA/GQA `num_kv_buffers=2` path still performs two. A mapped-page or physical-memory assertion would also cover the reported 50% waste.

Since the behavior depends on allocator state persisted by `create_kv_tensors()`, a Python-only shape test would not catch this regression.

PR #361, opened 2026-06-15
feat: support uniform cross-layer KV cache layout for CPU offloading

lianghao208:
#269 introduces a approach for supporting CPU offloading in kvcached, it forces vLLM to fall back on per-layer KV cache management rather than a uniform, all-layer strategy, which results in implicit extra overhead. 
This PR proposes a new approach that allows CPU offloading services to utilize the kvcached VMM API to allocate cross-layer contiguous memory layout. This effectively preserves system elasticity without sacrificing performance.

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -1331,6 +1331,269 @@ class GPUModelRunnerPatch(VersionAwarePatch, BasePatch):
         setattr(GPUModelRunner, "_reshape_kv_cache_tensors", _patched_reshape_kv)
         return True
 
+    @version_range(VLLM_V9_PLUS_RANGE)
+    def patch_use_uniform_kv_cache(self, GPUModelRunner) -> bool:
+        """Patch use_uniform_kv_cache to return False when kvcached cannot support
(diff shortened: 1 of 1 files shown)


shipiyouniao:
Could we coordinate this with #269 and bring the two approaches together here? My preference is to complete the vLLM offloading integration in #361, with VMM-backed cross-layer allocation where supported and a per-layer fallback for the cases it cannot handle.

#269 addresses an important compatibility issue: the native uniform allocation path can bypass KVCached's VMM tensors. This PR aims to retain the cross-layer layout without that bypass. These look like two paths of the same integration, rather than two independent patches we should stack on `use_uniform_kv_cache()`.

Before treating this as a replacement for #269, could we make the supported vLLM versions/backends and fallback conditions explicit, then cover both paths with tests and real GPU offload/restore validation? In particular, CPU cache hits, eviction, output correctness, and the kernel-block-size fallback would be useful to check. A same-workload comparison against the per-layer path would also help establish the performance benefit.

I would keep #269 open until that coverage is demonstrated, but consolidate the implementation and follow-up work here if both authors are happy with that direction.

PR #363, opened 2026-06-24
fix(vllm): reset block hash on prefix-cache eviction

### LINKED ISSUE #362: vLLM prefix cache blocks keep stale block_hash after eviction
## Summary

When kvcached's vLLM prefix-cache block pool evicts or explicitly frees cached blocks, it removes the block from kvcached's own cache indexes and returns the block id to the kvcached allocator. However, the vLLM `KVCacheBlock` object can keep its cached `block_hash` value.

## Problem

A returned block object may later be reused for a different prefix. If its old `block_hash` is still present, vLLM block state and kvcached's prefix-cache indexes can diverge. In vLLM versions where `block_hash` is write-once until `reset_hash()` is called, recaching the reused block can also fail or keep stale state.

## Expected behavior

Whenever kvcached returns a cached or uncached block to the allocator through eviction, explicit eviction, immediate uncached free, or prefix-cache reset, the associated vLLM block hash state should be cleared before the block can be reused.

## Impact

This is a prefix-cache correctness issue. It is most visible when the same `KVCacheBlock` object is evicted and later reallocated for a different prefix.

shipiyouniao:
## Summary

Fixes stale or missing vLLM `KVCacheBlock.block_hash` metadata when kvcached caches, evicts, or frees prefix-cache blocks. The compatibility path covers the writable property used by older supported vLLM versions and the read-only property plus `set_block_hash()` introduced in vLLM 0.24.

Closes #362.

## Root Cause

`ElasticBlockPool` replaces vLLM's `BlockPool.cache_full_blocks`, so it must preserve the surrounding vLLM manager's block metadata contract even though kvcached uses its own cache indexes. Cached blocks need their current hash recorded, and evicted or freed blocks need both hash fields reset before reuse.

Without this lifecycle, a block can retain stale identity across prefixes. Conversely, leaving the hash unset also breaks vLLM consumers outside the replaced block-pool implementation.

## Concrete Benefits

- **Mamba same-step correctness:** `MambaManager` uses `block.block_hash` to prevent a request from reusing another request's just-cached block before its recurrent/SSM state is materialized. An unset hash bypasses this guard and can produce incorrect output.
- **NIXL PD transfer efficiency:** the decode side uses `block.block_hash is None` to determine which blocks still need transfer. An unset hash causes an already-held prefix to be transferred again. In a real 1P1D NIXL run, preserving this metadata reduced a repeated-request transfer from 45.6 MB to 1.6 MB (96.6%).

## Changes

- add compatibility helpers that use `set_block_hash()` when available, fall back to the legacy writable property, and use `reset_hash()` when available
- reset both `_block_hash` and `_block_hash_num_tokens` for older or mock objects without `reset_hash()`
- reset hashes on lazy-cache eviction, uncached immediate free, explicit `evict_blocks()`, and `reset_prefix_cache()`
- make the main test block match vLLM 0.24's read-only property and `set_block_hash()` API
- retain a regression test for the older writable-property fallback
- cover eviction, reset, explicit eviction, reallocation, and recache flows

## Validation

- `python -m pytest tests/test_prefix_cache.py -q` -> 29 passed
- complete `pre-commit run --all-files` under Python 3.11 -> all hooks passed
- manual mypy matrix under Python 3.9, 3.10, 3.11, 3.12, and 3.13 -> all passed
- CI containers used standard `runc` with GPU devices explicitly hidden

The PR remains one focused commit.

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -274,6 +274,31 @@ def _make_cache_key(block_hash: Any, group_id: int) -> bytes:
     return bytes(block_hash) + group_id.to_bytes(4, "big", signed=False)
 
 
+def _reset_block_hash(block: Any) -> None:
+    """Clear vLLM's cached block hash before returning a block to kvcached."""
+    reset_hash = getattr(block, "reset_hash", None)
+    if callable(reset_hash):
+        reset_hash()
+        return
+    if hasattr(block, "_block_hash"):
+        block._block_hash = None
+    if hasattr(block, "_block_hash_num_tokens"):
+        block._block_hash_num_tokens = None
+
+
+def _set_block_hash(block: Any, key: Any) -> None:
+    """Set vLLM's cached block hash across API versions.
+
+    vLLM 0.24 and later expose a read-only property plus set_block_hash().
+    Older supported versions expose a writable block_hash property instead.
+    """
+    set_block_hash = getattr(block, "set_block_hash", None)
+    if callable(set_block_hash):
+        set_block_hash(key)
+    else:
+        block.block_hash = key
+
+
 class ElasticBlockPoolPatch(VersionAwarePatch, BasePatch):
     """Inject ElasticBlockPool into vLLM's block pool module"""
 
@@ -461,6 +486,12 @@ class ElasticBlockPoolPatch(VersionAwarePatch, BasePatch):
                     if key in self._cached_blocks:
                         continue
 
+                    # ElasticBlockPool tracks cached blocks through its own maps,
+                    # but vLLM manager code may still read KVCacheBlock.block_hash
+                    # after cache_full_blocks. Preserve that metadata contract and
+                    # clear it before the block is evicted or reused.
+                    if getattr(block, "block_hash", None) is None:
+                        _set_block_hash(block, key)
                     self._cached_blocks[key] = block
                     self._block_id_to_key[block.block_id] = key
 
@@ -471,10 +502,11 @@ class ElasticBlockPoolPatch(VersionAwarePatch, BasePatch):
                 """
                 ids_to_free: list[int] = []
                 for _ in range(min(num_to_evict, len(self._evictable_blocks))):
-                    bid, _block = self._evictable_blocks.popitem(last=False)
+                    bid, block = self._evictable_blocks.popitem(last=False)
                     key = self._block_id_to_key.pop(bid, None)
                     if key is not None:
                         self._cached_blocks.pop(key, None)
+                    _reset_block_hash(block)
                     ids_to_free.append(bid)
                 if ids_to_free:
                     self.kv_cache_manager.free(ids_to_free)
@@ -557,6 +589,7 @@ class ElasticBlockPoolPatch(VersionAwarePatch, BasePatch):
                             self._evictable_blocks[block.block_id] = block
                         else:
                             # Uncached block (e.g. partial): free immediately
+                            _reset_block_hash(block)
                             uncached_to_free.append(block.block_id)
                 if uncached_to_free:
                     self.kv_cache_manager.free(uncached_to_free)
@@ -576,10 +609,13 @@ class ElasticBlockPoolPatch(VersionAwarePatch, BasePatch):
                 for bid in block_ids:
                     key = self._block_id_to_key.pop(bid, None)
                     if key is not None:
-                        self._cached_blocks.pop(key, None)
+                        block = self._cached_blocks.pop(key, None)
+                        if block is not None:
+                            _reset_block_hash(block)
                         removed += 1
                     if bid in self._evictable_blocks:
-                        self._evictable_blocks.pop(bid)
+                        block = self._evictable_blocks.pop(bid)
+                        _reset_block_hash(block)
                         ids_to_free.append(bid)
 
                 if ids_to_free:
@@ -593,6 +629,8 @@ class ElasticBlockPoolPatch(VersionAwarePatch, BasePatch):
 
                 # Free all evictable blocks back to kvcached
                 if self._evictable_blocks:
+                    for block in self._evictable_blocks.values():
+                        _reset_block_hash(block)
                     ids_to_free = list(self._evictable_blocks.keys())
                     self._evictable_blocks.clear()
                     self.kv_cache_manager.free(ids_to_free)
--- a/tests/test_prefix_cache.py
+++ b/tests/test_prefix_cache.py
@@ -10,6 +10,7 @@ to test the prefix cache logic in isolation without GPU or vLLM dependency.
 
 import sys
 import types
+from typing import Optional
 from unittest import mock
 
 # ---------------------------------------------------------------------------
@@ -51,6 +52,21 @@ class MockKVCacheBlock:
         self.block_id = block_id
         self.ref_cnt = ref_cnt
         self.is_null = False
+        self._block_hash: Optional[object] = None
+        self._block_hash_num_tokens: Optional[int] = None
+
+    @property
+    def block_hash(self):
+        return self._block_hash
+
+    def set_block_hash(self, value, num_tokens=None):
+        assert self._block_hash is None
+        self._block_hash = value
+        self._block_hash_num_tokens = num_tokens
+
+    def reset_hash(self):
+        self._block_hash = None
+        self._block_hash_num_tokens = None
 
 
 class MockKVCacheManager:
@@ -86,6 +102,18 @@ class MockRequest:
 
     def __init__(self, block_hashes: list):
         self.block_hashes = block_hashes
+
+
+def test_set_block_hash_supports_legacy_writable_property():
+    """vLLM before 0.24 uses a writable block_hash property."""
+    from kvcached.integration.vllm.patches import _set_block_hash
+
+    block = types.SimpleNamespace(block_hash=None)
+    key = b"legacy-key"
+
+    _set_block_hash(block, key)
+
+    assert block.block_hash == key
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


kvcached/integration/vllm/patches.py
```diff
@@ -460,6 +469,8 @@ def cache_full_blocks(
                     if key in self._cached_blocks:
                         continue
 
+                    if getattr(block, "block_hash", None) is None:
+                        block.block_hash = key
```
lianghao208: Do we need to set `block.block_hash` in ElasticBlockPool? If we don't set this attribute, is `_reset_block_hash` still necessary?

shipiyouniao: Yes, `ElasticBlockPool` needs to set `block.block_hash` because it replaces vLLM's `BlockPool.cache_full_blocks` and does not delegate to the upstream implementation. In upstream vLLM, `cache_full_blocks` sets `KVCacheBlock.block_hash` with the group-aware cache key before inserting the block into the prefix cache map.

vllm/v1/core/block_pool.py:273-281 (vllm 0.22.1)

```python
assert blk.block_hash is None
block_hash = new_block_hashes[i]

# Update and added the full block to the cache.
block_hash_with_group_id = make_block_hash_with_group_id(
    block_hash, kv_cache_group_id
)
blk.block_hash = block_hash_with_group_id
self.cached_block_hash_to_block.insert(block_hash_with_group_id, blk)
```

The reset is also part of the same lifecycle. vLLM evicts a cached block by reading `block.block_hash`, removing it from `cached_block_hash_to_block`, and then resetting the hash:

vllm/v1/core/block_pool.py:382-390

```python
if block_hash is None:
    # The block doesn't have hash, eviction is not needed
    return False

if self.cached_block_hash_to_block.pop(block_hash, block.block_id) is None:
    # block not found in cached_block_hash_to_block,
    # eviction is not needed
    return False

block.reset_hash()
```

`KVCacheBlock` also documents that `_block_hash` is the cached identity of a full cached block:

vllm/v1/core/kv_cache_utils.py:123-125

```python
# The hash key (block hash + group id) of the block, only available
# when the block is full and cached.
_block_hash: BlockHashWithGroupId | None = None
```

So `_reset_block_hash()` is not independent dead code. It is paired with the assignment in `ElasticBlockPool.cache_full_blocks()`. Once ElasticBlockPool initializes `block.block_hash`, it must clear that field before returning/evicting the block, otherwise a reused block may retain stale prefix-cache identity or fail a later block_hash assignment.

I can add a short comment to make this lifecycle explicit.

lianghao208: kvcached replaces the whole vLLM BlockPool (including BlockPool.cache_full_blocks) and it seems `block.block_hash` is not used in any BlockPool functions after the replacement? Thus, setting `KVCacheBlock.block_hash` in kvcached is still necessary?

lianghao208: vllm/v1/core/block_pool.py:273-281 (vllm 0.22.1) ------replaced by kvcached(not using block.block_hash)
vllm/v1/core/block_pool.py:382-390 ------ no need to replace(not using block.block_hash)
If we don't initialize `block.block_hash` in ElasticBlockPool, then there is no need to `_reset_block_hash()`?

shipiyouniao: If we only looked at `ElasticBlockPool`'s own internal maps, `block.block_hash` would indeed be redundant because `ElasticBlockPool` uses `_block_id_to_key` and `_cached_blocks` for eviction.

The reason I still think we should preserve this field is that `KVCacheBlock.block_hash` is part of the contract expected by vLLM code outside `BlockPool`. For example, vLLM 0.22.1's Mamba manager reads it to avoid reusing blocks cached in the same step:

vllm/v1/core/single_type_kv_cache_manager.py:948-957

```python
assert isinstance(self.kv_cache_spec, MambaSpec)
if (
    len(new_computed_blocks) > 0
    and new_computed_blocks[-1].block_hash in self.cached_blocks_this_step
):
    # Mamba can't rely on blocks generated by other requests in the current step
```

And after caching blocks it asserts the field is initialized:

vllm/v1/core/single_type_kv_cache_manager.py:1108-1116

```python
num_cached_blocks_after = self.num_cached_block.get(request.request_id, 0)
if num_cached_blocks_after > num_cached_blocks_before:
    for block in self.req_to_blocks[request.request_id][
        num_cached_blocks_before:num_cached_blocks_after
    ]:
        if block.is_null:
            continue
        assert block.block_hash is not None
        self.cached_blocks_this_step.add(block.block_hash)
```

This came up while testing Qwen3.5-Next / hybrid-linear attention compatibility against the latest kvcached path. I understand `example 08` currently keeps hybrid-linear handling conservative, but if kvcached replaces vLLM's `BlockPool`, it is safer to preserve the `KVCacheBlock` metadata contract. Otherwise this path can break as soon as hybrid/linear support starts exercising the manager-level `block_hash` assumptions.

So this is not for `ElasticBlockPool`'s own lookup path. It is to keep the replacement compatible with the surrounding vLLM manager logic. Since we set the field on cached blocks, we should also reset it before those blocks are reused.

lianghao208: I see, that makes sense.:)

kvcached/integration/vllm/patches.py
```diff
@@ -273,6 +273,15 @@ def _make_cache_key(block_hash: Any, group_id: int) -> bytes:
     return bytes(block_hash) + group_id.to_bytes(4, "big", signed=False)
 
 
+def _reset_block_hash(block: Any) -> None:
+    """Clear vLLM's cached block hash before returning a block to kvcached."""
+    reset_hash = getattr(block, "reset_hash", None)
+    if callable(reset_hash):
+        reset_hash()
+    elif hasattr(block, "_block_hash"):
+        block._block_hash = None
```
lianghao208, line 284: should we also reset `_block_hash_num_tokens`?

vllm/v1/core/kv_cache_utils.py#L162
```
    def reset_hash(self):
        """Reset the block hash when the block is evicted."""
        self._block_hash = None
        self._block_hash_num_tokens = None
```

shipiyouniao: Good catch. I just updated `_reset_block_hash()` to reset both fields when `reset_hash()` is unavailable, and extended the test mock accordingly.

RixinLiu:
Thanks for the fix — the direction (restoring the `KVCacheBlock.block_hash` metadata contract) is right and addresses two real issues. One blocking compatibility point, plus two smaller asks:

## 1. `block.block_hash = key` crashes on vLLM ≥ 0.24.0 — the write needs to be capability-probed

vLLM removed the `block_hash` property setter in **0.24.0** and replaced it with a `set_block_hash()` method, so `block.block_hash = key` now raises `AttributeError: property 'block_hash' has no setter` (verified at runtime on a real 0.24.0 `KVCacheBlock`). This fires inside `cache_full_blocks` the moment prefix caching caches the first full block. Exact boundary:

| vLLM | `KVCacheBlock.block_hash` | `block.block_hash = key` |
|---|---|---|
| ≤ 0.23.0 | property **with** setter | ✅ works |
| ≥ 0.24.0 | **read-only** property + `set_block_hash()` | ❌ AttributeError |

A hard switch to `block.set_block_hash(key)` would break the *other* end — kvcached's version gate goes down to 0.8.4, where `set_block_hash()` doesn't exist. So neither hard-write covers the supported range; the write should probe, symmetric with the `_reset_block_hash` helper this PR already adds:

```python
def _set_block_hash(block: Any, key: Any) -> None:
    """Set vLLM's cached block hash, compatibly across the supported vLLM range.

    vLLM >=0.24.0 made block_hash a read-only property (write via set_block_hash());
    vLLM <=0.23.0 has a writable block_hash setter. Mirrors _reset_block_hash.
    """
    set_hash = getattr(block, "set_block_hash", None)
    if callable(set_hash):
        set_hash(key)
    else:
        block.block_hash = key
```

and in `cache_full_blocks`:

```python
                    if getattr(block, "block_hash", None) is None:
                        _set_block_hash(block, key)   # was: block.block_hash = key
```

(Keep the existing `is None` guard and the `_reset_block_hash` calls — they're required, since `set_block_hash()` asserts the hash is None, so a reused block must be reset before it can be set again.)

## 2. The unit tests can't catch this — `MockKVCacheBlock` has a setter that real 0.24.0 doesn't

`tests/test_prefix_cache.py`'s `MockKVCacheBlock` defines a `@block_hash.setter`, so `block.block_hash = key` passes in the tests even though real 0.24.0 rejects it — that's why all 28 tests are green while the engine would crash. Making the mock faithful to 0.24.0 (read-only `block_hash` property + a `set_block_hash()` method + `reset_hash()`) would exercise the same write path as production and guard against this regression.

## 3. Please state the concrete benefit in the description

The description ("vLLM manager code may still read block_hash … preserve the contract") is a bit abstract. Naming the two consumers and what breaks without the write would help reviewers:

- **Mamba same-step guard (correctness):** `MambaManager` fills `cached_blocks_this_step` from `block.block_hash` (`single_type_kv_cache_manager.py`); unset → the guard never fires → a request can reuse another request's just-cached block whose recurrent/SSM state isn't materialized yet → silent wrong output.
- **NIXL PD prefix skipping (efficiency):** the decode side sizes the P→D transfer from `get_unhashed_block_ids_all_groups()` (`kv_cache_manager.py`), which filters on `block.block_hash is None`; unset → D re-pulls the entire prefix it already holds. Measured on a real 1P1D nixl run: a repeated request went 45.6 MB → 1.6 MB (−96.6%) once `block_hash` is set.

shipiyouniao:
Thanks, addressed in `4ae8f893`:

- added a capability-probed `_set_block_hash()` that uses `set_block_hash()` on vLLM 0.24+ and falls back to the writable property on older supported versions;
- changed the main `MockKVCacheBlock` to the v0.24 read-only-property API and added a separate legacy fallback regression;
- expanded the PR description with the Mamba same-step correctness and NIXL PD transfer benefits.

Revalidated with 29 prefix-cache tests, full Python 3.11 pre-commit, and the Python 3.9-3.13 mypy matrix. All passed.

PR #365, opened 2026-06-24
fix: roll back partial KV allocation on page miss

### LINKED ISSUE #364: KVCacheManager does not roll back partial allocation when alloc_page fails
## Summary

`KVCacheManager.alloc()` can partially consume reserved blocks or blocks from existing pages before it needs to request another physical page. If `PageAllocator.alloc_page()` then raises `RuntimeError` because the shared physical pool is exhausted, the partial allocation is not rolled back.

## Problem

This is a race/failure-mode issue under multi-instance pressure. `available_size()` can report enough logical capacity, but another instance may consume physical pages before the current instance reaches `alloc_page()`. In that case, the current allocation attempt can fail after it has already removed block ids from reserved blocks or page free lists.

## Expected behavior

If a physical page allocation fails after partial block allocation, `KVCacheManager.alloc()` should roll back the partial block ids and return `None` as an allocation miss. The scheduler or caller can then apply its existing allocation-miss handling instead of leaking blocks or crashing.

## Impact

This complements the existing EngineCore allocation-miss handling. It protects the lower-level allocator path where the failure originates, especially when multiple kvcached instances share the same physical KV pool.

shipiyouniao:
## Summary

Fixes #364.

Fixes a lower-level allocation failure path in `KVCacheManager.alloc()` where `PageAllocator.alloc_page()` can raise after some block ids have already been reserved for the current allocation request.

## Root Cause

`available_size()` and physical page allocation are not atomic across multiple instances sharing the same physical pool. A caller may pass the capacity check, consume reserved blocks or blocks from an existing page, and then fail when requesting a new physical page.

Before this change, that `RuntimeError` escaped without rolling back the partially allocated block ids.

## What Changed

- Catch `RuntimeError` around `page_allocator.alloc_page()`.
- Return `None` as an allocation miss instead of crashing the allocation path.
- Roll back any partially allocated block ids before returning `None`.
- Use an internal `_free(..., release_empty_pages=False)` rollback path so a partial-allocation rollback does not unmap/release physical pages that were not intentionally returned to the physical pool.
- Keep normal public `free()` behavior unchanged for real releases.
- Add unit coverage for:
  - reserved-block partial allocation rollback;
  - existing-page partial allocation rollback;
  - failure before any partial allocation.

## Tests

```bash
python -m pytest tests/test_kvcache_manager_alloc_miss.py -q
python -m py_compile kvcached/kv_cache_manager.py tests/test_kvcache_manager_alloc_miss.py
```

### DIFF
--- a/kvcached/kv_cache_manager.py
+++ b/kvcached/kv_cache_manager.py
@@ -283,8 +283,25 @@ class KVCacheManager:
 
         while remaining_need > 0:  # Allocate the remaining blocks from pages
             if not self.avail_pages:
-                page = self.page_allocator.alloc_page()
-                page.init(self.block_mem_size)
+                try:
+                    page = self.page_allocator.alloc_page()
+                    page.init(self.block_mem_size)
+                except RuntimeError as exc:
+                    logger.warning(
+                        "Physical KV page allocation failed; returning "
+                        "allocation miss instead of crashing scheduler: %s",
+                        exc,
+                    )
+                    if ret_index:
+                        try:
+                            self._free(ret_index, release_empty_pages=False)
+                        except Exception as rollback_exc:
+                            logger.warning(
+                                "Failed to roll back partially allocated KV "
+                                "blocks after page allocation miss: %s",
+                                rollback_exc,
+                            )
+                    return None
                 # A page may have zero usable blocks when block_mem_size is
                 # large (e.g. HYBRID_LINEAR) and every aligned block would
                 # straddle the page boundary. Park it in full_pages so it's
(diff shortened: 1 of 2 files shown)


shipiyouniao:
Superseded by #430. Its rollback restores both page blocks and the reservation ledger, releases pages created by the failed allocation, and the latest commit narrows recovery to the intended alloc_page()/page.init() failure path. The remaining CI failure on #430 is only the new test-classification check for tests/test_alloc_rollback.py, not a behavioral regression.

PR #366, opened 2026-06-29
Fix autopatch env handling for script-set variables

### LINKED ISSUE #320: Autopatch fails when ENABLE_KVCACHED is set inside a Python script (requires explicit from kvcached import autopatch)

cmssmc00:
Fixes #320.
Related to #316.

Summary:

* Allow `ENABLE_KVCACHED` to enable vLLM/SGLang autopatching, in addition to `KVCACHED_AUTOPATCH`.
* Keep autopatching disabled by default.
* Update the disabled-path log message to mention both env vars.
* Add regression tests for script-set `ENABLE_KVCACHED` and `KVCACHED_AUTOPATCH` behavior.

Tests:

* `PYTHONPATH=$PWD python -m pytest tests/test_autopatch_env.py -q`

### DIFF
--- a/kvcached/integration/sglang/autopatch.py
+++ b/kvcached/integration/sglang/autopatch.py
@@ -23,13 +23,16 @@ logger = get_kvcached_logger()
 
 
 def _env_enabled() -> bool:
-    return os.getenv("KVCACHED_AUTOPATCH", "false").lower() in ("true", "1")
+    return (
+        os.getenv("ENABLE_KVCACHED", "false").lower() in ("true", "1")
+        or os.getenv("KVCACHED_AUTOPATCH", "false").lower() in ("true", "1")
+    )
 
 
 @when_imported("sglang")
 def _patch_sglang(_sglang: types.ModuleType) -> None:
     if not _env_enabled():
-        logger.debug("Disabled by KVCACHED_AUTOPATCH")
+        logger.debug("Disabled by ENABLE_KVCACHED/KVCACHED_AUTOPATCH")
         return
 
     # Create patch manager and register version-specific SGLang patches
--- a/kvcached/integration/vllm/autopatch.py
+++ b/kvcached/integration/vllm/autopatch.py
@@ -25,13 +25,16 @@ logger = get_kvcached_logger()
 
 
 def _env_enabled() -> bool:
-    return os.getenv("KVCACHED_AUTOPATCH", "false").lower() in ("true", "1")
+    return (
+        os.getenv("ENABLE_KVCACHED", "false").lower() in ("true", "1")
+        or os.getenv("KVCACHED_AUTOPATCH", "false").lower() in ("true", "1")
+    )
 
 
 @when_imported("vllm")
 def _patch_vllm(_vllm: types.ModuleType) -> None:
     if not _env_enabled():
-        logger.debug("Disabled by KVCACHED_AUTOPATCH")
+        logger.debug("Disabled by ENABLE_KVCACHED/KVCACHED_AUTOPATCH")
         return
 
     # Create patch manager and register version-specific vLLM patches
--- a/tests/test_autopatch_env.py
+++ b/tests/test_autopatch_env.py
@@ -1,0 +1,122 @@
+# SPDX-FileCopyrightText: Copyright contributors to the kvcached project
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import subprocess
+import sys
+import textwrap
+from pathlib import Path
+from unittest import mock
+
+
+# Mock heavy dependencies so this regression test can run without torch/CUDA.
+# This must happen before importing kvcached.
+_torch_mock = mock.MagicMock()
+_torch_mock.__version__ = "2.6.0"
+_torch_mock.version.hip = None
+_torch_mock.version.cuda = None
+sys.modules.setdefault("torch", _torch_mock)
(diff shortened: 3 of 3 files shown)


shipiyouniao:
Thanks for looking into this. It looks like the original #320 case is already covered by #328, which moved the enable check to engine-import time. That implementation is still on main, and this was also noted when #320 was closed: https://github.com/ovg-project/kvcached/issues/320#issuecomment-4828701698.

There is one behavior difference worth keeping separate: with `ENABLE_KVCACHED=1` and `KVCACHED_AUTOPATCH=0`, main currently leaves automatic patching disabled. The OR condition here would enable it despite the explicit `0`. So this is not just an additional regression test for the original issue; it also changes the opt-out behavior.

Could you close this PR if there is no remaining reproduction on current main? If you have found a separate case that #328 does not cover, please share it and we can discuss that scope and the intended flag precedence. Either way, I would avoid keeping `Fixes #320` as the justification for changing those semantics.

Issue #371, opened 2026-06-30
Page-prealloc deadlock: a co-located kvcached engine hangs in `Init C++ PageAllocator` and never finishes startup


jiarong0907:
## Summary

With `KVCACHED_PAGE_PREALLOC_ENABLED=true` (the default), a kvcached vLLM engine can **deadlock during KV-pool initialization**: the process prints

```
Init C++ PageAllocator: ... enable_prealloc=1, ...
```

and then **hangs forever** in that C++ prealloc step — it never reaches uvicorn (`Application startup complete`), never binds `/health`, and the unit sits idle in `futex_wait_queue` (kernel futex wait) with no further output and no error.

We hit this in two situations, both removed by disabling prealloc:

1. **Two co-located engines (init deadlock).** Start a first kvcached engine (it comes up fine), then start a **second** kvcached engine on the same GPU while the first is running. The **second** engine deadlocks in `Init C++ PageAllocator`. The first is unaffected. Reproduced directly in our logs (details below).
2. **A single engine (runtime stall).** For the hybrid linear-attention coder `Qwen/Qwen3.6-27B-FP8`, a *single* kvcached engine with prealloc enabled stalls under sustained saturated load: the KV pool reports healthy/near-full but decode never makes progress and clients time out (600 s). Same fix.

**Disabling prealloc (`KVCACHED_PAGE_PREALLOC_ENABLED=false`) fixes both** — the engine maps KV pages on demand and starts/serves normally. Prealloc is the only changed variable.

## Environment

| Component | Value |
|---|---|
| GPU | NVIDIA RTX PRO 6000 Blackwell, 96 GB, `sm_120` |
| CUDA toolkit / driver | 12.8 / 13.0 |
| vLLM | 0.19.0 (V1, `VLLM_USE_V1=1`) |
| torch | 2.10.0+cu128 |
| flashinfer | 0.6.6 (`VLLM_ATTENTION_BACKEND=FLASHINFER`) |
| kvcached | 0.1.5, built from source at upstream `main` incl. PR #367 / commit `fea40de` |
| kvcached mode | plugin autopatch (`ENABLE_KVCACHED=true`, `KVCACHED_AUTOPATCH=1`) |
| kvcached util | `KVCACHED_GPU_UTILIZATION=0.95` |

[...]

Issue #334, opened 2026-05-19
vLLM startup hangs after PageAllocator init on single-GPU world_size=1 with kvcached enabled


shipiyouniao:
# vLLM startup hangs after PageAllocator init on single-GPU world_size=1 with kvcached enabled

## Summary

When kvcached is enabled on the latest `main`, a single-GPU vLLM startup can hang after KV cache / PageAllocator initialization.

The process does not exit, but the API server port never starts listening, so `/health` and `/v1/models` remain unreachable.

This does **not** reproduce when running the same vLLM startup configuration without kvcached.

## Environment

- kvcached: latest `main`
  - commit: `0d4d581dc5b17ea8f07a83a9f9cf5345ba24478b`
- vLLM: `0.18.1`
- GPU: `NVIDIA GeForce RTX 4090`
- OS: `Linux`
- Model: `Qwen3-8B-FP8`
- tensor_parallel_size: `1`
- world_size observed in logs: `1`

## Reproduction Setup

I can reproduce this with a direct single-GPU startup, without relying on any controller readiness logic.

Environment variables:

```bash
ENABLE_KVCACHED=true
KVCACHED_AUTOPATCH=1
KVCACHED_IPC_NAME=kvcached_vllm_GPU2
VLLM_SERVER_DEV_MODE=1
CUDA_VISIBLE_DEVICES=2
```

vLLM startup configuration:

```text
model: /mnt/test/models/Qwen3-8B-FP8
port: 12346
gpu-memory-utilization: 0.49
kv-cache-memory-bytes: 536870912
max-model-len: 6144
kv-cache-dtype: fp8_e4m3
enable-sleep-mode: true
enable-prefix-caching: false
enforce-eager: true
disable-log-stats: true
```

## Actual Behavior

The process stays alive, GPU memory increases to about `10.8 GiB`, but the API server never starts listening.

Observed symptoms:

- `http://127.0.0.1:12346/health` stays connection refused
- `http://127.0.0.1:12346/v1/models` stays connection refused
- the process does not crash
- there is no explicit Python exception in the startup log
- the log consistently stops after PageAllocator initialization

Relevant log tail:

```text
Loading weights took 120.46 seconds
Model loading took 8.8 GiB memory and 121.099412 seconds
GPU KV cache size: 7,280 tokens
Maximum concurrency for 6,144 tokens per request: 1.18x
init engine (profile, create kv cache, warmup model) took 1.01 seconds
[...]

PR #377, opened 2026-07-02
Fix unsigned underflow in available physical page calculation

cmssmc00:
Partially addresses #371.

Summary:
- Avoid unsigned underflow when subtracting GPU utilization headroom from available physical memory.
- Return 0 usable physical memory when available memory is below the configured headroom.
- Prevent the prealloc path from seeing an artificially huge available-page count under memory pressure.

Tests:
- git diff --check

### DIFF
--- a/.github/workflows/pre-commit.yml
+++ b/.github/workflows/pre-commit.yml
@@ -14,6 +14,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
+      - name: Run GPU-free C++ unit tests
+        run: bash tests/cpp/run_tests.sh
+
       - uses: actions/setup-python@v5
         with:
           python-version: '3.11'  # or match your local Python version
--- a/csrc/inc/math_utils.hpp
+++ b/csrc/inc/math_utils.hpp
@@ -1,0 +1,14 @@
+// SPDX-FileCopyrightText: Copyright contributors to the kvcached project
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstddef>
+
+namespace kvcached {
+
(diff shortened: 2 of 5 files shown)


shipiyouniao:
The saturating subtraction looks correct. Could we add a regression test for the boundary that caused the unsigned underflow?

It would be useful to cover `available > headroom`, `available == headroom`, and `available < headroom`, with the latter two producing zero usable pages. The PR currently records only `git diff --check`, so the exact `size_t` boundary can regress without CI noticing. If the GPU query makes this difficult to test directly, extracting the saturating subtraction into a small helper would allow a GPU-free unit test.

cmssmc00:
Thanks for the feedback. I extracted the saturating subtraction into a GPU-free constexpr helper and added regression tests covering available > headroom, available == headroom, and available < headroom. The latter two cases return zero. I also added the GPU-free test to CI.

RixinLiu:
Thanks for the fix! Now i will merge it into main branch.

#377 fixes an unsigned-underflow bug in the allocator's physical-memory accounting: when free GPU memory drops below the configured headroom — exactly the situation of a co-located second instance — the subtraction wraps around and the allocator believes it has ~2^40 pages available, so allocation and preallocation charge ahead on memory that does not exist instead of refusing honestly and returning 0.

That false balance is the amplifier in #371's co-located deadlock: it pushes the main thread into blocking C++ paths where it holds the GIL forever while the prealloc thread waits for the GIL to call back into Python, the same deadlock family as #334, whose single-process leg #336 already fixes by caching the worker-IPC decision, and whose full cure still requires releasing the GIL in the blocking bindings.

PR #380, opened 2026-07-02
fix(vllm): floor max cached blocks at 1 for sub-block-size budgets (#343)

### LINKED ISSUE #343: _get_max_cached_blocks truncates small positive KVCACHED_MAX_CACHED_TOKENS to the disabled

​```python
​```

SuperMarioYL:
## Summary

Fixes #343.

`_get_max_cached_blocks` (`kvcached/integration/vllm/patches.py`) converts the
unified `KVCACHED_MAX_CACHED_TOKENS` config into a block budget with:

```python
return MAX_CACHED_TOKENS // block_size
```

When `KVCACHED_MAX_CACHED_TOKENS` is a **small positive** value below one KV
block (e.g. `8` with a `16`-token `block_size`), integer division floors to `0`.
That is the exact value the `== 0` **"disabled — evict on free"** sentinel
produces, so a user who asked for a small non-zero cache gets prefix caching
**silently turned off** instead of a small cap.

## Fix

Handle the sentinels explicitly and floor the positive branch at one block:

```python
if MAX_CACHED_TOKENS < 0:
    return -1          # unlimited (unchanged)
if MAX_CACHED_TOKENS == 0:
    return 0           # disabled  (unchanged)
max_cached_blocks = MAX_CACHED_TOKENS // block_size
if max_cached_blocks == 0:
    logger.warning(...)  # no longer silent
    return 1
return max_cached_blocks
```

- A positive-but-sub-block budget now keeps prefix caching **enabled** at the
  smallest viable unit (1 block), which the consumer at
  `ElasticBlockPool` (`max_cached_blocks >= 0 and len(evictable) > max_cached_blocks`)
  treats as a valid cap of one evictable block.
- The clamp is **no longer silent** — a `warning` reports the effective
  granularity and points at `KVCACHED_MAX_CACHED_TOKENS=0` for explicit disable.
- `< 0` (unlimited), `== 0` (disabled), and `>= block_size` budgets are
  **byte-for-byte unchanged**; the only behavioral change is the previously
  broken `0 < tokens < block_size` case.

## Tests

Adds `tests/test_get_max_cached_blocks.py` — GPU-free (no `vmm_ops`, GPU, or
installed vLLM), mirroring the existing `tests/test_make_cache_key.py`. Covers
the `-1 / 0 / sub-block / exact-multiple / non-multiple / large` boundaries, the
`#343` regression (`0 < tokens < block_size` never collapses to the disabled
sentinel), and the warn-on-clamp / no-warn-on-full-block behavior.

```
$ pytest tests/test_get_max_cached_blocks.py -q
12 passed
```

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -13,7 +13,7 @@ from typing import TYPE_CHECKING, Any, Iterable, Optional
 
 from kvcached.integration.patch_base import BasePatch, enable_kvcached
 from kvcached.integration.version_utils import VersionAwarePatch, VersionRange, version_range
-from kvcached.utils import KVCachedConfigError
+from kvcached.utils import KVCachedConfigError, get_kvcached_logger
 
 if TYPE_CHECKING:
     # These types are imported from vLLM at runtime via getattr()
@@ -27,6 +27,9 @@ if TYPE_CHECKING:
         KVCacheBlock = Any  # type: ignore[misc,assignment]
         KVCacheEvent = Any  # type: ignore[misc,assignment]
         Request = Any  # type: ignore[misc,assignment]
+
+
+logger = get_kvcached_logger()
 
 
 def _is_attention_spec(spec: Any) -> bool:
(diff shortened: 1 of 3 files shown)


SuperMarioYL:
Gentle ping — this has been green for a while. Happy to rebase or address any feedback.

RixinLiu:
LGTM.

Reviewed this and #405, which fixes the same issue #343 in different ways. I will merge this one.

The reason is illustrated in below table:
| `MAX_CACHED_TOKENS` | `block_size` | #380 | #405 | requested ceiling |
|---:|---:|---|---|---:|
| 8 | 16 | 1 block = 16 tok (+warning) | 1 block = 16 tok | 8 |
| 16 | 16 | 1 block = 16 tok | 1 block = 16 tok | 16 |
| 17 | 16 | 1 block = 16 tok | **2 blocks = 32 tok** | 17 |
| 31 | 16 | 1 block = 16 tok | **2 blocks = 32 tok** | 31 |
| 32 | 16 | 2 blocks | 2 blocks | 32 |

PR #383, opened 2026-07-05
fix: make FTensor page map operations idempotent

### LINKED ISSUE #381: FTensor map/unmap should tolerate idempotent page operations
### Problem

`FTensor::map()` currently returns `false` when the target page is already mapped, and `FTensor::unmap()` returns `false` when the target page is already unmapped.

That makes repeated map/unmap requests indistinguishable from real VMM failures at the caller boundary. In runtime paths that coordinate page mapping across workers, duplicate requests can be benign but still surface as hard failures.

### Expected behavior

Mapping an already-mapped page and unmapping an already-unmapped page should be idempotent no-ops. Real VMM failures should still be reported by the underlying `mem_map` / `mem_unmap` calls.

shipiyouniao:
### Summary  Fixes #381.  Make `FTensor::map()` and `FTensor::unmap()` tolerate idempotent page operations:  - mapping an already-mapped page returns success - unmapping an already-unmapped page returns success - real VMM map/unmap failures still propagate through the existing `CHECK_GPU` paths  This avoids treating duplicate runtime page-map coordination messages as hard failures.  ### Status

Withdrawn after maintainer and internal review. This was a preventive DFX finding from code inspection, not a reproduced runtime failure. We could not establish a legitimate duplicate map/unmap path, and silently accepting duplicates could hide an upper-layer bookkeeping error. The separate partial-transaction problem remains tracked by #412/#418.

### DIFF
--- a/csrc/ftensor.cpp
+++ b/csrc/ftensor.cpp
@@ -102,8 +102,7 @@ bool FTensor::map(offset_t offset) {
 
   page_id_t page_id = offset / page_size_;
   if (mapping_.find(page_id) != mapping_.end()) {
-    LOGGER(ERROR, "Page %ld is already mapped.", page_id);
-    return false;
+    return true;
   }
 
   auto vaddr = reinterpret_cast<generic_ptr_t>(
@@ -122,8 +121,7 @@ bool FTensor::unmap(offset_t offset) {
 
   page_id_t page_id = offset / page_size_;
   if (mapping_.find(page_id) == mapping_.end()) {
-    LOGGER(ERROR, "Page %ld is not mapped.", page_id);
-    return false;
+    return true;
   }
 
   auto vaddr = reinterpret_cast<generic_ptr_t>(


RixinLiu:
The PR says "real VMM map/unmap failures still propagate through the existing CHECK_GPU paths." That's true, but CHECK_GPU and the duplicate-map check protect against different things.

CHECK_GPU catches CUDA VMM API errors — driver or hardware failures. The duplicate-map check catches logical errors in page allocation — e.g., two callers mapping the same virtual slot, which means the upper-layer bookkeeping has a bug. CUDA will happily remap a virtual address to new physical memory without returning an error, so CHECK_GPU will never fire in that case.

Also, I looked for scenarios where duplicate map/unmap can legitimately happen in the current code:

- IPC broadcast replay: _broadcast_map_to_kv_tensors does asyncio.gather once, raises on failure, and has no retry logic. Messages are not replayed.
- Preallocation vs on-demand race: alloc_page fast path takes from reserved_page_list_ (already mapped) and returns without calling map_pages. The slow path takes from free_page_list_ (unmapped). The two paths operate on different pages — they don't overlap.

So under normal operation, a duplicate map means something is wrong upstream. Making it silently succeed removes the only signal that would surface that bug.

Feel free to ping me if you really find a situation that can cause this problem :)

shipiyouniao:
Thanks for checking the concrete paths.

This change came from an internal DFX/code-review pass as a preventive finding. We noticed that duplicate map/unmap requests were indistinguishable from VMM failures at the caller boundary and treated retries or repeated coordination messages as potentially benign. It was identified through code inspection, not from a test failure or a production reproduction.

After rechecking the current broadcast and preallocation paths, we do not have a legitimate scenario that produces duplicate map/unmap operations. Our internal review also agrees that silently accepting them would hide an upstream bookkeeping error, so this PR should be closed.

The partial-batch and cross-rank rollback problem tracked by #412/#418 is separate. That transaction path should distinguish pre-existing mappings from mappings created by the current operation without changing the global `FTensor::map()` / `unmap()` contract.

Thanks for the detailed analysis.

Issue #375, opened 2026-07-01
Discussion: stable extension points for production kvcached integrations


shipiyouniao:
Hi, we noticed that the README highlights Red Hat's Sardeenz as a production-oriented integration built on top of kvcached. That direction is very interesting to us.

We are evaluating a similar but broader production integration layer around kvcached for SGLang/vLLM serving stacks. The goal is not to replace kvcached's core memory manager, but to make it easier to operate kvcached in a production platform: deployment, lifecycle management, admission policy, observability, failure diagnosis, and coordination across multiple runtime instances.

One concern is that a production integration can easily end up depending on private monkey patches or internal module layouts. That becomes fragile across kvcached, SGLang, and vLLM upgrades.

Would the project be open to defining a few stable extension points for external integration layers?

Some areas we are interested in:

- Runtime reservation reporting
  - e.g. runtime-owned pools that kvcached should account for but not manage directly.

- Allocator / pool capability discovery
  - e.g. whether a backend supports KV pooling, Mamba/linear state pooling, activation/workspace pooling, or only accounting.

- Non-KV memory pooling direction
  - e.g. whether kvcached intends to stay focused on KV cache only, or whether pooling/accounting for other runtime memory such as activation buffers, temporary workspace, or model-specific state would be in scope in the future.

- Lifecycle hooks
  - e.g. clean startup/shutdown, worker registration, IPC readiness, failure propagation, and safe degradation.

- Metrics / observability hooks
  - e.g. exporting pool usage, reservation breakdown, reclaim stats, per-runtime status, and error state.

- Version-compatible patch boundaries
  - e.g. what should live in kvcached upstream vs. what should be left to integration projects.

The goal is to keep production integrations aligned with kvcached instead of carrying long-lived downstream patches.

Would maintainers prefer these to be proposed as small PRs case by case, or is there interest in a more explicit integration API for production platform integrations?

jiarong0907:
@shipiyouniao Thanks for the detailed writeup! 

Yes, we are open to defining stable extension points. Just to clarify, the stable extension points here are used to support the orchestration layer, right? We are also headache about the upgrades of upstream vLLM and SGLang. Since it is hard to merge the interfaces with upstream engines, we are thinking about having an automated way of integrating kvcached with these engine upgrades and keep a kvcached-vLLM and kvcached-SGLang repos in our org. This is part is still under developmenet.

[...]

PR #385, opened 2026-07-09
Add read-only observability snapshots

shipiyouniao:
## Summary

Adds a read-only observability surface for kvcached runtime integrations and KV cache pools.

kvcached owns the allocator measurements and byte accounting. External monitoring systems can periodically collect and forward structured snapshots without depending on private allocator or engine-patch details.

## What is included

- `kvcached.observability` dataclasses for runtime state, KV pool state, and capability discovery
- block and byte measurements for available, allocated, reserved, virtual, and mapped KV memory
- a thread-safe weak pool registry that does not extend `KVCacheManager` lifetime
- `KVCacheManager.observability_snapshot()` and `observability_snapshot_dict()`
- vLLM and SGLang runtime snapshot helpers
- automatic registration of vLLM and SGLang KV pools, with cleanup on shutdown
- integration helpers that return snapshots for every live registered pool
- unit tests with no CUDA/GPU dependency

## Design boundaries

- Read-only: no allocation, admission, GC, or scheduling policy is added.
- Pull-based: kvcached measures current state on demand; collectors decide polling and transport.
- Engine-neutral schema: consumers do not need to inspect SGLang or vLLM patch internals.
- Related to #375, but this PR only provides the observability/accounting surface.

## Validation

Local validation:

```text
pytest tests/test_observability.py -q
# 6 passed

ruff, codespell, isort, clang-format, PyMarkdown, actionlint,
and trailing-whitespace pre-commit hooks
# all passed

mypy 1.11.1, Python targets 3.9 / 3.10 / 3.11 / 3.12 / 3.13
# 24 kvcached files + 15 test files passed for every target
```

GitHub Actions is green for pre-commit and all five MyPy jobs.

GPU validation used the SGLang v0.5.13 image on one explicitly isolated RTX 4090. A real CUDA VMM-backed MHA pool was created, sampled, allocated by 32 blocks, sampled again, freed, and sampled a third time:

```text
visible GPUs:             1 x NVIDIA GeForce RTX 4090
allocated blocks:        1 -> 33 -> 1
allocated bytes:         512 -> 16896 -> 512
virtual total bytes:     33554432
mapped bytes:            8388608
identity:                integration=sglang, pool_name=mha, group_id=7
```

### DIFF
--- a/kvcached/integration/sglang/interfaces.py
+++ b/kvcached/integration/sglang/interfaces.py
@@ -7,6 +7,15 @@ from typing import Any, Dict, List, Optional, Tuple, Union
 import torch
 
 from kvcached.kv_cache_manager import KVCacheManager
+from kvcached.observability import (
+    build_runtime_snapshot,
+    get_registered_kv_cache_pool_snapshot_dicts,
+    get_registered_kv_cache_pool_snapshots,
+)
+from kvcached.pool_registry import (
+    clear_registered_kv_cache_pools,
+    register_kv_cache_pool,
+)
 from kvcached.tp_ipc_util import start_worker_listener_thread
 from kvcached.utils import CONTIGUOUS_LAYOUT, PAGE_SIZE, get_kvcached_logger, normalize_gpu_device
 from kvcached.vmm_ops import (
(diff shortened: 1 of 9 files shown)


jiarong0907:
@shipiyouniao If the changes look good to you, we can merge this PR.

shipiyouniao:
Thanks, the latest changes look good to me. I added regression coverage for the corrected per-layer semantics, negative block gauges, the synchronized registry path, and reserved-block accounting; the targeted tests and all current CI checks pass. From my side, #385 is ready to merge.

One follow-up is #436. Since #385 introduces the snapshot reader that can run concurrently with the allocator, and downstream monitoring/exporters may poll these fields continuously, the unlocked C++ counter reads should be fixed soon. I suggest landing #436 before #410/#421 and before enabling production polling. Clamping `available_blocks` prevents one invalid gauge value, but it does not remove the C++ data race or protect the raw `free_pages` / `inuse_pages` fields.

jiarong0907:
> Thanks, the latest changes look good to me. I added regression coverage for the corrected per-layer semantics, negative block gauges, the synchronized registry path, and reserved-block accounting; the targeted tests and all current CI checks pass. From my side, #385 is ready to merge.
> 
> One follow-up is #436. Since #385 introduces the snapshot reader that can run concurrently with the allocator, and downstream monitoring/exporters may poll these fields continuously, the unlocked C++ counter reads should be fixed soon. I suggest landing #436 before #410/#421 and before enabling production polling. Clamping `available_blocks` prevents one invalid gauge value, but it does not remove the C++ data race or protect the raw `free_pages` / `inuse_pages` fields.

Sounds good to me!

One thing i want to understand a bit more is that if we use lock to remove the data race, would that introduce too much overhead for kvcached because of the data polling thread could hold the lock and the kvcached process have to wait for the lock.

shipiyouniao:
Good question. I ran a focused benchmark:

| Test | Relaxed atomic | Mutex |
| --- | ---: | ---: |
| Single-thread getter | 1.64 ns/op | 19.68 ns/op |
| Hot loop, no polling | 34.63 Mops/s | 21.85 Mops/s |
| Hot loop, 1-second polling | 34.54 Mops/s | 21.77 Mops/s |

At a normal 1-second polling interval, the polling thread itself introduced no measurable additional contention. The main overhead comes from acquiring the mutex on every foreground getter call.

More importantly, adding `lock_` directly to the existing getters is unsafe: `resize()` already holds `lock_` and calls `get_num_inuse_pages()`, which causes a self-deadlock with a naively locked getter.

I therefore think #436 should use relaxed atomic counters for the hot getters, plus a single locked snapshot method if monitoring requires a coherent multi-field view. Thanks for raising this concern.

PR #390, opened 2026-07-17
fix(vllm): evict prefix-cache blocks by page to release memory (#359)

### LINKED ISSUE #359: Idle instance pins GBs beyond MAX_CACHED_TOKENS due to cached-block fragmentation across 2MB pages, starving co-located instances
With prefix caching enabled, an instance that has gone idle retains far more physical GPU memory than the KVCACHED_MAX_CACHED_TOKENS bound implies, measured ~8.9 GiB pinned vs. the nominal ~2.2 GiB (16000 tokens) on Qwen3-4B. Cached blocks are retained in place, scattered across the 2MB pages they were originally allocated from, and free_pages() can only unmap a page when it is completely empty. The pinned set is therefore bounded by pages touched, not tokens cached.

ryanManocha:
## Problem

With prefix caching enabled, an idle instance retains far more physical memory
than `KVCACHED_MAX_CACHED_TOKENS` implies — #359 measures ~8.9 GiB pinned
against a nominal ~2.2 GiB on Qwen3-4B.

The cap is counted in blocks, but memory is returned in pages: `free_pages()`
can only unmap a page once *every* block on it is free. `_evict_blocks_from_pool`
picks victims in pure LRU order, so the surviving blocks stay scattered across
the pages they were originally allocated from and each one pins a whole 2 MB
page. The pinned set ends up bounded by pages touched rather than blocks kept,
and a co-located instance is starved of memory the cap says was released.

## Root Cause

`ElasticBlockPool._evict_blocks_from_pool()` (`kvcached/integration/vllm/patches.py`)
evicts `len(_evictable_blocks) - max_cached_blocks` blocks oldest-first, with no
knowledge of which page each block sits on. Evicting N scattered blocks satisfies
the cap while freeing zero pages.

## What Changed

- **`kvcached/kv_cache_manager.py`** — new `get_page_occupancy(page_ids)`
  reporting how many blocks each page still holds, derived from
  `InternalPage.get_block_range()` so blocks straddling a page boundary are
  accounted for the same way the C++ allocator accounts for them.
- **`kvcached/integration/vllm/patches.py`** — `_evict_blocks_from_pool()` now
  orders victims by page: prefer pages the pool can empty outright, cheapest
  first, breaking ties toward the page whose most recently used block is oldest.
  Pages holding blocks from running requests can never be emptied by eviction
  and are skipped. Falls back to LRU order for any remaining budget, and when
  page information is unavailable.

The eviction *count* is unchanged — only the choice of victims. No additional
blocks are dropped, so prefix-cache hit rate is not traded for memory.

## Verification

Measured on an RTX PRO 4000 Blackwell (24 GB), 8 layers, 16 KiB blocks, 2 MB
pages (128 blocks/page). 4096 blocks cached, every stride-th touched so an
age-only policy spares it, then evicted down to 512. **Both columns evict the
same 3584 blocks:**

| stride | freed before | freed after |
|---|---|---|
| 1 | 0.84 GB | 0.88 GB |
| 2 | 0.75 GB | 0.88 GB |
| 4 | 0.50 GB | 0.88 GB |
| 8 | 0.03 GB | 0.88 GB |

Age-only eviction collapses as the survivors scatter: at stride 8 it evicts 3584
blocks and frees essentially nothing. Page-aware selection holds flat.

`benchmarks/bench_frag/` contains both harnesses: `bench_frag.py` measures the
allocator's pinned-to-held ratio directly (reproducing #359's premise at
1.0x/2.0x/4.0x as blocks scatter), and `bench_evict.py` produces the table above.

## Tests

Added `tests/test_page_aware_eviction.py` (mocks torch and the C extension, no
GPU needed, following `tests/test_prefix_cache.py`):

- `test_evicting_a_pages_worth_frees_a_page` — **fails on main**, passes here
- `test_skips_pages_pinned_by_active_blocks`
- `test_evicts_requested_count`
- `test_falls_back_to_lru_without_page_allocator`

All 28 existing `tests/test_prefix_cache.py` tests still pass (32 total).

## Scope

Numbers above come from driving `ElasticBlockPool` directly rather than a live
vLLM server, so the issue's ~8.9 GiB → ~2.2 GiB figure on Qwen3-4B is not
independently confirmed here. They were collected on a PyTorch build that
predates this GPU's `sm_120` architecture (warns at startup); memory-mapping
behaviour should be unaffected.

## Issue

Closes #359

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -175,6 +175,7 @@ benchmarks/bench_vmm/*.bin
 benchmarks/bench_latency_benefit/results/
 benchmarks/bench_latency_benefit/plot*
 benchmarks/gsm8k/*.jsonl
+benchmarks/bench_idle_footprint/*.jsonl
 benchmarks/gsm8k/*.txt
 
 # tools
--- a/benchmarks/bench_frag/README.md
+++ b/benchmarks/bench_frag/README.md
@@ -1,0 +1,37 @@
+# bench_frag
+
+Measures how much physical memory stays pinned when retained KV blocks are
+scattered across pages ([#359](https://github.com/ovg-project/kvcached/issues/359)).
+
+kvcached returns memory to the GPU one page at a time, and only once every block
+on that page is free. A cache bound expressed in tokens therefore says nothing
+about how much memory an idle instance actually holds: 1024 retained blocks
+packed together occupy 8 pages, while the same 1024 blocks spread one per page
+occupy 128. The `waste` column is that ratio.
+
+## Run
+
+```bash
+python bench_frag.py
+```
+
+Defaults to 16 layers and 16384 blocks of 16 KiB, i.e. 128 blocks per 2 MB page.
+Each row frees all but `KEEP` blocks at the given stride and reports the memory
+still mapped afterwards.
+
+| column | meaning |
+|---|---|
+| `stride` | spacing between retained blocks; 1 is packed, 16 is one per page |
+| `kept` | blocks retained (constant across rows) |
+| `held GB` | memory the retained blocks themselves need |
+| `pinned GB` | memory still mapped, from `get_mapped_memory_size()` |
+| `waste` | `pinned / held` — 1.0x is ideal, higher means fragmentation |
+
+`stride=1` is the floor: it is what a page-aware policy converges toward. Rows
+above it show what scattering costs.
+
+## Scope
+
+This drives `KVCacheManager` directly, so it measures the allocator's behaviour
+rather than any engine's eviction policy. It reproduces the conditions #359
+describes; it does not by itself exercise `ElasticBlockPool`'s block selection.
--- a/benchmarks/bench_frag/README.md
+++ b/benchmarks/bench_frag/README.md
@@ -35,3 +35,27 @@ above it show what scattering costs.
 This drives `KVCacheManager` directly, so it measures the allocator's behaviour
 rather than any engine's eviction policy. It reproduces the conditions #359
 describes; it does not by itself exercise `ElasticBlockPool`'s block selection.
+
+# bench_evict
+
+Measures how much memory prefix-cache eviction actually returns. Caches 4096
+blocks, touches every stride-th block so an age-only policy spares it, then
+evicts down to 512 and reports the memory released.
+
+```bash
+python bench_evict.py
+```
+
+Measured on an RTX PRO 4000 Blackwell (24GB), 8 layers, 16 KiB blocks, 2 MB
+pages (128 blocks per page). Both columns evict the same 3584 blocks:
+
+| stride | freed before (LRU) | freed after (page-aware) |
+|---|---|---|
+| 1 | 0.84 GB | 0.88 GB |
+| 2 | 0.75 GB | 0.88 GB |
+| 4 | 0.50 GB | 0.88 GB |
+| 8 | 0.03 GB | 0.88 GB |
+
+Age-only eviction degrades as the retained blocks scatter: at stride 8 it evicts
+3584 blocks and frees almost nothing, because each surviving block pins a page.
+Page-aware selection holds flat, evicting the same count.
--- a/benchmarks/bench_frag/bench_evict.py
+++ b/benchmarks/bench_frag/bench_evict.py
@@ -2,14 +2,19 @@
 # SPDX-License-Identifier: Apache-2.0
 """Measure memory released by prefix-cache eviction (issue #359).
 
-Fills ElasticBlockPool's prefix cache, releases the blocks so they become
-evictable, then evicts down to a cap and reports how much physical memory came
-back. Pages only unmap once every block on them is free, so an eviction policy
-that ignores pages can hit the cap while releasing nothing.
+Pages unmap only once every block on them is free, so an eviction policy that
+picks victims by age alone can hit its block cap while releasing no memory: the
+survivors stay scattered and each one pins a whole page.
 
-Run on each branch and compare the "freed" column:
+This models that. Cache a large run of blocks, then touch every stride-th block
+so it looks recently used. Evicting down to the cap must then choose between the
+cold blocks (scattered around the hot ones) and the pages they sit on. The
+"freed" column is how much physical memory the eviction actually returned.
+
+Run on each branch and compare:
     python bench_evict.py
 """
+import time
 import types
 
 import torch
@@ -27,10 +32,8 @@ DTYPE = torch.float16
 DEVICE = f"cuda:{TP_RANK}"
 KV_SHAPE = (2, NUM_BLOCKS, BLOCK_SIZE, 8, 64)
 
-# Cache this many blocks, then evict down to KEEP. Mirrors an instance going
-# idle with MAX_CACHED_TOKENS worth of prefix cache left over.
-CACHED = 4096
-KEEP = 512
+CACHED = 4096  # blocks cached before eviction
+KEEP = 512  # cap to evict down to; must be well below CACHED
 
 
 class _Block:
@@ -53,7 +56,6 @@ def build_pool():
                   async_sched=False)
     alloc_kv_cache(kvcache_shape=KV_SHAPE, block_size=BLOCK_SIZE, dtype=DTYPE,
                    device=DEVICE, num_layers=NUM_LAYERS)
-    import time
     t0 = time.time()
     while not kv_tensors_created():
         if time.time() - t0 > 10.0:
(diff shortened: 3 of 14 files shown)


shipiyouniao:
I think the page-aware policy is being applied in one path where it changes prefix-cache semantics without providing a physical-memory benefit.

`_evict_blocks_from_pool()` is called both when enforcing `max_cached_blocks` and from `get_new_blocks()` when the current request is short of logical blocks. Page-aware eviction is useful for the former, but in the latter case evicting any cached block makes that logical block immediately reusable; the page does not need to be fully unmapped.

I reproduced this with this PR's `ElasticBlockPool` path: I filled blocks 1 through 8 in LRU order (1 oldest, 8 newest), left no free logical blocks, and called `get_new_blocks(1)`. The page-aware path evicted and reused block 8 while block 1 remained cached, because block 8 was the only block on a page that could be emptied. This shows that equal eviction count does not imply equal cache behavior: victim identity changes and a newer prefix can be discarded before the LRU victim.

Could we keep pure LRU for allocation-shortage eviction and use page-aware selection only for cache-cap trimming or explicit physical-memory release, or pass an explicit strategy flag? This would also keep the full evictable-block scan and page sort out of the normal allocation-pressure path. A regression test where the oldest block resides on a fuller page and a newer block resides on a nearly empty page would capture the distinction.

ryanManocha:
@shipiyouniao Good catch, you're right that the allocation-shortage path gets no physical-memory benefit from page-aware selection, and reordering victims there only trades a newer prefix for an older one. I fixed this by making the strategy explicit per caller.

`_evict_blocks_from_pool(num_to_evict, page_aware=True)` now takes a strategy flag:

* **Allocation shortage (`get_new_blocks`)** calls it with `page_aware=False`, so eviction stays pure LRU since the freed logical block is reused immediately and no page is unmapped. This also keeps the evictable-block scan and page sort out of the allocation path, as you pointed out.
* **Cache-cap trimming (`free_blocks`)** continues using `page_aware=True`, since that's the only path where page-aware selection can actually release physical memory.

The eviction count is unchanged in both cases.

I also added a regression test (`test_lru_eviction_ignores_page_layout`) with the exact scenario you described: the oldest block is on a full page while the newest block is alone on a nearly empty page. It verifies that the allocation path evicts the oldest block. The test fails with `page_aware=True` and passes with `page_aware=False`, so it guards this distinction going forward.

Thanks for the detailed repro!

RixinLiu:
I reproduced #359 and measured this PR against it: idle footprint drops from 10.3 GB to 5.2 GB on a real vLLM server. It works.

## Experiment design

- **Metric**: kvcached's own `used_size` (MemInfo shm), not `nvidia-smi`, which counts weights and activations too. One page id is mapped in every layer and both KV buffers, so on Qwen3-4B a 2 MiB page id costs **144 MiB**.
- **Lower bound**: the *packed floor* — what the cap costs with zero holes. 16000 tokens / block_size 16 = 1000 blocks, 64 per page → **16 pages = 2.25 GB**. Nothing beats that without caching less than the cap allows. (Matches the issue's "nominal ~2.2 GiB".)
- **Confidence**: two runs per config, and idle memory is stable across runs.
- **Workload**: 2500 requests, concurrency 96, varied prompt lengths, and decode lengths sampled from [16, 256]. The decode distribution dominates:

| workload | idle pages |
|---|---|
| 4000 req, concurrency 32, `max_tokens=4` | 27 |
| 5000 req, concurrency 96, varied prompts, `max_tokens=4` | 27 |
| **2500 req, concurrency 96, varied prompts, `max_tokens` in [16,256]** | **73** |

More requests and more concurrency changed nothing; **staggered decode lengths** did. Prefill grabs many blocks at once and `_alloc` drains one page before moving on, so prefill-shaped traffic packs itself. Decode grabs **one block at a time**, and many requests decoding at different rates drop those single blocks into holes across many pages. A prefill-only benchmark will report kvcached as near-optimal and miss this entirely.

## Results

L40, Qwen3-4B, vLLM 0.22.1, two runs each:

| | idle memory | pages | vs floor |
|---|---|---|---|
| `main` | 10.27 / 10.55 GB | 73 / 75 | 4.6x |
| **this PR** | **5.20 / 5.34 GB** | **37 / 38** | **2.3x** |
| packed floor | 2.25 GB | 16 | 1.0x |

Throughput identical (11.7–11.9 req/s), hits bit-identical (93,952).

**The gain lands entirely when traffic stops.** Sampling once a second, `main` and this PR overlap for the whole run (~10 GB) and separate only as the last requests drain. That is the policy's definition, not a flaw: while requests are in flight they hold blocks all over the pool, so hardly any page is fully evictable and there is nothing to choose from. A CPU-side ablation makes it exact — with 1000 blocks held by running requests, **0 of 89** pages are eligible at any eviction budget. Worth stating in the PR description so nobody expects steady-state gains.

**Hit-rate cost — a toy measurement, read it as an existence proof.** It follows from the above that the two policies make *identical* choices while traffic is running, and only diverge once the pool goes quiet (measured: hit rate during the first round of traffic is 0.4435 for `main` vs 0.4420 for this PR). So the measurement has to be: send traffic, let it go idle for 45 s so the trim runs, then send more — and have that second round ask for the **cold** prompts the first round cached, since hot prefixes survive under any policy and fresh prompts were never cached at all. Both rounds therefore draw from one fixed pool of 200 prompts, and the hit rate afterwards reads out what the trim kept:

| | memory while idle | hit rate after idle |
|---|---|---|
| `main` | 9.00 / 8.86 GB | 0.4200 / 0.4200 |
| this PR | 4.92 / 4.78 GB | 0.4138 / 0.4121 |

I would not carry the ~0.70 pp anywhere. The prompts are random token ids, the reuse pattern is uniform sampling from a 200-prompt pool rather than anything a real deployment produces, and the number moves with knobs I picked (pool size, gap length, hot/cold ratio) — one model, one GPU, one cap, two runs. All it establishes is that the cost is non-zero and small *here*.

Structurally there is a reason not to expect it to be large: this PR changes *which* blocks are evicted, not *how many*, so both sides retain exactly `max_cached_blocks` = 1000 blocks — 2.25 GB of actual KV either way. The 4+ GB memory difference is empty space inside pages, not cache. What is left to cost anything is the change in *identity* of the retained set: **LRU keeps the newest 1000, this PR keeps a set skewed toward blocks whose pages could not be emptied.** Someone should measure this properly on a real trace before anyone quotes a number.

## Follow-up: the allocation side

Eviction only cleans up after the fact; which page a block came from is decided at allocation, and `_alloc` has no notion of it — `avail_pages.popitem()` takes the most recently touched page, and a partly consumed page goes back at the tail. A long prefill run therefore walks the small holes at the tail (2 blocks here, 3 there) and smears over a dozen pages, each pinned as long as any of those blocks stays cached.

Taking the smallest page that fits the whole remaining run instead (and the emptiest when none does) is one line in `_alloc` plus a ~20-line helper:

| | idle memory | pages | vs floor |
|---|---|---|---|
| `main` | 10.27 / 10.55 GB | 73 / 75 | 4.6x |
| this PR | 5.20 / 5.34 GB | 37 / 38 | 2.3x |
| **this PR + page-aware allocation** | **3.09 / 3.09 GB** | **22 / 22** | **1.4x** |

Hits and throughput unchanged. The two halves are complementary — allocation alone, with LRU eviction, only reaches 67 pages, since the survivors are still just "the newest 1000 blocks" spread everywhere.

## What the commit contains

| file | | what it does |
|---|---|---|
| `kvcached/kv_cache_manager.py` | +38 −1 | the change: `_alloc` picks the smallest page that fits the whole remaining run instead of `avail_pages.popitem()` |
| `tests/test_bestfit_page_selection.py` | 186 | 7 tests for the selection rule and for a run landing on one page; five fail without the change |
| `benchmarks/bench_idle_footprint/run_idle_footprint.py` | 221 | main driver — serve, load, idle, report memory and a per-second timeline |
| `benchmarks/bench_idle_footprint/run_reuse_after_idle.py` | 135 | traffic → idle gap → traffic, for the hit-rate cost |
| `benchmarks/bench_idle_footprint/sim_eviction.py` | 201 | CPU-only: how many pages are eligible for a given number of live blocks |
| `benchmarks/bench_idle_footprint/workload.py` | 159 | load generator — hot prefixes plus a cold tail, prompt and decode lengths sampled |
| `benchmarks/bench_idle_footprint/probe_mem.py` | 98 | reads kvcached's MemInfo segments from `/dev/shm` |
| `benchmarks/bench_idle_footprint/README.md` | 131 | method, geometry, and the reference numbers |
| `.gitignore` | +1 | ignore the timeline files the runner writes |

RixinLiu:
@ryanManocha Thanks for the contribution! I reproduced your results, and show the benefit, and i also enhanced it by improving the allocation strategy. You may take a look, i will merge it after your review

ryanManocha:
@RixinLiu The allocation changes look good, and thanks for validating them on a real server. I went through the code and tests, ran everything locally, and confirmed the issue is fixed. Looks good to me and I think it's ready to merge!

PR #397, opened 2026-07-20
feat(vllm): support contiguous KV layout for hybrid linear-attention (Qwen3-Next/GDN)

RixinLiu:
## What this PR does

Enable the contiguous KV layout for hybrid linear-attention models (Qwen3-Next / Qwen3.5 GDN, Mamba2), including the `ratio > 1` configurations that are the default on vLLM 0.22+. The changes form one logical chain, laid out below.

## Background: why hybrid pools are unified

In a hybrid model, attention and mamba layers draw block ids from one shared BlockPool, so every block id must map to a byte span of the same size. The mamba recurrent state is one indivisible blob per block (~2 MB on Qwen3.5-35B, dominated by the fp32 SSM state), so it sets a rigid floor; vLLM stretches the flexible side to match by raising the attention `block_size` until the attention page covers the mamba page. The resulting pool is a single flat buffer per slot: K and V interleaved inside each block's cell, mamba state reinterpreting the same address space — there are no separate K and V halves.

## Step 1: `num_kv_buffers` reconciliation (contiguous hybrid works at all)

kvcached's allocator describes a pool as `num_kv_buffers` mirrored halves (2 for MHA/GQA's split K/V, 1 for MLA). The hybrid unified pool was passing 2 out of habit, but as described above it is structurally ONE buffer. The C++ contiguous allocator sizes its compound page as `kPageSize * num_layers * num_kv_buffers`, so the stale 2 made the compound page twice too large and tripped `FTensor::map`'s offset-alignment assert on every odd page id (the non-contiguous layout ignores this factor, which is why the bug never surfaced there). Passing 1, dropping the blanket "hybrid must be non-contiguous" guard, and restoring `_reshape_mamba_contiguous` makes contiguous hybrid work — validated token-for-token on both linear-attention kernel families (Mamba2 via Zamba2-1.2B, GDN via Qwen3-Next) on vLLM 0.13, where `block_size == kernel_block_size` (ratio = 1).

## Step 2: ratio > 1 (contiguous hybrid works on modern vLLM)

vLLM 0.22+ auto-aligns the hybrid `block_size` to the mamba page (e.g. 1056 tokens on Qwen3.5-35B) while attention kernels still index 16-token kernel blocks, so `ratio = block_size / kernel_block_size > 1` is now the default. Under the historical per-block layout (each slot's kernel blocks packed contiguously inside its segment), a slot's kernel-block id then advances with two different step sizes — a small in-block step and a large cross-block jump — which a single-stride `as_strided` view cannot express; this combination used to fail loud, leaving contiguous hybrid unusable on modern vLLM. The unlock is a verified structural fact: a virtual block is owned by exactly one KV-cache group at a time (groups draw disjoint ids from the shared BlockPool), so a block's whole span is single-typed and the same bytes can carry a per-owner arrangement. Attention-owned blocks are now laid out as `ratio` rows of `[slot0: K|V][slot1: K|V]...`, giving slot i's kernel block kb the element address `kb * (num_layers * 2h) + i * 2h` — globally linear, because the span size `T = ratio * num_layers * 2h` makes the cross-block step equal the in-block step. Mamba-owned blocks keep the slot-sequential layout consumed by `_reshape_mamba_contiguous`, unchanged. At ratio == 1 the new formula degenerates byte-for-byte into the historical layout, and mapping is untouched: pages are chosen purely from the virtual block index and always cover the whole span, so the C++ allocator needs no changes.

## Validation

CPU: existing ratio==1 byte-offset tests pass unchanged (backward compat); new tests pin the kernel-granular strides (ratio 1/2/4, FlashAttn and FlashInfer shapes, inner strides included), prove the attention and mamba interpretations of a to its span (no gaps, no overlap, no cross-block leakage), and keep the`raw_info` contracts. GPU (vLLM 0.22.1): tiny GDN hybrids at ratio=5 with 1-slot and 2-slot pools, token-for-token identical to the no-kvcached baseline for BOTH contiguous (first time this combination runs on 0.22+) and non-contiguous.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ GPT-OSS support in SGLang updated to **v0.5.9**.
 | Engine | Versions | Attention types | Example models |
 |--------|----------|-----------------|----------------|
 | SGLang | ≥ v0.4.9 (tested up to v0.5.10) | MHA / GQA / MLA | Llama 3.1/3.3, Qwen 2.5, DeepSeek-V3, openai/gpt-oss-20b, etc. |
-| vLLM | ≥ v0.8.4 (tested up to v0.25.1) | MHA / GQA / MLA | Llama 3.1/3.3, Qwen 2.5, DeepSeek-V3, openai/gpt-oss-20b |
+| vLLM | ≥ v0.8.4 (tested up to v0.24.0) | MHA / GQA / MLA | Llama 3.1/3.3, Qwen 2.5, DeepSeek-V3, openai/gpt-oss-20b |
 
 ## Example use cases
 
--- a/benchmarks/bench_gemma_hetero/README.md
+++ b/benchmarks/bench_gemma_hetero/README.md
@@ -1,0 +1,44 @@
+# bench_gemma_hetero — kvcached correctness on heterogeneous-KV-group models (Gemma 3/4)
+
+Byte-exact correctness of kvcached on a **heterogeneous-KV-group** hybrid model,
+where sliding-window and full-attention layers have different block geometry
+(`block_size`/`cell_size`) but the same `block_mem_size`. `google/gemma-4-12B-it`
+is the reference: sliding-window layers `block_size=16/cell=4096`, full-attention
+layers `block_size=64/cell=1024` (full layers use 1 shared KV head x head_dim 512). Support for this landed on branch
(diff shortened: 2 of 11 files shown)


RixinLiu:
@copilot resolve the merge conflicts in this pull request

PR #402, opened 2026-07-22
fix(vllm): preserve duplicate prefix cache blocks

### LINKED ISSUE #401: vLLM prefix cache loses block metadata for duplicate in-flight prefixes
## Problem

`ElasticBlockPool._cached_blocks` maps `(block_hash, group_id)` to a single physical block. In `cache_full_blocks()`, a second block with the same key is treated as an idempotent cache call and skipped.

That is not always idempotent. Concurrent requests can materialize the same full prefix before either physical block becomes reusable. vLLM keeps both block identities, but kvcached currently records only the first one. The skipped block receives neither `block_hash` metadata nor a reverse-index entry.

Under a high-concurrency hybrid-model workload this eventually reaches vLLM's cache accounting with an unset hash:

```text
single_type_kv_cache_manager.py
assert block.block_hash is not None
```

The EngineCore then exits and an otherwise valid request fails.

## Relation to #363

#363 correctly restored block-hash set/reset lifecycle, but duplicate physical blocks sharing one hash are still collapsed by the cache index. This is a separate case on top of that fix.

## Expected behavior

- Preserve every physical block that materializes a cached prefix, even when another block has the same `(block_hash, group_id)`.
- Keep hash lookup behavior compatible by returning one matching block.
- Evict and reset metadata by exact `block_id`, without deleting sibling blocks with the same hash.

## Reproduction

A unit reproduction is:

1. Allocate two distinct blocks.
2. Cache both with the same request prefix hash.
3. Verify both blocks retain `block_hash` metadata.
4. Evict one block and verify the sibling remains cached and keeps its metadata.

shipiyouniao:
## Summary

Preserve every physical vLLM prefix-cache block when concurrent requests materialize the same `(block_hash, group_id)`.

Closes #401.

## Root cause

`ElasticBlockPool._cached_blocks` currently maps each cache key to one `KVCacheBlock`. `cache_full_blocks()` treats an existing key as an idempotent call and skips the new block.

Concurrent requests can produce the same full prefix before either physical block becomes reusable. In that case the second block is distinct, not an idempotent repeat. Skipping it leaves `KVCacheBlock.block_hash` unset and omits its reverse-index entry. vLLM's later cache accounting can then fail at `assert block.block_hash is not None`, terminating the EngineCore.

This complements #363: that PR fixed hash set/reset lifecycle, while this PR preserves multiple physical blocks that legitimately share one hash.

## Changes

- change the prefix index from `cache key -> block` to `cache key -> {block_id: block}`
- keep lookup compatibility by returning one matching block
- remove and reset cached entries by exact `block_id`, so evicting one duplicate does not delete its siblings
- handle a reused block moving from an old cache key to a new one
- add a regression test for two blocks with the same prefix, followed by separate release and eviction

## Validation

- `python -m pytest tests/test_prefix_cache.py -q` -> `30 passed`
- complete `pre-commit run --all-files` -> all hooks passed
- manual mypy matrix for Python 3.9, 3.10, 3.11, 3.12, and 3.13 -> all passed
- high-concurrency two-instance vLLM integration replay with prefix caching -> `1440/1440` requests completed, with no block-hash assertion or EngineCore failure

The branch is based on current upstream `main` (`927ef66c`) and contains one focused commit.

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -441,16 +441,36 @@ class ElasticBlockPoolPatch(VersionAwarePatch, BasePatch):
                 self.null_block = self.kv_block_pool[0]
                 self.null_block.is_null = True
 
-                # Prefix cache: (block_hash, group_id) -> KVCacheBlock
+                # Prefix cache: (block_hash, group_id) -> {block_id: block}.
+                # Multiple in-flight requests can materialize the same prefix
+                # before either block becomes reusable. vLLM preserves every
+                # such block because request block tables are append-only.
                 # The key embeds group_id to support hybrid attention
                 # (multiple KV cache groups with different attention types).
-                self._cached_blocks: dict[Any, KVCacheBlock] = {}
+                self._cached_blocks: dict[Any, dict[int, KVCacheBlock]] = {}
                 # Reverse index: block_id -> cache key for O(1) eviction.
                 # Each block_id belongs to exactly one group.
                 self._block_id_to_key: dict[int, Any] = {}
                 # LRU evictable pool: blocks with ref_cnt==0 retained for
                 # cross-request prefix reuse. Insertion order = LRU order.
                 self._evictable_blocks: OrderedDict[int, KVCacheBlock] = OrderedDict()
+
+            def _get_one_cached_block(self, key: Any) -> Optional[KVCacheBlock]:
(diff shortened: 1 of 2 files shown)


RixinLiu:
The duplicate-hash handling mirrors vLLM's native BlockHashToBlockMap. LGTM.

PR #403, opened 2026-07-22
Add exhaustive CPU-only pytest CI

### LINKED ISSUE #395: CI runs no tests — add a GitHub Actions job for the CPU-only pytest suite
The only workflow in .github/workflows/ is pre-commit.yml (lint/format/mypy). None of the tests under tests/ run in CI, so every PR merges with zero automated test coverage and regressions are only caught when someone happens to run the right test by hand.

Lanoxia:
## Summary

Adds a hosted CPU-only pytest gate for #395. This revision is intentionally limited to the workflow, its runner, exhaustive test classification, and the minimal import-isolation fixes needed by the CPU suite.

## What changed

- Add a Python 3.9-3.13 CPU pytest workflow.
- Partition every top-level `tests/test_*.py` module into exactly one of `cpu`, `gpu`, or `integration`.
- Fail CI on unclassified, duplicate, unknown, unsafe, or unsorted manifest entries.
- Drive the CPU runner from `tests/manifests/cpu.txt` instead of a fixed list in the shell script.
- Add regression tests for the classification validator.
- Keep only the small torch/vLLM mocks required to collect dependency-isolated CPU tests.

## Scope

The NIXL P/D example, GPU-utilization integration change, and H20 validation report have been removed from this PR and will be handled separately.

## Validation

- `python tools/check_test_classification.py` -> `total=18, cpu=5, gpu=3, integration=10`
- `bash tools/run_cpu_tests.sh -q` -> `66 passed`
- `git diff --check`
- Ruff, isort, codespell, PyMarkdown, actionlint, and mypy hooks pass locally.

Closes #395

### DIFF
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -22,8 +22,17 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      # CPU-only PyTorch is what lets most of the suite run on a hosted
+      # runner: without it every module that reaches torch through an import
+      # chain has to be classified away from the cpu manifest.
+      - name: Install CPU-only PyTorch
+        run: python -m pip install torch --index-url https://download.pytorch.org/whl/cpu
+
       - name: Install test dependencies
-        run: python -m pip install "pytest>=8,<9" packaging
+        run: |
+          python -m pip install "pytest>=8,<9" packaging
+          python -m pip install -r requirements.txt
+          python -m pip install pyyaml numpy aiohttp
 
       - name: Validate test classification
         run: python tools/check_test_classification.py
--- a/.github/workflows/cpu-tests.yml
+++ b/.github/workflows/cpu-tests.yml
@@ -1,0 +1,32 @@
+name: CPU Tests
+
+on:
+  pull_request:
+  push:
+
+permissions:
+  contents: read
+
+jobs:
+  cpu-pytest:
+    name: CPU pytest (${{ matrix.python-version }})
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.9", "3.10", "3.11", "3.12", "3.13"]
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install test dependencies
+        run: python -m pip install "pytest>=8,<9" packaging
+
+      - name: Validate test classification
+        run: python tools/check_test_classification.py
+
+      - name: Run CPU-only pytest suite
+        run: bash tools/run_cpu_tests.sh
--- a/tests/manifests/README.md
+++ b/tests/manifests/README.md
@@ -1,0 +1,16 @@
+# Test execution boundaries
+
+Every top-level `tests/test_*.py` module must appear in exactly one manifest.
+`tools/check_test_classification.py` enforces complete, exclusive coverage and
+fails CI when a new test has not been classified.
+
+- `cpu.txt`: dependency-isolated tests that run on a hosted CPU runner without
+  PyTorch, a serving engine, the compiled VMM extension, or shared services.
+- `gpu.txt`: tests whose primary boundary is CUDA/HIP hardware and the compiled
+  extension.
+- `integration.txt`: tests that require a serving-engine or broader project
+  runtime, shared memory, multiprocessing, external services, or an end-to-end
+  environment. Some integration tests may also require a GPU.
+
+The CPU workflow executes only `cpu.txt`. GPU and integration runners may use
+their manifests as those CI environments are added.
--- a/tests/manifests/cpu.txt
+++ b/tests/manifests/cpu.txt
@@ -1,0 +1,5 @@
+tests/test_make_cache_key.py
+tests/test_page_aware_eviction.py
+tests/test_prefix_cache.py
+tests/test_test_classification.py
+tests/test_vllm_nixl_compat.py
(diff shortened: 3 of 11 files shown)


Lanoxia:
Ready for review.

I started with the CPU-only pytest coverage from #395, then added a small preflight mode for the NIXL P/D smoke harness so reviewers can validate the script/configuration path on a non-GPU machine before running the full CUDA/vLLM test:

`CHECK_ONLY=1 bash tools/run_vllm_nixl_pd_smoke.sh`

I kept the CI suite intentionally narrow to tests that are already mocked or pure-Python and do not require vLLM, SGLang, CUDA, or the C++ extension. Happy to split the NIXL P/D example/preflight work into a separate PR if that is easier to review.

shipiyouniao:
Thanks for taking on #395. I think this PR now needs to be split before further review.

The current branch combines several independently reviewable and independently risky changes:

- CPU-only pytest CI
- a NIXL P/D example and smoke harness
- GPU-utilization behavior changes in the serving integrations
- an H20 validation report

Those changes have different validation requirements and owners. Keeping them in one PR makes it difficult to determine what the CI change actually depends on and makes a future revert unnecessarily broad. Could you please keep this PR focused on the CPU test workflow, its runner, and only the minimal test-isolation fixes required by that workflow, and move the other changes into separate PRs? The existing pre-commit failure should also be resolved after the split.

There is also a coverage-maintenance concern with the current fixed three-file list in `tools/run_cpu_tests.sh`: newly added CPU-safe regression tests will remain silently excluded unless someone remembers to edit the script.

For test classification, I suggest using execution boundaries as the primary mechanism, for example:

- `tests/cpu/`: no physical accelerator, serving engine, or compiled VMM extension required
- `tests/gpu/`: requires CUDA/HIP and the compiled extension
- `tests/integration/`: requires vLLM/SGLang, multiprocess services, or an end-to-end environment

Then the hosted job can run only `pytest tests/cpu`. If moving files is too disruptive, explicit CPU/GPU/integration manifest files are also workable, provided CI validates that every `tests/test_*.py` module appears in exactly one manifest.

Pytest markers such as `cpu`, `gpu`, `integration`, and `slow` are useful secondary metadata and should use `--strict-markers`, but `pytest -m cpu` alone is not a complete isolation mechanism: pytest imports modules during collection before marker deselection, so GPU-only modules can still fail on a hosted CPU runner due to CUDA, vLLM, or extension imports.

Once the PR is focused and the classification rule prevents silent omissions, this will provide a useful merge gate for the regression tests being added across the project.

Lanoxia:
Thanks for the detailed guidance. I have rewritten the branch to keep this PR focused on the CPU test gate only.

The NIXL P/D example, GPU-utilization change, and H20 report are no longer part of this diff. The fixed three-file list has also been replaced with exhaustive `cpu` / `gpu` / `integration` manifests. A validator now fails on missing, duplicate, unknown, unsafe, or unsorted entries, so newly added `tests/test_*.py` modules cannot be silently omitted.

Current local validation:
- test classification: 18 modules exactly once (5 CPU, 3 GPU, 10 integration)
- hosted CPU suite: 66 passed
- Ruff, isort, codespell, PyMarkdown, actionlint, and mypy: passed

The PR now contains one focused commit on top of the latest main. I would appreciate another look when convenient.

RixinLiu:
## Summary

A hosted CPU-only pytest gate for #395: partition every `tests/test_*.py` into exactly one of `cpu`, `gpu`, `integration`, run the `cpu` set on GitHub-hosted runners across Python 3.9–3.13, and fail CI on any module that is not classified.

Closes #395.

## What the job does

1. install CPU-only PyTorch, `requirements.txt`, and the test-only deps
2. `tools/check_test_classification.py` — fails if any module is unclassified, classified twice, listed but absent, or the manifests are unsorted
3. `tools/run_cpu_tests.sh` — runs the `cpu` manifest

Runs on `ubuntu-latest`, matrix over Python 3.9, 3.10, 3.11, 3.12, 3.13.

## Classification

Each module was classified by what it actually does on a hosted runner, not by its name:

| category | count | criterion |
|---|---|---|
| `cpu` | 14 | collects and passes with CPU torch, no device, no engine |
| `gpu` | 5 | import fails on `kvcached.vmm_ops`/`sglang`, or the tests fail without a device |
| `integration` | 6 | needs a running engine, or is a `__main__` script that collects no pytest tests |

`gpu`: `test_ElasticMLAMemoryPool`, `test_kvcache_manager`, `test_paged_allocator_aliasing` fail at import; `test_alloc_kv_cache_alignment` (28 failed) and `test_hybrid_contiguous_layout` (15 failed) collect but need a device.

`integration`: `test_offline_serving` needs `vllm`, `test_sleep_manager` needs live endpoints, and `test_elastic_serving`, `test_prealloc_gil_deadlock`, `test_traffic_monitor`, `test_utils` contain no pytest tests at all.

## My commits on top

- `a737e06` — bind the `interfaces` stub to the parent package. `mock.patch()` resolves a dotted target with `getattr()`, which a hand-installed `sys.modules` entry never sets, so the suite failed on Python 3.9. Pre-existing on `main`; only visible once CI runs 3.9.
- `b8cc41d` — install CPU-only PyTorch, so torch-dependent modules can be classified `cpu` instead of away.
- `dcb197b` — merge `main`, to cover the seven test modules added while this PR was open.
- `d4451a8` — reclassify: `cpu` 5→14, `gpu` 3→5, `integration` 10→6.
- `ed8c944` — print the unclassified module and what to do instead of a traceback.

## Validation

| | Python 3.9 | 3.11 (torch 2.8) | 3.11 (torch 2.13) |
|---|---|---|---|
| `run_cpu_tests.sh` | 114 passed | 114 passed | 114 passed |
| before `a737e06` | 30 failed / 14 passed | 44 passed | 44 passed |

`check_test_classification.py` → `total=25, cpu=14, gpu=5, integration=6`. Adding an unclassified module exits 1 with the module named and the three manifests listed. `tests/test_test_classification.py` → 5 passed.

Issue #311, opened 2026-04-20
kvcached incompatible with PD disaggregation (NixlConnector)


AAbouzeid:
## Description

kvcached crashes during initialization when used with vLLM's PD (prefill-decode) disaggregation via NixlConnector. Two bugs were identified. A third concern (UCX + VMM memory compatibility for the actual transfer) appears to be addressed in recent UCX versions but remains untested.

PD disaggregation works correctly without kvcached using the same setup.

Related: #302

## Environment

| Component | Version / Detail |
|---|---|
| Machine | 2x A100 80GB SXM (NV12 NVLink) |
| CUDA | 12.8 |
| Python | 3.12.3 |
| vLLM | 0.19.0 |
| kvcached | 0.1.5 |
| KV connector | NixlConnector (UCX transport, `cuda_ipc,cuda_copy,tcp`) |
| Model | Qwen/Qwen2.5-1.5B-Instruct |

## Steps to reproduce

Start prefill and decode instances with kvcached + NixlConnector:

**Prefill (GPU 0):**
```bash
ENABLE_KVCACHED=true KVCACHED_AUTOPATCH=1 \
UCX_TLS=cuda_ipc,cuda_copy,tcp \
VLLM_NIXL_SIDE_CHANNEL_PORT=5600 \
CUDA_VISIBLE_DEVICES=0 vllm serve Qwen/Qwen2.5-1.5B-Instruct \
  --host 0.0.0.0 --port 8100 --max-model-len 1024 --gpu-memory-utilization 0.8 \
  --kv-transfer-config \
  '{"kv_connector":"NixlConnector","kv_role":"kv_producer","kv_buffer_device":"cuda"}'
```

**Decode (GPU 1):**
```bash
ENABLE_KVCACHED=true KVCACHED_AUTOPATCH=1 \
UCX_TLS=cuda_ipc,cuda_copy,tcp \
VLLM_NIXL_SIDE_CHANNEL_PORT=5601 \
CUDA_VISIBLE_DEVICES=1 vllm serve Qwen/Qwen2.5-1.5B-Instruct \
  --host 0.0.0.0 --port 8200 --max-model-len 1024 --gpu-memory-utilization 0.8 \
  --kv-transfer-config \
  '{"kv_connector":"NixlConnector","kv_role":"kv_consumer","kv_buffer_device":"cuda"}'
```

**Proxy (port 9100):**
```bash
python3 disagg_prefill_proxy_server.py \
[...]

Issue #302, opened 2026-04-14
Does KVCached support PD disaggregation?


wdx727:
Does KVCached support PD disaggregation, and what are the differences in usage?

jiarong0907:
Hi @wdx727, kvcached only hacks the part related to KV cache allocation. In my understanding, it should be naturally compitable with PD disaggregation, although we haven't test it yet. 

Feel free to test and report any issues if you have.

PR #351, opened 2026-06-01
fix: support HMA for vLLM OffloadingConnector

yu-zh47:
Fixes #267.

## Summary

This PR backports vLLM 0.21.0's HMA support for `OffloadingConnector` to older vLLM versions.

In vLLM 0.16.0 through 0.20.x, `OffloadingConnector` exists but does not inherit `SupportsHMA`. Since vLLM 0.16 enables the hybrid memory allocator path by default, `vllm serve ... --kv-offloading-size N` fails during connector creation with:

```text
ValueError: Connector OffloadingConnector does not support HMA but HMA is enabled.

Issue #93, opened 2025-08-31
[TODO] Support kvcached offloading to other storage like CPU memory


jiarong0907:
When the GPU memory is almost full, kvcached can support offloading KV cache to CPU memory or even disks. Do this using CUDA UVM or more application semantics?

PR #404, opened 2026-07-22
Add P/D connector support matrix and smoke summaries

Lanoxia:
## Summary

- Add `docs/PD_DISAGGREGATION.md` to track vLLM prefill/decode connector status for kvcached.
- Document the current NIXL smoke-test entrypoint, pass criteria, and known failure signatures from prior P/D investigation.
- Make `tools/run_vllm_nixl_pd_smoke.sh` write one JSON summary line per completed case to `${LOG_DIR}/summary.jsonl`.
- Link the P/D support matrix from the README feature list.

## Motivation

Issue #302 asks whether kvcached supports prefill/decode disaggregation, while #311 captured concrete connector-level failures and open validation gaps. This PR turns that context into a repository-level support matrix and a reusable reporting path so future GPU runs can be posted back to issues or PRs with consistent evidence.

## Validation

- `bash -n tools/run_vllm_nixl_pd_smoke.sh`
- `git diff --check`

Full NIXL validation requires a CUDA/vLLM host with the relevant P/D connector dependencies, so I did not run the end-to-end smoke test on my local macOS environment.

Refs #302


RixinLiu:
Thanks for putting this out! Merging this as the first step toward P/D support rather than a statement about it, with one thing held back and some commits pushed.

## Held back: the README link

P/D disaggregation isn't supported yet, every connector in the matrix is `Untested`, and the one with a harness has never been run end to end. So I dropped the whole README changes.

## Pushed

**`aa4e7fb` — grouped the matrix, dropped `SharedStorageConnector`.** vLLM 0.22.1 registers fifteen connectors; the doc covered six without saying how they were chosen, so a reader couldn't tell a deliberate exclusion from an oversight. Now split into *under validation* / *planned* / *not covered*, with a reason on every out-of-scope row. `OffloadingConnector` and `SimpleCPUOffloadConnector` moved into *planned* since #351 and #93/#269 are already working on them. `SharedStorageConnector` is gone: I grepped 0.18.1 (the version #311 was filed against) and 0.22.1 and it doesn't exist in either.

**`090491e` — fixed two bugs in the summary writer.** The heredoc is `<<'PY'`, so Python received the two characters `\` and `n` instead of a newline:

**`bc652b6`** — the README revert above.

## Next

I'm opening a tracking issue for P/D support and assigning it to you. No rush on it, fill in the matrix cells as you get time and hardware, and once a connector has a reproducible run behind it I'll fold the results into the pinned support matrix in #425 so there's one place to look.

PR #405, opened 2026-07-22
Fix small max cached token limits

### LINKED ISSUE #343: _get_max_cached_blocks truncates small positive KVCACHED_MAX_CACHED_TOKENS to the disabled

Lanoxia:
## Summary

- Fix `_get_max_cached_blocks()` so positive `KVCACHED_MAX_CACHED_TOKENS` values are rounded up to at least one cache block.
- Preserve the existing special cases: negative means unlimited, zero means disabled.
- Add a parameterized unit test for negative, zero, sub-block, exact-block, and non-divisible positive limits.

## Motivation

Before this change, `KVCACHED_MAX_CACHED_TOKENS=8` with `block_size=16` produced `0` cached blocks because the conversion used floor division. That silently disabled prefix caching even though the user requested a small positive cache budget.

## Validation

- `python -m pytest tests/test_prefix_cache.py`
- `python -m pytest tests/test_prefix_cache.py tests/test_make_cache_key.py tests/test_vllm_nixl_compat.py`
- `git diff --check`

I also tried `python -m pytest tests`, but this local macOS environment is missing optional project dependencies such as `torch`, `vllm`, and `aiohttp`, so full collection does not run here.

Fixes #343

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -262,12 +262,14 @@ def _get_max_cached_blocks(block_size: int) -> int:
 
     Returns -1 (unlimited) when MAX_CACHED_TOKENS < 0.
     Returns 0  (disabled — evict on free) when MAX_CACHED_TOKENS == 0.
-    Otherwise returns MAX_CACHED_TOKENS // block_size.
+    Otherwise returns enough blocks to cover MAX_CACHED_TOKENS.
     """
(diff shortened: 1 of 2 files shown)


RixinLiu:
Closed this pr, as #380 predates this one, and has a better solution, see the reason in https://github.com/ovg-project/kvcached/pull/380#issuecomment-5320682942

PR #407, opened 2026-07-23
fix(kv_cache_manager): free reserved blocks before the resize assertion, not after

### LINKED ISSUE #406: resize() crashes instead of degrading gracefully when reserved blocks are held
`KVCacheManager.resize()` crashes on a normal caller sequence instead of degrading gracefully

`resize()` is documented (and its own `NOTE` comment confirms) to support a
graceful-degrade path when the page allocator cannot shrink immediately:
free any outstanding reserved blocks, then enter the lazy `in_shrink` wait
until enough blocks are freed. The current implementation checks the
degrade precondition before doing the thing that establishes it:

```python
# kvcached/kv_cache_manager.py:380-401
if self.page_allocator.resize(new_mem_size):
    ...
    return True
# Failed to resize due to too many in-use blocks.
assert (len(self.reserved_blocks) == 0
        ), "Reserved blocks must be freed before resizing."
# NOTE: we can support resizing with reserved blocks, but we want to
# enforce this check for now to ensure correctness.
self.in_shrink = True
self.target_num_blocks = new_mem_size // self.block_mem_size
self.free_reserved()
return False
```

`free_reserved()` is the only call that clears `reserved_blocks`, and it
runs after the assertion, so it can never satisfy it. Any caller that has
reserved blocks outstanding (via `try_to_reserve()`) when the allocator
can't shrink immediately hits `AssertionError` instead of the documented
degrade path. This is reachable through the public API in a normal
sequence: reserve blocks, then have a `kvctl`-driven size reduction land
while they're still held.

Reproduction (stubs the compiled `kvcached.vmm_ops` extension, since this
box has no GPU to build/run it, and exercises `resize()`/`free_reserved()`
directly on a manually constructed manager):

```python
manager.reserved_blocks = [1, 2, 3]
manager.page_allocator.resize = lambda n: False  # can't shrink now
manager.resize(4096)
# AssertionError: Reserved blocks must be freed before resizing.
```

Verified against current HEAD (`b55096c`) in a clean `python:3.12-slim`
container.

I have a fix ready (call `free_reserved()` before the assertion instead of
after) and will open a PR referencing this issue.

AmirF194:
fix(kv_cache_manager): free reserved blocks before the resize assertion, not after

`KVCacheManager.resize()` is meant to degrade gracefully when the page
allocator cannot shrink immediately: free any outstanding reserved blocks,
then enter the lazy `in_shrink` wait. The current order does the opposite.
It asserts `len(self.reserved_blocks) == 0` first, and only calls
`free_reserved()` afterward, so the one call that could satisfy the
assertion never runs before it fires. Every caller that has reserved blocks
outstanding when a shrink cannot apply immediately (a normal sequence:
`try_to_reserve()` followed by a `kvctl`-driven limit reduction that the
allocator cannot honor right away) hits `AssertionError` instead of the
documented degrade path.

Fix: call `free_reserved()` before checking the assertion, so the
reserved blocks are cleared first and the assertion becomes a genuine
post-condition check instead of a precondition that nothing upstream
of it can satisfy.

Verified in a clean `python:3.12-slim` container. `kv_cache_manager.py`
hard-imports the compiled `kvcached.vmm_ops` CUDA/HIP extension, which this
environment has no GPU to build or run, so the regression test stubs that
extension (a fake `PageAllocator`/`InternalPage`/`kv_tensors_created`
registered in `sys.modules` before import) and constructs the manager
without running `__init__`, to exercise `resize()`/`free_reserved()`
directly. On `b55096c` (current main) the added test fails with the
`AssertionError` above; on this branch it passes. I have not run the
GPU-backed integration suite (`tests/test_kvcache_manager.py`), since this
box has no CUDA device.

Fixes #406

### DIFF
--- a/kvcached/kv_cache_manager.py
+++ b/kvcached/kv_cache_manager.py
@@ -374,8 +374,12 @@ class KVCacheManager:
     @synchronized
     def free_reserved(self):
         if self.reserved_blocks:
-            self.free(self.reserved_blocks)
-            self.reserved_blocks.clear()
+            # Detach first: once off the ledger the blocks are ordinary
+            # blocks, so free()'s sanity check (which forbids freeing a
+            # block that is still reserved) correctly stays quiet. Freeing
+            # before detaching trips that check under KVCACHED_SANITY_CHECK.
+            blocks, self.reserved_blocks = self.reserved_blocks, []
+            self.free(blocks)
 
     @synchronized
     def resize(self, new_mem_size: int):
--- a/kvcached/kv_cache_manager.py
+++ b/kvcached/kv_cache_manager.py
@@ -390,14 +390,15 @@ class KVCacheManager:
                 self.in_shrink = False
                 self.target_num_blocks = None
             return True  # Successfully resized.
-        # Failed to resize due to too many in-use blocks.
-        assert (len(self.reserved_blocks) == 0
-                ), "Reserved blocks must be freed before resizing."
+        # Failed to resize due to too many in-use blocks. Free any
+        # outstanding reserved blocks before entering the lazy shrink wait.
         # NOTE: we can support resizing with reserved blocks, but we want to
         # enforce this check for now to ensure correctness.
+        self.free_reserved()
+        assert (len(self.reserved_blocks) == 0
+                ), "Reserved blocks must be freed before resizing."
         self.in_shrink = True
         self.target_num_blocks = new_mem_size // self.block_mem_size
-        self.free_reserved()
         return False
 
     @synchronized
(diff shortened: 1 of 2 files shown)


RixinLiu:
Thanks for the contribution! The bug is real and the reorder is correct, verified against `main`. LGTM.

One heads-up: your PR surfaced a related pre-existing bug, so I pushed a commit directly to your branch — `free_reserved()` freed blocks *before* clearing the ledger, which made the degrade path still raise `ValueError` under `KVCACHED_SANITY_CHECK=true` (from `free()`'s reserved-block guard). It now detaches first, with a regression test that exercises the real `free()` under the sanity check.

AmirF194:
Thanks for catching and fixing the free-before-clear ordering in `free_reserved()`, and for adding the regression test under the sanity check. Appreciate you taking the time to push it directly rather than routing it back through me.

PR #409, opened 2026-07-23
fix(controller): don't fail the losing racer of a concurrent wakeup_model() call

### LINKED ISSUE #408: SleepManager.wakeup_model() reports failure for a caller whose own wake call actually succeeded, when two wakeups for the same model race
`SleepManager.wakeup_model()` reports failure for a caller whose own wake call actually succeeded, when two wakeups for the same model race

`wakeup_model()` checks membership in `self.sleeping_models`, awaits the real
upstream wake call, then removes the entry:

```python
# controller/sleep_manager.py:150-215
if model_name not in self.sleeping_models:
    return False
...
success = await self._call_vllm_wakeup_api(host, port)
...
del self.sleeping_models[model_name]
self.manual_sleep_models.discard(model_name)
...
return True
```

`self.sleeping_models` is a plain, unsynchronized dict, and the membership
check happens before an `await` boundary. `router.py` dispatches a wakeup
for a sleeping model on every inbound request, and `frontend.py`'s aiohttp
handlers can dispatch concurrently, so two simultaneous requests for the
same sleeping model both pass the `in` check, both issue their own real
upstream wake call, and both reach `del self.sleeping_models[model_name]`.
Only the first `del` succeeds; the second raises `KeyError`, which is
caught by the surrounding `except Exception`, logged as an error, and
turned into a `return False`, even though that caller's own upstream
wake call succeeded.

Reproduction (mocks only the upstream HTTP call, not `sleeping_models`
itself):

```python
manager.sleeping_models[model_name] = time.time() - 10
manager._call_vllm_wakeup_api = slow_wake  # always returns True
results = await asyncio.gather(
    manager.wakeup_model(model_name),
    manager.wakeup_model(model_name),
)
# results == [True, False]
```

Verified against current HEAD (`b55096c`) in a clean `python:3.11-slim`
container.

I have a fix ready (`self.sleeping_models.pop(model_name, None)` instead of
`del`, so the losing racer no longer raises) and will open a PR referencing
this issue.

AmirF194:
Fixes #408.

`wakeup_model()` checks `model_name in self.sleeping_models`, awaits the
real upstream wake call, then does `del self.sleeping_models[model_name]`.
That dict has exactly two writers in the package (`put_model_to_sleep` sets
it, `wakeup_model` deleted it): no lock guards the check-then-mutate across
the `await`. Two concurrent `wakeup_model()` calls for the same model both
pass the membership check, both call their own real upstream wake API, and
both reach the `del`. Only the first succeeds; the second raises
`KeyError`, caught by the surrounding `except Exception`, so that caller
reports `False` even though its own upstream wake call succeeded.

Fix: `self.sleeping_models.pop(model_name, None)` instead of `del`. The
losing racer no longer raises, so both callers reach `return True`, which
is correct since both of their own upstream calls succeeded.

Verified in a clean `python:3.11-slim` container at HEAD (`b55096c`):
`tests/test_wakeup_race.py` fires two concurrent `wakeup_model()` calls
with a mocked (always-succeeding) upstream wake API. Before the fix:
`[True, False]`. After: `[True, True]`. The mock only replaces the
upstream HTTP call, not `sleeping_models` or either of its two writers.

Also ran the repo's own gates on the changed files: `pre-commit run`
(ruff, codespell, isort, SPDX header check, trailing-whitespace, local
mypy) and the CI-only `mypy-3.12` manual hook, all passing.

Not verified: behavior against the real vLLM/SGLang wake APIs (only the
mocked upstream call), and whether two concurrent upstream wake calls to
the same backend are themselves safe to fire together, that's existing
behavior this change does not touch.

### DIFF
--- a/controller/sleep_manager.py
+++ b/controller/sleep_manager.py
@@ -202,7 +202,7 @@ class SleepManager:
                     f"No vLLM or SGLang configuration found for model {model_name}, using fallback behavior"
                 )
 
-            del self.sleeping_models[model_name]
+            self.sleeping_models.pop(model_name, None)
             self.manual_sleep_models.discard(model_name)
 
             logger.info(
--- a/tests/test_wakeup_race.py
+++ b/tests/test_wakeup_race.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright contributors to the kvcached project
+# SPDX-License-Identifier: Apache-2.0
+"""
+Regression test: two concurrent wakeup_model() calls for the same sleeping
+model must both report success when both of their own upstream wake calls
+succeed.
+"""
+
+import asyncio
+import sys
+import time
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent))
+sys.path.insert(0, str(Path(__file__).parent.parent / "controller"))
+
+from controller.sleep_manager import SleepConfig, SleepManager
+
+
+async def test_concurrent_wakeup_both_succeed():
+    model_name = "test-model"
+    config = SleepConfig(
+        min_sleep_duration=0,
+        vllm_models_config={
+            model_name: {
+                "host": "localhost",
+                "port": "8000"
+            }
+        },
+    )
+    manager = SleepManager(config)
+    manager.sleeping_models[model_name] = time.time() - 10
+
+    async def slow_wake(host, port):
+        await asyncio.sleep(0.05)
+        return True
+
+    manager._call_vllm_wakeup_api = slow_wake  # type: ignore[method-assign]
+
+    results = await asyncio.gather(
+        manager.wakeup_model(model_name),
+        manager.wakeup_model(model_name),
+    )
+
+    assert results == [
+        True, True
+    ], f"expected both racers to report success, got {results}"
+    assert model_name not in manager.sleeping_models
+
+
+async def main():
+    await test_concurrent_wakeup_both_succeed()
+    print("OK: concurrent wakeup_model() calls both report success")
+
+
+if __name__ == "__main__":
+    asyncio.run(main())


AmirF194:
No rush, just a check-in after a week. Pre-commit CI is still awaiting approval to run on this fork PR, nothing has executed yet.

AmirF194:
Closing to keep the queue clean; happy to reopen if there's interest.

AmirF194:
Reopening. My previous close was a mistake on my end, this is still open work and I'd like to see it through.

AmirF194:
Closing to keep the queue clean, no maintainer signal in a month. Happy to reopen if there is interest, the branch stays up.

AmirF194:
Sorry, reopened. I closed this by mistake, it is still on the roadmap in #419 and nothing here needed closing.

AmirF194:
Flagging an overlap: #484 adds a per-model lock around both put_model_to_sleep and wakeup_model, which serializes concurrent calls entirely rather than just fixing the losing racer's del/KeyError like this PR does. It also closes #462, the broader missing-synchronization issue I filed alongside this one. If #484 lands, this PR becomes redundant since the race it patches around can't happen anymore under the lock.

Happy to stand down here if you'd rather take #484 instead; otherwise let me know and I can rebase whichever one loses.

rishabhsinha17:
Thanks for the quick reply. My recommendation is #484 as the base with the refinement from my review there folded in, which keeps the thing this PR protects. One upstream call per wake is the point of #462, and the test here encodes the double call, slow_wake runs once per racer and both report True because each fired the full sequence. Under the #484 lock the losing racer returns False from the membership recheck instead, and router.py:149-157 drops that request. The refinement closes that half, handle_model_wakeup_on_request returns True once sleeping_models no longer holds the model after wakeup_model comes back False, and the losing request proceeds because the winner already did the work. Both racers succeed from the router's point of view, one upstream sequence fires, and this PR becomes unnecessary rather than conflicting. If that plan works for you, review input on #484 is the useful place for your expectations, a concurrent test at the handle_model_wakeup_on_request seam asserting both requests proceed with one upstream call would pin the merged behavior better than either PR's current test.

AmirF194:
That plan works. Posted the handle_model_wakeup_on_request gap on #484 with the router.py:149-155 trace, confirmed it's real: the losing racer's False currently gets treated as a wakeup failure there. Closing this one in favor of #484 plus that refinement.

AmirF194:
Reopening. I closed this on your recommendation, but RixinLiu's roadmap in #419 still lists this PR specifically and doesn't mention #484, so I'd rather wait for his call on which one he wants before standing down. The #484 review stands either way.

rishabhsinha17:
Since the reopen, #484 picked up the request-path refinement at 72f5016d, and I approved that head after verifying it. The two PRs now cover compatible halves, the per-model lock there plus the both-requests-proceed behavior this one protects. The #419 listing dates to August 11 and predates #484 entirely, so it records the state before there was anything to choose between, and the canonical-PR call is @RixinLiu's to make, with no need to close this before he weighs in.

RixinLiu:
Sorry for the confusion brought by my outdated roadmap. I will merge #484 and closed this PR

PR #415, opened 2026-07-30
fix(sglang): keep KV pool ownership local to each TP worker

shipiyouniao:
## Summary

- keep SGLang `KVCacheManager` map/unmap ownership local to each TP worker
- preserve the real TP world size for rank-aware kvcached IPC listener setup
- add a CPU regression test covering the ownership split

## Problem

SGLang constructs and drives a KV pool independently in every TP worker. Passing the real TP world size into each `KVCacheManager` makes every worker broadcast its own local map/unmap operations. During null-block reservation, peers receive a second map for their already-mapped local page 0 and emit `Page 0 is already mapped`.

The TP world size is still required by `init_kvcached()` for listener/socket setup, so this change only scopes the manager itself to `world_size=1`.

## Real TP2 validation

Tested on two physical RTX 4090 GPUs with strict Docker GPU isolation:

- image: `docker.1ms.run/lmsysorg/sglang:v0.5.13`
- model: `Qwen3-8B`
- topology: SGLang TP2 on physical GPUs 6 and 7
- PyTorch: `2.11.0+cu130`
- CUDA runtime: `13.0`
- MPS active thread percentage: `100`
- kvcached page preallocation: disabled

Both runs used the same image, model, launch arguments, and environment. The fixed run differed only by overlaying this PR's `kvcached/integration/sglang/interfaces.py`.

| Case | Commit | Manager world size | `Page 0 is already mapped` | Request result |
| --- | --- | ---: | ---: | --- |
| Base | `ce76a129` | 2 | 2 | 1/1 HTTP 200, non-empty `OK.` |
| Fix | `79b8e5d5` | 1 | 0 | 5/5 HTTP 200, non-empty `OK.` |

Neither same-condition run had a scheduler exception, CUDA error, OOM, illegal-memory error, or traceback. The base run remained serviceable, so the verified defect is the duplicate cross-worker map operation and its error, not a demonstrated server outage.

SGLang 0.5.13's idle pool invariant was set to non-fatal warning mode in both runs because the existing `scheduler_memory_leak` compatibility patch does not apply to that SGLang version. This is independent of the ownership change and was held constant across the comparison.

## Focused validation

- `python -m pytest -q tests/test_sglang_local_pool_ownership.py`
- `python -m ruff check kvcached/integration/sglang/interfaces.py tests/test_sglang_local_pool_ownership.py`
- `python -m isort --check-only kvcached/integration/sglang/interfaces.py tests/test_sglang_local_pool_ownership.py`
- repository pre-commit and mypy CI: passed

### DIFF
--- a/tests/test_sglang_local_pool_ownership.py
+++ b/tests/test_sglang_local_pool_ownership.py
@@ -1,0 +1,84 @@
+# SPDX-FileCopyrightText: Copyright contributors to the kvcached project
+# SPDX-License-Identifier: Apache-2.0
+
+import sys
+import types
+
+try:
+    import torch
+except ImportError:
+    torch = types.ModuleType("torch")
+    sys.modules["torch"] = torch
+
+if not hasattr(torch, "Tensor"):
+    torch.Tensor = object
+if not hasattr(torch, "dtype"):
+    torch.dtype = object
+if not hasattr(torch, "cuda"):
+    torch.cuda = types.SimpleNamespace(current_device=lambda: 0)
+
+try:
+    import kvcached.vmm_ops as vmm_ops
+except Exception:  # noqa: BLE001 - any extension import failure uses the CPU stub
+    vmm_ops = types.ModuleType("kvcached.vmm_ops")
+    sys.modules["kvcached.vmm_ops"] = vmm_ops
+
+for name, value in {
+    "PageAllocator": object,
+    "InternalPage": object,
+    "create_kv_tensors": lambda *args, **kwargs: [],
+    "init_kvcached": lambda *args, **kwargs: None,
+    "shutdown_kvcached": lambda: None,
+    "kv_tensors_created": lambda *args, **kwargs: True,
+    "map_to_kv_tensors": lambda *args, **kwargs: None,
+    "unmap_from_kv_tensors": lambda *args, **kwargs: None,
+}.items():
+    if not hasattr(vmm_ops, name):
+        setattr(vmm_ops, name, value)
+
+from kvcached.integration.sglang import interfaces  # noqa: E402
+
+
+def test_sglang_keeps_real_tp_size_for_ipc_but_owns_pool_locally(monkeypatch):
+    initialized = []
+    listeners = []
+    manager_args = []
+
+    class FakeManager:
+        def __init__(self, *args, **kwargs):
+            manager_args.append((args, kwargs))
+
+    monkeypatch.setattr(interfaces, "_kvcached_initialized", False)
+    monkeypatch.setattr(interfaces, "_kvcached_device", None)
+    monkeypatch.setattr(interfaces, "_async_sched", False)
+    monkeypatch.setattr(interfaces, "_world_size", 1)
+    monkeypatch.setattr(interfaces, "_pp_rank", 0)
+    monkeypatch.setattr(interfaces, "_init_kvcached_impl", lambda *args: initialized.append(args))
+    monkeypatch.setattr(
+        interfaces,
+        "start_worker_listener_thread",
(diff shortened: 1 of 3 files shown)


tests/test_sglang_local_pool_ownership.py
```diff
@@ -0,0 +1,70 @@
...
+except ImportError:
+    torch = types.ModuleType("torch")
+    torch.Tensor = object
+    torch.dtype = object
+    torch.cuda = types.SimpleNamespace(current_device=lambda: 0)
+    sys.modules["torch"] = torch
+
+try:
+    import kvcached.vmm_ops  # noqa: F401
+except ImportError:
```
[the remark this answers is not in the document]
shipiyouniao: Fixed in 79b8e5d. The CPU fallback now handles any compiled-extension import failure (including OSError from missing runtime/shared libraries), matching the repository pattern, while retaining the real extension whenever it imports successfully. Focused pytest, ruff, isort, and mypy checks pass.

shipiyouniao:
Moving this back to draft until a real SGLang TP runtime reproduces the duplicate local-pool mapping failure and confirms the ownership split fixes it. The CPU regression test alone is not enough evidence for changing runtime ownership semantics.

shipiyouniao:
Real SGLang TP2 validation is now complete on two RTX 4090 GPUs with SGLang v0.5.13 and Qwen3-8B. Under identical launch conditions, the base commit emitted two duplicate page-0 mapping errors while the fix emitted none; base returned 1/1 valid response and the fix returned 5/5 valid responses. I updated the PR description with the exact environment, commits, caveat, and results, and corrected the claim so it does not imply a verified server outage.

shipiyouniao:
Addressed the fresh Copilot review on the current head by isolating every SGLang module global mutated by `init_kvcached()` (`_kvcached_initialized`, device, async mode, TP world size, and PP rank).

Validation:
- focused pytest: 1 passed
- Linux full `pre-commit run --all-files`: passed
- Linux Python 3.12 focused pytest: 1 passed

Updated head: `53c57330`.

tests/test_sglang_local_pool_ownership.py
```diff
@@ -0,0 +1,84 @@
...
+    monkeypatch.setattr(interfaces, "_kvcached_device", None)
+    monkeypatch.setattr(interfaces, "_async_sched", False)
+    monkeypatch.setattr(interfaces, "_world_size", 1)
+    monkeypatch.setattr(interfaces, "_pp_rank", 0)
+    monkeypatch.setattr(interfaces, "_init_kvcached_impl", lambda *args: initialized.append(args))
+    monkeypatch.setattr(
+        interfaces,
+        "start_worker_listener_thread",
+        lambda tp_rank, pp_rank: listeners.append((tp_rank, pp_rank)),
+    )
```
jeff3071, line 60: This test will fail since #413 change `start_worker_listener_thread` interface.
Mock `device_index` may solve it.

jeff3071:
Run this patch with Qwen3-8B TP=2 on 2xA2000. There is no CUDA or engine failure.

tests/test_sglang_local_pool_ownership.py
shipiyouniao: Thanks, reproduced the `device_index` failure and fixed it in 665677f. The test now checks the actual device index, including TP rank 2 on device 3, `hip:3` normalization, and TP=1. I also rebased onto current main and kept the new Mamba allocator test in the CPU manifest.

The full CPU suite passed on Python 3.9-3.13 (263 tests each), along with MyPy, pre-commit and C++ tests. The rebuilt T4 native/device regression suite passed 104 tests, and all 11 GitHub checks are green. Thanks for the TP2 A2000 check too!

jeff3071 (approved):
LGTM.

PR #420, opened 2026-07-31
Add portable self-hosted GPU CI profiles

Lanoxia:
## Summary

This PR turns the existing GPU validation workflow into repository-owned,
portable CI that can run on any persistent Linux x86-64 NVIDIA host.

- Add a self-hosted GitHub Actions workflow with `core`, `vllm`, `sglang`,
  `engines`, and two-GPU `nixl` profiles.
- Provision isolated core, vLLM, and SGLang Python environments with a
  one-command bootstrap script.
- Add engine correctness smoke tests, CPU-only workflow contract tests, repeat
  support, idle-GPU protection, and a host-wide execution lock.
- Preserve environment snapshots, server/client logs, GPU process state, and
  machine-readable summaries as workflow artifacts on success or failure.
- Restore the GPU resize regression test and make CUDA toolkit stub linking
  portable in containerized environments.
- Document runner registration, repository variables, trigger policy, security
  boundaries, and deployment steps.

## Motivation

GPU tests currently depend on manually prepared local environments. This makes
them difficult to reproduce, schedule, and inspect after a failure. The new
workflow keeps the host-specific responsibility small: a maintainer registers a
persistent GPU machine, while the repository owns environment provisioning,
profile selection, test execution, safety checks, and artifact collection.

Fork pull requests are intentionally prevented from executing arbitrary code on
a persistent lab machine. Same-repository pull requests require an explicit
`gpu-ci` label from a maintainer.

## Validation

- Public pre-commit CI is green across the general hook job and MyPy on Python
  3.9, 3.10, 3.11, 3.12, and 3.13:
  https://github.com/Lanoxia/kvcached/actions/runs/30614464694
- End-to-end GitHub Actions dispatch was validated with an ephemeral macOS
  ARM64 self-hosted runner: push routing, exact-SHA checkout, command
  execution, artifact upload, and automatic deregistration all passed:
  https://github.com/Lanoxia/kvcached/actions/runs/30622646345
- The core profile passed 36 tests per run for five consecutive runs on one
  NVIDIA H20 (180 consecutive test executions).
- The restored resize regression passed on H20.
- The public vLLM/NIXL prefill-decode harness was validated separately on two
  NVIDIA H20 GPUs, both with and without kvcached.
- The workflow/bootstrap/engine contract suite passes locally (`10 passed`).
- `isort --check-only`, actionlint 1.7.7, and `git diff --check` pass.

## Deployment after merge

Register a persistent Linux x86-64 NVIDIA runner with:

```text
self-hosted, linux, x64, gpu, kvcached
```

Then run `bash tools/bootstrap_gpu_ci_envs.sh` and configure the three
repository variables printed by the script:

```text
KVCACHED_GPU_PYTHON
KVCACHED_VLLM_PYTHON
KVCACHED_SGLANG_PYTHON
```

No project-specific cloud, scheduler, notebook service, or private
infrastructure is required.


tools/run_gpu_ci.sh
```diff
@@ -0,0 +1,268 @@
...
+    "status": "passed" if os.environ["EXIT_CODE"] == "0" else "failed",
+}, indent=2, sort_keys=True))
+PY
+  exit "${exit_code}"
+}
+trap finalize EXIT
+
+nvidia-smi --query-gpu=index,name,memory.total,driver_version \
+  --format=csv,noheader | tee "${GPU_CI_ARTIFACT_DIR}/nvidia-smi.txt"
+nvidia-smi --query-compute-apps=pid,process_name,used_memory \
```
shipiyouniao: This runner does not establish which physical GPU(s) belong to the job. The idle check queries the entire host, while PyTorch and the engines later use whichever devices happen to be visible. On a shared multi-GPU host, an unrelated process can reject the job; with `GPU_CI_REQUIRE_IDLE=0`, an external scheduler can reserve GPU N while the tests still default to GPU 0.

Could we make selected devices an explicit workflow/runner contract (for example, `KVCACHED_GPU_VISIBLE_DEVICES` exported as `CUDA_VISIBLE_DEVICES`), scope `nvidia-smi` to those IDs, and assert one visible logical GPU for core/vLLM/SGLang or two for NIXL? The alternative is to document and enforce that the runner service itself exposes only the reserved devices.

Lanoxia: Addressed in 6b10d11. The workflow now requires an explicit KVCACHED_GPU_VISIBLE_DEVICES (or scheduler-provided CUDA_VISIBLE_DEVICES), exports that selection to all subprocesses, scopes nvidia-smi queries to the selected devices, rejects duplicate or incorrect device counts, and verifies the logical GPU count through PyTorch. Core, vLLM, and SGLang require one selected GPU; NIXL requires exactly two. The runner documentation and contract tests were updated as well.

RixinLiu:
Thanks for the GPU CI work, i always want to have sth like this, thanks!

One thing I'd like to add on top: a comprehensive model compatibility matrix as a separate, manually-triggered-only workflow.

The matrix i put it here in #425, if the model is too big, we can choose the smaller one of the same family. 

For each cell: start the engine with kvcached, send a short generation request, verify the output is not garbled. Report a matrix of pass / crash-at-startup / garbled-output.

This is too heavy to run on every PR or even daily, I'd trigger it manually before releases or after major changes. Would you be able to build this as an additional workflow on top of the current GPU CI infrastructure?

This one is not that urgent, you may consider open a new PR for it :)

tools/run_engine_smoke.sh
```diff
@@ -0,0 +1,174 @@
...
+  exit 2
+fi
+"${PYTHON}" -c "import ${ENGINE}" >/dev/null
+
+SERVER_PID=""
+cleanup() {
+  local exit_code=$?
+  set +e
+  if [[ -n "${SERVER_PID}" ]] && kill -0 "${SERVER_PID}" 2>/dev/null; then
+    kill "${SERVER_PID}" 2>/dev/null
```
shipiyouniao: The cleanup only signals `SERVER_PID`, but both vLLM and SGLang can spawn engine/worker child processes. If the parent does not reap them within 30 seconds, the subsequent `kill -9` can orphan GPU workers and leave ports or device memory occupied on this persistent self-hosted runner. Please launch the server in a dedicated process group and terminate that group, then assert that no processes introduced by the smoke remain on the selected GPU(s).

tools/run_engine_smoke.sh
```diff
@@ -0,0 +1,174 @@
...
+PY
+
+if grep -E "Traceback|CUDA error|illegal memory access|core dumped|Segmentation fault" \
+  "${SERVER_LOG}" >/dev/null 2>&1; then
+  echo "Failure signature found in ${ENGINE} server log" >&2
+  tail -n 120 "${SERVER_LOG}" >&2
+  exit 1
+fi
+
+if ! grep -i "kvcached" "${SERVER_LOG}" >/dev/null 2>&1; then
```
shipiyouniao: This is not sufficient activation evidence: any kvcached logger output (including patch discovery, skips, or failures) can satisfy a case-insensitive grep for `kvcached`, while the unpatched engine may still answer the request correctly. Please require a specific successful patch/allocator-initialization signal (and reject failed/skipped patch results), with a negative contract test showing that a correct vanilla-engine response cannot pass.

shipiyouniao (commented):
Re-reviewed the current head. The explicit GPU assignment fix addresses the earlier isolation concern, and the contract suite passes locally. I found two remaining reliability gaps in the engine smoke path.

tools/run_engine_smoke.sh
Lanoxia: Addressed in 6ca21bb. Engine servers now run in a dedicated session/process group; cleanup terminates the full process group, records compute PIDs on the selected GPU before and after the smoke, and fails if any process introduced by the test remains. The new process-group and residual-PID contract coverage passes.

tools/run_engine_smoke.sh
Lanoxia: Addressed in 6ca21bb. vLLM and SGLang now emit an engine-specific readiness marker only after the kvcached allocator initializes successfully, and the smoke requires that exact marker. I also added a negative contract test confirming that a correct vanilla-engine response with only generic kvcached log text cannot pass.

.github/workflows/gpu-tests.yml
```diff
@@ -0,0 +1,108 @@
...
+          EVENT: ${{ github.event_name }}
+          DISPATCH_PROFILE: ${{ inputs.profile }}
+          LABELS: ${{ join(github.event.pull_request.labels.*.name, ' ') }}
+          MIN_MERGES: ${{ vars.KVCACHED_GPU_SCHEDULED_MIN_MERGES || 5 }}
+        run: bash tools/resolve_gpu_ci_plan.sh
+
+  gpu:
+    name: GPU ${{ needs.plan.outputs.profile }}
+    needs: plan
+    if: needs.plan.outputs.run == 'true'
```
shipiyouniao, line 76: [P1] Keep runner authorization outside the PR-controlled plan

The plan checkout uses the pull-request merge ref, so `tools/resolve_gpu_ci_plan.sh` is contributor-controlled. This job then trusts only that script's `run` output. Once GitHub allows the fork workflow to run, a PR can change the script to emit `run=true` without having any `gpu-ci` label. An isolated check with empty labels confirmed that the replacement output satisfies this condition; no real runner was used. A read-only token does not protect the persistent host from the code it executes.

Could we put authorization in a trusted dispatcher or protected approval boundary tied to the reviewed SHA, rather than accepting the checked-out script as the gate? Please include an unlabelled PR that changes the plan script in the regression coverage. Merely moving this condition into another PR-editable file would leave the same issue.

tools/resolve_gpu_ci_plan.sh
```diff
@@ -0,0 +1,103 @@
...
+    exit 2
+    ;;
+esac
+
+# The scheduled slot is a budget, not an obligation: only spend the GPU once
+# enough pull requests have squash-merged since the last successful pass.
+if [[ "${EVENT}" == "schedule" ]]; then
+  last="$(gh api \
+    "repos/${GITHUB_REPOSITORY}/actions/workflows/gpu-tests.yml/runs?event=schedule&status=success&per_page=1" \
+    --jq '.workflow_runs[0].head_sha // empty')"
```
shipiyouniao, line 88: [P2] Count from the last actual GPU pass, not the last successful workflow

A scheduled run where Plan succeeds and GPU is skipped can still have workflow conclusion `success`, so this query selects that skipped run on the next day and advances the baseline. I reproduced one new commit per daily slot for eight slots: every slot counted only one commit and returned `run=false`, despite eight commits since the last real GPU pass. Using the actual GPU-pass SHA made the same script select `all`.

Could we select a run whose GPU job actually completed successfully, or persist the last successful GPU-tested SHA separately? A multi-slot test with successful-but-skipped workflows would catch this; the current plan tests do not exercise that history lookup.

tests/helpers/elasticity.py
```diff
@@ -0,0 +1,197 @@
...
+
+MB = 1024 * 1024
+
+# The same hybrid model the engine smoke tests use, for the same reason: a
+# GQA-only model exercises none of the KV layouts where kvcached's bugs have
+# actually been. Its per-block recurrent state does not fit the default 2MB
+# page, so raise the page size before kvcached initializes -- both engines
+# inherit this environment when they fork their worker.
+DEFAULT_MODEL = os.getenv("KVCACHED_TEST_MODEL", "Qwen/Qwen3.5-4B")
+os.environ.setdefault("KVCACHED_PAGE_SIZE_MB", "4")
```
shipiyouniao, line 41: [P1] Set the page size before importing kvcached

The import of `kvcached.cli.utils` above imports `kvcached.utils`, which eagerly evaluates `PAGE_SIZE`. This later `setdefault` cannot change that cached constant. In a fresh subprocess with no page-size override, importing this helper produced `KVCACHED_PAGE_SIZE_MB=4` but `kvcached.utils.PAGE_SIZE=2097152`. The smoke test exports its setting only inside its own shell, so it does not configure the subsequent elasticity process.

The default hybrid model is precisely why this helper requests 4 MiB, but the in-process allocator can still use 2 MiB. Please supply the page size before launching each elasticity Python process, and add a fresh-process check that the environment and loaded allocator constant agree. This probe checked configuration only, not a real model launch.

tools/run_gpu_ci.sh
```diff
@@ -0,0 +1,384 @@
...
+    MODEL="${MODEL:-}" \
+    LOG_DIR="${GPU_CI_ARTIFACT_DIR}/vllm-${iteration}" \
+      bash tools/run_engine_smoke.sh
+
+    # FlashInfer's JIT shells out to `ninja`, which pip put in the engine
+    # environment's bin directory; invoking the interpreter by path does not
+    # put that directory on PATH.
+    PATH="$(dirname "$(command -v "${VLLM_PYTHON}")"):${PATH}" \
+    ENABLE_KVCACHED=true VLLM_USE_V1=1 \
+      "${VLLM_PYTHON}" tests/test_elastic_serving.py \
```
shipiyouniao, line 365: [P2] Pass the selected model through to the elasticity checks too

The workflow's model input becomes `MODEL`, which the smoke script consumes, but the elasticity helper reads only `KVCACHED_TEST_MODEL`. Both engine profiles therefore switch back to `Qwen/Qwen3.5-4B` for the second half of the run. I confirmed that supplying `MODEL=example/small-model` leaves `elasticity.DEFAULT_MODEL` at the default.

This matters when a runner selects a smaller model or a local model path: the smoke can pass, then elasticity unexpectedly downloads or loads another model and may fail for unrelated capacity/network reasons. Could we propagate the selected model to both elasticity invocations, while retaining their default when the input is blank, and cover the complete profile's model selection in a contract test?

shipiyouniao (changes requested):
Thanks for the follow-up work here. I reviewed the current head and reran the four workflow/engine contract test modules on Linux: 37 passed. A few gaps remain outside that coverage: the runner authorization boundary, the scheduled-run baseline, and two elasticity-test configuration paths.

I checked these with isolated local probes against this revision. The scheduling probe used a disposable Git history and a stubbed run-history response; the authorization check did not execute anything on a GitHub self-hosted runner. No engine/GPU workload was run for this review. Details and suggested regression coverage are inline.

These are within the existing GPU CI scope; I do not think the separate model compatibility matrix needs to block this PR. The PR description also needs a refresh after the fixes, since it still describes forks as excluded and interpreter configuration through repository variables.

PR #423, opened 2026-08-04
Add automated vLLM and SGLang upstream synchronization

Lanoxia:
## Summary

This PR adds repository-owned synchronization for the OVG-maintained vLLM and
SGLang integration branches.

- Synchronize each integration branch with its official upstream `main` on a
  daily schedule or manual dispatch.
- Rebase the small OVG patch stack and update one stable automation branch.
- Run structural compatibility checks against the engine interfaces patched by
  kvcached before pushing any update.
- Open or update one compatibility PR per engine instead of creating duplicate
  PRs.
- Preserve JSON and Markdown reports for conflicts and failed checks so a
  maintainer or repair agent has actionable input.
- Skip an engine cleanly until its OVG repository is configured.

The PR also updates the SGLang allocator compatibility layer for the current
split allocator package and `num_new_pages` calling convention detected during
the upstream exercise.

## Configuration

Set these repository variables after the OVG engine forks are ready:

```text
OVG_VLLM_REPOSITORY
OVG_VLLM_BASE_BRANCH
OVG_SGLANG_REPOSITORY
OVG_SGLANG_BASE_BRANCH
```

The base branches default to `kvcached-main`. Cross-repository writes use the
`OVG_SYNC_TOKEN` secret; dry runs do not receive that token.

## Validation

- 19 synchronization, compatibility, workflow-contract, and SGLang interface
  tests passed locally.
- A real dry run against `Lanoxia/vllm:kvcached-main` rebased cleanly onto the
  latest `vllm-project/vllm:main`; compatibility scan and three plugin tests
  passed.
- A real dry run against `Lanoxia/sglang:kvcached-main` rebased cleanly onto the
  latest `sgl-project/sglang:main`; compatibility scan and three plugin tests
  passed.
- Both exercises completed without pushing temporary automation branches.

## Deployment boundary

The workflow is ready without private infrastructure. Final deployment only
requires the OVG engine fork locations and a token permitted to update those
repositories and open synchronization PRs.


tools/sync_engine_upstream.py
```diff
@@ -0,0 +1,341 @@
...
+                repository, "rev-parse", "HEAD"
+            ).stdout.strip()
+            result.message = (
+                f"The upstream {args.strategy} completed, but a compatibility "
+                "check failed. No branch was pushed."
+            )
+            return result
+
+    result.result_commit = git(repository, "rev-parse", "HEAD").stdout.strip()
+    if args.push:
```
shipiyouniao, line 280: This update path can overwrite compatibility fixes added to an open sync PR. Every run starts from `base_branch`, creates a fresh local `sync_branch`, and the fresh clone has already fetched the current remote automation head; consequently `--force-with-lease` accepts that fetched head as its lease and replaces it, including any human/repair-agent commits added after the previous automated rebase.

That conflicts with the documented repair flow: an open compatibility PR is exactly where upstream-drift fixes and regression tests are likely to be added before merge. Please either continue from/rebase the existing automation branch, or persist the last bot-produced head and refuse/skip the update when the remote branch contains commits not owned by that run. A regression test should cover an existing automation branch with an additional repair commit and verify that a scheduled update preserves it or fails closed rather than overwriting it.

Lanoxia: Addressed in 94dd649. Each automation-owned update records a managed-head marker; subsequent runs verify that the remote tip is still automation-owned and exit with remote-diverged without pushing if a maintainer or repair agent added commits. Pushes now use an explicit expected-SHA force-with-lease as an additional race guard. A regression test appends a human repair commit and confirms the next scheduled sync preserves it unchanged.

tools/sync_engine_upstream.py
```diff
@@ -0,0 +1,397 @@
...
+            raise CommandError(
+                ["git", "show-ref", "--verify", "--quiet", remote_sync_ref],
+                remote_branch.returncode,
+                remote_branch.stdout,
+            )
+
+    git(repository, "checkout", "-b", args.sync_branch)
+    result.result_commit = git(repository, "rev-parse", "HEAD").stdout.strip()
+    if args.strategy == "rebase":
+        integrate_command = ["rebase", upstream_ref]
```
shipiyouniao, line 276: [P1] Preserve repairs recorded only in merge commits

I reproduced this with an integration patch branch merged using `--no-ff`, adding `repair.txt` while completing the merge. After advancing upstream, `--strategy rebase` returned `synced`: the ordinary patch survived, but `repair.txt` disappeared from the sync result even though it remained in the integration base. Plain rebase drops the merge commit and its merge-only changes. Since the scheduled workflow always selects rebase, could we define and test how previously merged integration repairs survive the next sync? A useful regression is two sync cycles with a reviewed merge-resolution change between them. The CLI already offers merge; the missing piece is making the scheduled/promotion policy preserve those repairs.

tools/sync_engine_upstream.py
```diff
@@ -0,0 +1,397 @@
...
+                repository, "rev-parse", remote_sync_ref
+            ).stdout.strip()
+            remote_message = git(
+                repository,
+                "show",
+                "-s",
+                "--format=%B",
+                remote_sync_ref,
+            ).stdout.splitlines()
+            if SYNC_MANAGED_TRAILER not in remote_message:
```
shipiyouniao: [P2] Check more than the retained commit-message marker

There is an additional case beyond the appended-commit test: run a managed sync, add `repair.txt` to its tip with `git commit --amend --no-edit`, then advance upstream and run sync again. The marker survives the amend, so the update returns `synced` and removes the repair. I reproduced that loss. The explicit lease cannot prevent it because the expected SHA was fetched after the amend. Could the ownership check verify the generated content/history as well, with a regression asserting that an amended repair is preserved or the update stops without pushing?

tools/sync_engine_upstream.py
```diff
@@ -0,0 +1,397 @@
...
+                "check failed. No branch was pushed."
+            )
+            return result
+
+    if args.push:
+        if args.update_existing_branch:
+            git(
+                repository,
+                "commit",
+                "--allow-empty",
```
shipiyouniao: [P2] Avoid rewriting an unchanged pending sync

With a sync branch still pending, rerunning after the commit timestamp changes but without changing either upstream or the integration base creates and pushes a new marker commit. I reproduced different head SHAs with identical trees. The initial ancestor check only looks at the integration base, so it does not recognize this already-generated update. This can retrigger downstream CI and invalidate approvals every day without a code change. Could we add a repeated-identical-input test and preserve the existing branch SHA in that case? PR creation should still be recoverable if an earlier run pushed successfully but failed before opening the PR.

.github/workflows/engine-upstream-sync.yml
```diff
@@ -0,0 +1,185 @@
...
+            exit 0
+          fi
+          repository="${TARGET#https://github.com/}"
+          repository="${repository%.git}"
+          branch="automation/upstream-${ENGINE}"
+          existing="$(
+            gh pr list \
+              --repo "${repository}" \
+              --state open \
+              --head "${branch}" \
```
shipiyouniao, line 174: [P2] Validate the existing PR's base before replacing its branch

The manual input allows changing `base_branch`, but all runs for an engine share `automation/upstream-${ENGINE}`. If that branch already has an open PR targeting a different base, the sync step replaces its contents first; this lookup then finds the old PR by head alone, and `gh pr edit` changes only its body. The new update is therefore still proposed against the old base. Could we check the existing PR base before pushing and stop on a mismatch, or isolate sync branches by base? Please cover an existing PR for base A followed by a dispatch selecting base B.

shipiyouniao (changes requested):
Thanks for adding the protection for appended repair commits. I rechecked the current head and exercised the sync script with disposable local Git repositories. Two repair-loss cases remain: an amended automation marker and a repair recorded only in a merge commit. I also reproduced an unchanged run rewriting the branch; the workflow's base-override handling needs a check too.

The inline comments include the reproduction steps and suggested regression coverage. These are findings against the current PR, not a request to adopt a particular replacement strategy. I have not pushed the local rebase or any implementation changes to your branch.

shipiyouniao (approved):
Addressed the four findings from my previous review in 9a84f16:

- Scheduled sync now merges upstream, preserving repairs recorded only in merge commits. Explicit rebase stops on merge history instead of silently dropping those repairs.
- The branch guard checks the recorded tree and parent, so an amended repair is preserved. A repaired branch already merged into the integration base no longer blocks the next sync.
- Identical inputs reuse the pending commit, while still rerunning checks and allowing failed PR creation to recover.
- The existing PR base is checked before pushing and before publication; workflow runs are serialized.

Also added the missing vLLM worker-method checks and clarified the limits of the static scan. The upstream conflict is resolved, and the original author's history is preserved.

Validation on the final commit: 326 CPU tests passed on each Python version from 3.9 through 3.13, plus all MyPy checks, GPU-free C++ tests, and pre-commit. All 11 GitHub checks are green. This validates the sync tooling, not a live end-to-end engine adaptation deployment.

The review threads are resolved. @RixinLiu, could you double-check the updated sync policy and fixes before merging?

PR #427, opened 2026-08-06
fix(vllm): keep prefix caching enabled for small positive KVCACHED_MAX_CACHED_TOKENS

### LINKED ISSUE #343: _get_max_cached_blocks truncates small positive KVCACHED_MAX_CACHED_TOKENS to the disabled

rishabhsinha17:
## Summary

Keep prefix caching enabled when `KVCACHED_MAX_CACHED_TOKENS` is positive but smaller than `block_size`.

Fixes #343.

## Root cause

`_get_max_cached_blocks()` computes `MAX_CACHED_TOKENS // block_size`. A positive budget smaller than `block_size` (e.g. `KVCACHED_MAX_CACHED_TOKENS=8` with `block_size=16`) integer-divides to `0`, which is indistinguishable from the `== 0` "disabled — evict on free" sentinel, so prefix caching is silently turned off even though the user asked for a small positive budget.

## Changes

- floor the positive-budget result at one block: `max(1, MAX_CACHED_TOKENS // block_size)`, as suggested in the issue
- the `< 0` (unlimited) and `== 0` (explicitly disabled) sentinels are unchanged
- add `tests/test_max_cached_blocks.py`, a CPU-only unit test in the same style as `tests/test_make_cache_key.py` (no vLLM, no `vmm_ops` extension): sentinel passthrough, exact multiples, rounding down, and the regression cases `1, 8, 15` tokens with `block_size=16` → `1` block (previously `0`)

## Validation

```bash
python -m pytest tests/test_max_cached_blocks.py -q  # 7 passed
```

`ruff check` and `isort --check-only` pass on the changed files.

Related: #380 and #405 also address this issue.

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -306,12 +306,16 @@ def _get_max_cached_blocks(block_size: int) -> int:
 
     Returns -1 (unlimited) when MAX_CACHED_TOKENS < 0.
     Returns 0  (disabled — evict on free) when MAX_CACHED_TOKENS == 0.
-    Otherwise returns MAX_CACHED_TOKENS // block_size.
+    Otherwise returns MAX_CACHED_TOKENS // block_size, floored at 1 so that a
+    positive token budget smaller than block_size still keeps prefix caching
+    enabled instead of silently disabling it.
     """
     from kvcached.utils import MAX_CACHED_TOKENS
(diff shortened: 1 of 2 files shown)


rishabhsinha17:
Closing in favor of #380, which predates this PR and takes the same approach to #343. (#405 also addresses it.)

PR #429, opened 2026-08-06
fix(sglang): move free indices to CPU before torch.unique in paged allocator free path

### LINKED ISSUE #382: SGLang paged allocator free path should move indices to CPU before torch.unique
### Problem

`ElasticPagedTokenToKVPoolAllocator.free()` computes page ids from `free_index` and then converts the result to Python indices for `kvcached_allocator.free()`.

When `free_index` is a GPU tensor, doing the unique/conversion path directly can couple allocator free bookkeeping to CUDA device execution and synchronization behavior. This is fragile for high-concurrency SGLang serving paths.

### Expected behavior

Move the free index tensor to CPU before deriving unique page ids and converting them to Python indices. The allocator backend only needs page id integers for bookkeeping, so the Python conversion should happen from CPU data.

rishabhsinha17:
## Summary

Derive free page ids from CPU data in the SGLang paged allocator free path.

Fixes #382.

## Root cause

`ElasticPagedTokenToKVPoolAllocator.free()` ran `torch.unique(free_index // self.page_size)` on the possibly-GPU `free_index` tensor and only then converted the result to Python ints. That couples allocator free bookkeeping to CUDA device execution and synchronization behavior, which is fragile on high-concurrency SGLang serving paths. The allocator backend only needs page-id integers for bookkeeping.

## Changes

- move the tensor to CPU before deriving unique page ids:

  ```python
  free_index_cpu = free_index.cpu()
  return torch.unique(free_index_cpu // page_size).tolist()
  ```

- host the conversion in a module-level `_free_page_ids(free_index, page_size)` helper rather than inline in the injected class, so it is unit-testable without an installed SGLang or a GPU — the same pattern the vLLM integration uses for `_make_cache_key`
- drop the previous broad `try/except` int-conversion fallback: a CPU tensor's `.tolist()` is the single well-defined path

## Validation

Added `tests/test_free_page_ids.py` (CPU-only): dedup + sorted page ids across pages, partial-page frees, empty input, plain-`int` results, `page_size` ∈ {1, 2, 16, 64}, and a spy-tensor test asserting the conversion happens on a `.cpu()` copy of the input.

```bash
python -m pytest tests/test_free_page_ids.py -q  # 10 passed
```

`ruff check` and `isort --check-only` pass on the changed files.

Related: #384 also addresses this issue.

### DIFF
--- a/kvcached/integration/sglang/patches.py
+++ b/kvcached/integration/sglang/patches.py
@@ -27,6 +27,24 @@ def _is_supported_gpu_device(device: str) -> bool:
     return device_str.startswith("cuda") or device_str.startswith("hip")
 
 
+def _free_page_ids(free_index: Any, page_size: int) -> List[int]:
+    """Derive the unique page ids covered by ``free_index`` as Python ints.
+
+    The allocator backend only needs page-id integers for bookkeeping, so the
+    tensor is moved to CPU *before* the unique/int conversion. Running
+    ``torch.unique`` on a GPU ``free_index`` would couple free bookkeeping to
+    CUDA execution and synchronization behavior, which is fragile on
+    high-concurrency serving paths.
+
+    Module-level (rather than a method on the injected allocator class) so it
+    is unit-testable without an installed SGLang or a GPU.
+    """
+    import torch
+
+    free_index_cpu = free_index.cpu()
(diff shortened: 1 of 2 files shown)


rishabhsinha17:
Closing in favor of #384, which predates this PR and takes the same approach to #382.

PR #432, opened 2026-08-06
test: make test_sleep_manager.py runnable under pytest via pytest-asyncio

### LINKED ISSUE #431: tests/test_sleep_manager.py async tests can never run under pytest — no pytest-asyncio configuration
### Problem

`pytest tests/test_sleep_manager.py` fails in every environment because the project has no async test support:

```
FAILED tests/test_sleep_manager.py::test_basic_functionality - Failed: async def functions are not natively supported.
ERROR tests/test_sleep_manager.py::test_real_vllm_instances
ERROR tests/test_sleep_manager.py::test_sglang_configuration
...
1 failed, 8 errors in 1.54s
```

Two causes:

1. pytest collects the `async def test_*` functions but neither `pytest-asyncio` nor any other async plugin is installed or configured, so the first test fails with "async def functions are not natively supported".
2. The remaining step functions take a `manager` parameter, which pytest resolves as a fixture that does not exist (`fixture 'manager' not found`) — the file was written as a sequential script that shares one manager through `main()`, not as independent pytest tests.

Consequences:

- `controller/sleep_manager.py` has zero automated test coverage in any pytest run, even though a test file for it exists.
- Async behavior has to be tested through workarounds: #409 tests a `wakeup_model()` race by wrapping `asyncio.gather` in `asyncio.run()` inside a synchronous test, because a native `async def` test cannot run.
- The CPU-only CI efforts (#395, #403, #428) all have to exclude this file wholesale.

### Expected behavior

- `pytest tests/test_sleep_manager.py` runs: async tests execute under `pytest-asyncio`, and a `manager` fixture provides a fresh `SleepManager` per test.
- Tests are self-contained and degrade gracefully when no live vLLM/SGLang endpoint is reachable (the `SleepManager` HTTP helpers already catch exceptions and return `False`, so this only requires the tests to register their own models).
- The existing live-server script mode (`python tests/test_sleep_manager.py`) keeps working unchanged.

I have a fix ready and will open a PR referencing this issue.

rishabhsinha17:
## Summary

Make `tests/test_sleep_manager.py` actually runnable under pytest: `9 passed` where every environment previously got `1 failed, 8 errors`.

Fixes #431.

## Problem

The file is written as a sequential live-server script (`main()` shares one `SleepManager` across steps), but pytest also collects its `async def test_*` functions. Without an async plugin the first test fails with "async def functions are not natively supported", and the step functions' `manager` parameter is resolved as a nonexistent fixture, erroring the other eight. As a result `controller/sleep_manager.py` has zero automated coverage, and async behavior has to be tested through workarounds — #409 wraps `asyncio.gather` in `asyncio.run()` inside a sync test for exactly this reason.

## Changes

- add `pytest-asyncio` to `requirements.txt` and configure `asyncio_mode = "auto"` in `pyproject.toml`, so plain `async def` tests and fixtures run without per-test markers (and future PRs like #409 can write native async tests)
- add a `manager` fixture providing a fresh `SleepManager` per test; `main()` keeps its own shared manager, so the live-server script mode is unchanged
- make each step self-contained: the sleep/wake and detection tests register their own models from `example-config.yaml` instead of relying on state left behind by earlier steps (`SleepManager`'s HTTP helpers already catch connection failures and return `False`, so these degrade gracefully and quickly when no live endpoint is reachable; the fresh-manager path previously crashed with `TypeError: argument of type 'NoneType' is not iterable` in the SGLang detection step)
- convert deterministic print-only checks into assertions (config defaults, model registry add/remove, config updates, API-method presence). This surfaced that two of the printed "checks" referenced methods that do not exist — `_call_vllm_wake_api` and `handle_model_wakeup` printed `False` and could never fail; the assertions now cover the real `_call_vllm_wakeup_api` and `handle_model_wakeup_on_request`

## Validation

```bash
python -m pytest tests/test_sleep_manager.py -q   # 9 passed in 1.86s   (before: 1 failed, 8 errors)
python tests/test_sleep_manager.py                # script mode intact, exit 0
```

The full CPU-only suite passes with `asyncio_mode = "auto"` enabled (82 passed, 1 skipped — the 73 previously-passing tests plus these 9). No live vLLM/SGLang endpoints are required; with servers configured, the script path exercises the real sleep/wake cycle exactly as before. `ruff check`, `isort --check-only`, and mypy pass on the changed files.

Once this and #428 (or #403) land, the CPU CI job can stop excluding this file.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -61,6 +61,10 @@ dev = [
     "wheel",
 ]
 
+[tool.pytest.ini_options]
+# Run `async def` tests and fixtures on asyncio without per-test markers.
+asyncio_mode = "auto"
+
 [tool.setuptools]
 include-package-data = true
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,5 @@ setuptools>=64
 wheel
 wrapt
 pytest
+pytest-asyncio
 posix_ipc>=1.0
--- a/tests/manifests/cpu.txt
+++ b/tests/manifests/cpu.txt
@@ -10,6 +10,7 @@ tests/test_prefix_cache.py
 tests/test_resize_reserved_order.py
 tests/test_sglang_allocator_patch.py
 tests/test_shm_info_tracker.py
+tests/test_sleep_manager.py
 tests/test_test_classification.py
 tests/test_vllm_nixl_compat.py
 tests/test_vllm_tp_world_size.py
--- a/tests/manifests/integration.txt
+++ b/tests/manifests/integration.txt
@@ -1,6 +1,5 @@
 tests/test_elastic_serving.py
 tests/test_offline_serving.py
 tests/test_prealloc_gil_deadlock.py
-tests/test_sleep_manager.py
 tests/test_traffic_monitor.py
 tests/test_utils.py
--- a/tests/test_sleep_manager.py
+++ b/tests/test_sleep_manager.py
@@ -11,6 +11,8 @@ import sys
 from pathlib import Path
 from typing import Dict, List
 
+import pytest
+
 # Add the controller directory to the path
 sys.path.insert(0, str(Path(__file__).parent.parent))
 sys.path.insert(0, str(Path(__file__).parent.parent / "controller"))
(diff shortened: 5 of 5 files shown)


rishabhsinha17:
Rebased onto main and folded in the manifest side of this, now that #403 has landed.

`tests/test_sleep_manager.py` sat in `integration.txt` because its async tests required live vLLM/SGLang endpoints. With the changes here they are self-contained (each test registers its own models, and the HTTP helpers already return False on connection failure), so the file moves `integration -> cpu` and its 9 tests join the CPU run.

- `tools/check_test_classification.py`: total=26, cpu=16, gpu=5, integration=5
- `pytest tests/test_sleep_manager.py`: 9 passed
- full cpu manifest locally: 119 passed, with only `test_shm_info_tracker.py` erroring because `/dev/shm` does not exist on macOS

Closed #428 in favor of #403; this is the part that did not overlap.

RixinLiu:
LGTM

PR #433, opened 2026-08-06
fix(vllm): support per-token-head KV quantization (fp8_per_token_head) via cache_dtype_str forwarding

### LINKED ISSUE #424: kvcached garbles output with per-token-head KV cache quantization (fp8_per_token_head / nvfp4)
`patches.py:get_kv_cache_shape` does not pass `cache_dtype_str` to the upstream helper. Per-token-head quantization modes (fp8_per_token_head, int8_per_token_head, nvfp4) widen `head_size` by a few elements to inline per-head fp32 scales into the KV page. Without `cache_dtype_str`, kvcached computes the un-widened shape, so every page stride is wrong and output is garbled.

Per-tensor fp8 (`--kv-cache-dtype fp8`) is unaffected — its scale lives outside the page and does not change the shape.

Reproduces on any model with `--kv-cache-dtype fp8_per_token_head`. Model-independent — the bug is in kvcached's shape calculation, not in model-specific code.

Fix: pass `cache_config.cache_dtype` through to `get_kv_cache_shape` so the widened head_size is included. Short-term alternative: fail loud on per-token-head / nvfp4 until the shape path is fixed.

rishabhsinha17:
## Summary

Fixes #424 for the non-contiguous layout, and fails loud on the contiguous layout where a second, independent bug remains.

With this PR, `--kv-cache-dtype fp8_per_token_head` + kvcached (`KVCACHED_CONTIGUOUS_LAYOUT=false`) produces output **identical to vLLM without kvcached**. The default contiguous layout raises a clear `NotImplementedError` at startup instead of silently garbling.

## Root cause (as diagnosed in #424)

Per-token-head quantization modes (`fp8_per_token_head`, `int8_per_token_head`, `nvfp4`) widen `head_size` to inline per-token-head fp32 scales into the KV page (e.g. 128 → 132 for fp8). kvcached called `get_kv_cache_shape` without `cache_dtype_str`, computing the un-widened shape, so every page stride was wrong.

## Changes

- forward `cache_config.cache_dtype` at all three `get_kv_cache_shape` call sites, through a new module-level `_get_kv_cache_shape_compat()` that inspects the backend's signature and omits the argument on older vLLM versions that do not declare it (it exists since vLLM 0.20)
- `_cache_dtype_str()` extracts the dtype string from the model runner (`cache_config` directly or via `vllm_config`)
- **contiguous-layout guard**: with inline-scale modes the widened K+V block bytes (e.g. 33,792) no longer divide the physical page size, and the contiguous layout's linear block math breaks *independently of the tensor strides* — output stays garbled even with the corrected shape (details below). Until that math is scale-aware, raise `NotImplementedError` pointing at `KVCACHED_CONTIGUOUS_LAYOUT=false`, mirroring the existing HYBRID_LINEAR page-identity guard. Detection prefers vLLM's own `kv_cache_uses_per_token_head_scales` with a string-heuristic fallback.

## Verification (L40S, vLLM 0.20.0, `VLLM_ATTENTION_BACKEND=TRITON_ATTN`, Qwen3-0.6B, greedy)

| Run | kvcached | Layout | Result |
|---|---|---|---|
| baseline | off | — | clean |
| main | on | non-contiguous | garbled (`" Paris!!!!!!!…"`) |
| **this PR** | on | non-contiguous | **clean — token-identical to baseline** |
| main / shape-fix only | on | contiguous | garbled (identical junk in both, i.e. independent of the shape fix) |
| **this PR** | on | contiguous | loud `NotImplementedError` at startup |
| this PR, `auto` | on | contiguous | clean (regression) |
| this PR, `fp8` per-tensor | on | contiguous | clean (regression) |

Debug instrumentation during the investigation confirmed the widened shape is correctly computed and bound on this PR (`raw0.shape=(50269, 2, 16, 8, 132)`, contiguous inner strides `(…, 1056, 132, 1)`), which is how the contiguous-layout failure was isolated as a separate bug: the garbled output there is byte-identical with and without the widened binding.

The contiguous-layout root cause is worth its own issue — the widened block size is the only supported mode where K+V block bytes are not a power of two (33,792 vs 32,768/65,536), so blocks straddle physical page boundaries in the linear layout. I can file it with the evidence above.

## Tests

`tests/test_kv_cache_shape_compat.py` (CPU-only, no vLLM needed): forwarding to modern signatures, fallback on legacy signatures without `TypeError`, `None`/`auto` passthrough, dtype-string extraction paths, and inline-scale detection for all mode strings.

```
$ python -m pytest tests/test_kv_cache_shape_compat.py -q
8 passed
```

`ruff check`, `isort --check-only`, and the full CPU-only suite pass on the changed files.

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -7,6 +7,7 @@ vLLM-specific patches using unified patch infrastructure.
 
 from __future__ import annotations
 
+import inspect
 import types
 from collections import OrderedDict
 from typing import TYPE_CHECKING, Any, Iterable, Optional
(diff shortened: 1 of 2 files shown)


RixinLiu:
Verified and merging, thanks for tracking this down.

Reproduced on vLLM 0.24.0 / L40 / `TRITON_ATTN` / Qwen3-0.6B: without the forwarding the backend returns `(…, 128)` instead of `(…, 132)`, every page stride is wrong, and output is garbled. With it the shape is correct.

Two things I'm handling in a follow-up rather than blocking on:

**The guard's stated cause isn't right.** The comment blames the widened block bytes no longer dividing the physical page. I tested that directly: `KVCACHED_PAGE_SIZE_MB=66` makes 16,896 divide the page exactly, no straddling block at all, and the output is still garbled, byte-identical to the 2 MiB run. The real cause is that vLLM builds its scale address map from `untyped_storage()`, ignoring `stride()` and `storage_offset()`.

**The non-contiguous path isn't fully fixed either**, it only looks fixed at low block ids. Same root cause; details in the follow-up.

Worth knowing: `test_inline_scales_detection` fails on any machine with vLLM installed, because the two branches of `_kv_cache_uses_inline_scales` disagree on `nvfp4`, vLLM's `kv_cache_uses_per_token_head_scales` returns `False`, the string fallback returns `True`.

Follow-up PR: #<TBD>, closes #434.

Issue #425, opened 2026-08-05
kvcached support matrix


RixinLiu:
We need a comprehensive compatibility matrix covering the model architectures kvcached interacts with. This tracks what works, what crashes, and what produces wrong output, so users know what to expect and contributors know where to focus.

## Test models

| Architecture | Model | Why |
|---|---|---|
| Standard GQA | Qwen3-8B | Baseline — most common setup |
| SWA + dense hybrid | GPT-OSS-20B | Alternating sliding-window / full-attention layers |
| Hybrid linear | Qwen3.5-9B | 3:1 GDN hybrid, some layers have no KV |
| Cross-layer KV sharing | Gemma-4-E2B-it | Layers share KV with earlier layers |
| Heterogeneous KV groups | Gemma-4-12B-it | Per-layer KV head count varies |

## vLLM

| Model | contiguous | non-contiguous |
|---|:-:|:-:|
| Qwen3-8B | | |
| GPT-OSS-20B | | |
| Qwen3.5-9B | | |
| Gemma-4-E2B-it | | |
| Gemma-4-12B-it | | |

## SGLang

| Model | contiguous | non-contiguous |
|---|:-:|:-:|
| Qwen3-8B | | |
| GPT-OSS-20B | | |
| Qwen3.5-9B | | |
| Gemma-4-E2B-it | | |
| Gemma-4-12B-it | | |

Legend: ✅ pass / ❌💥 crash at startup (link issue #) / ❌🗑️ garbled output (link issue #) / ➖ N/A

## How to contribute

Pick a blank cell, run the model with kvcached enabled, and report the result. A single `--max-model-len 1024` short generation is enough to fill a cell.

jeff3071:
I tested SGLang with KVCACHED_CONTIGUOUS_LAYOUT=true

Env:
  Commit: [666671b](https://github.com/ovg-project/kvcached/commit/666671bd8f5eba1ebe14116ee2b04d3b2dbff9da)
  SGLang: 0.5.12
  Python: 3.12.3
  OS: Linux 6.17.0-29-generic / Ubuntu 24.04.1, x86_64
  GPU: NVIDIA RTX PRO 6000 Blackwell 


|Model| Result|
|---|---|
|Qwen/Qwen3-8B|PASS|
|openai/gpt-oss-20b|PASS|
|Qwen/Qwen3.5-9B|PASS|
|google/gemma-4-E2B-it|PASS|

[...]

PR #440, opened 2026-08-10
Add manually triggered model compatibility matrix

Lanoxia:
## Summary

- Add a separate, manually triggered model compatibility workflow for #425.
- Cover the five requested model architectures across vLLM, SGLang, contiguous, and non-contiguous layouts.
- Run cells sequentially so model downloads are reused and one failure does not stop the remaining matrix.
- Record pass, crash-at-startup, or garbled-output for every cell, with JSON/Markdown summaries and per-cell logs.
- Allow filters and a smaller same-family model override for limited GPU memory.

## Validation

- 21 matrix runner, workflow contract, engine smoke, and GPU CI contract tests passed.
- Ruff, codespell, isort, clang-format, PyMarkdown, actionlint, whitespace, and mypy hooks pass.
- Full 20-cell CPU-only preflight passes.
- Synthetic pass, crash-at-startup, and garbled-output paths produce complete reports and continue across failed cells.

## Dependency

This PR is stacked on #420 because it reuses the portable GPU CI runner, environment provisioning, locking, and artifact collection introduced there. The compatibility-matrix commit is ready for review now; after #420 merges, this branch will be rebased onto the updated main so the final diff contains only the matrix changes.

Refs #425


tools/run_model_compatibility_matrix.py
```diff
@@ -0,0 +1,282 @@
...
+            ["bash", str(args.smoke_script)],
+            cwd=ROOT,
+            env=env,
+            stdout=output,
+            stderr=subprocess.STDOUT,
+            text=True,
+        )
+    duration = round(time.monotonic() - started, 3)
+    if result_file.exists():
+        result = json.loads(result_file.read_text(encoding="utf-8"))
```
shipiyouniao, line 163: [P1] Don't reuse a previous cell result as the current result

cell_dir is reused and result.json is not removed before launching the smoke test. If a previous run wrote a PASS and the new process exits before writing its result, this branch reads the old PASS and ignores completed.returncode. I reproduced child exit code 1 returning status='pass', exit_code=0 by leaving a previous result in place. This can also defeat --fail-on-non-pass. Please isolate or remove stale results before each run and reconcile the reported result with the actual subprocess exit status. A regression test should cover rerunning the same cell after a successful run.

tools/run_model_compatibility_matrix.py
```diff
@@ -0,0 +1,282 @@
...
+
+    args.artifact_dir.mkdir(parents=True, exist_ok=True)
+    results = []
+    for index, (entry, engine, layout) in enumerate(cells, start=1):
+        print(
+            f"[{index}/{len(cells)}] engine={engine} model={entry['model']} "
+            f"layout={layout}",
+            flush=True,
+        )
+        result = run_cell(
```
shipiyouniao, line 251: [P2] Keep malformed cell results from aborting the whole matrix

run_cell() directly parses result.json and raises for unsupported statuses, while this loop has no per-cell error handling. A truncated file containing just '{' reproduced JSONDecodeError escaping run_cell(); the outer handler then exits before later cells or the final summary run. Please record a failed cell with the diagnostic and continue with the remaining cells. Tests should cover malformed JSON and an unsupported status, and verify that the final report is still produced.

shipiyouniao (changes requested):
The matrix is useful, but its results need to remain trustworthy when a cell fails or is rerun. I reproduced two failure-handling issues against this revision using CPU-only probes of run_cell(): a previous PASS survives a new child-process failure, and a truncated result.json raises out of the cell instead of letting the matrix continue. These are runner-level reproductions, not GPU/model failures.

Could you add regression tests for both cases, including checking that later cells still execute and the final report retains the failed cell?

PR #450, opened 2026-08-17
Add page-granular CPU offloading foundation

Lanoxia:
## Summary

- Add page-granular CPU offloading while preserving each kvcached virtual page ID.
- Store complete KV page bundles in bounded pinned CPU memory and restore them transactionally.
- Add PageAllocator and KVCacheManager offload/restore state, transition guards, memory accounting, and Python bindings.
- Add policy and planning primitives that exclude active pages, respect CPU capacity, and estimate transfer-versus-recompute cost.
- Add provider-neutral GPU validation, benchmarks, a self-hosted workflow, and reproducible artifact collection.

## Safety model

Offload commits a complete CPU copy before releasing the GPU mapping. Restore maps the same virtual page ID, copies the complete payload back, and only then makes the page available again. Failed restore operations retain the CPU copy and roll back the GPU mapping.

## Validation

- CPU Tests pass on Python 3.9, 3.10, 3.11, 3.12, and 3.13.
- Pre-commit, GPU-free C++ tests, and MyPy on Python 3.9-3.13 pass.
- 26 focused CPU-offload policy, geometry, workflow, and binding-contract tests pass.
- Real VMM validation passed on two NVIDIA H20 GPUs: 80 offload/restore round trips per GPU, byte-for-byte restoration, 256 MiB reclaimed per cycle, and dual-GPU isolation.
- Measured transfer throughput was approximately 50 GB/s device-to-host and 54 GB/s host-to-device.

## Scope

This PR provides the kvcached page-level data path, control plane, safety invariants, and validation infrastructure. Engine-specific automatic victim discovery and pressure-policy integration can build on these explicit page offload/restore primitives in follow-up work.


shipiyouniao (commented):
Thanks for including the design doc and the H20 round-trip results. Could you rebase this onto current main before we do the detailed review? There are conflicts in PageAllocator, KVCacheManager and setup.py, and the transaction and extension-binding changes from #418 and #497 are relevant to this path.

The existing `CPU_OFFLOAD_DESIGN.md` is a useful starting point. A short walkthrough with two sequence diagrams would make this much easier to review:

- Normal offload/restore: which thread owns the page, how in-flight worker work is drained, when the transfer completes, and when mappings, accounting and block availability change.
- Failure/recovery: what happens after a partial copy or map, an unmap failure, or a missing worker response; which CPU/GPU copy remains owned and how retry or cleanup works. Please mark any TP/PP cases that are outside this PR's scope.

After rebasing, please tie the tests to those transitions, including injected failures, not just successful round trips. We can then review the current data path and its recovery behavior together without chasing an outdated diff. No need for a separate large document; extending the existing one is enough.

PR #451, opened 2026-08-18
fix(vllm): correct heterogeneous prefix-cache hashes

### LINKED ISSUE #446: Prefix caching makes gemma-4 generations diverge from vanilla vLLM (output stays coherent, not garbled)
### Summary

With prefix caching on, kvcached returns **different generations for the same prompt within a single run** on `google/gemma-4-12B-it`. Turning prefix caching off makes kvcached byte-identical to vanilla vLLM. A model without sliding-window layers is unaffected.

### Symptom

Environment: vLLM 0.24.0, transformers 5.10.4, torch 2.11.0+cu130, Python 3.12, kvcached `main` @ 315bf54, A100-SXM4-40GB (driver 580.173.02). `google/gemma-4-12B-it`: 48 layers = 40 `sliding_attention` + 8 `full_attention`, `sliding_window=1024`.

The workload sends one 1464-token prompt four times. Greedy, `max_num_seqs=1`, so batch composition is fixed.

| | 1st | 2nd | 3rd | 4th |
|---|---|---|---|---|
| vanilla vLLM | A | A | A | A |
| kvcached | A | A | **B** | **B** |

A = "The topic that repeats above is the history of computing being long and full of detours."
B = "The topic that repeats above is the history of computing."

Identical prompt, greedy, no batching variation — the answer must not depend on what ran before it.

### Isolation

23 requests / 22,236 prompt tokens, `temperature=0`, `max_tokens=64`, `enforce_eager`, A100-40GB. Cells are "requests whose generated token ids differ from the no-kvcached baseline, out of 23".

| | prefix caching on | prefix caching off |
|---|---|---|
| concurrent | 12/23 | **0/23** |
| `max_num_seqs=1` | 5/23 | **0/23** |

Prefix caching is the only variable that matters; concurrency only amplifies it.

jeff3071:
## Summary

Fix #446 

Gemma 4 uses different physical block sizes across KV-cache groups, while request hashes are computed at a shared finer granularity. kvcached previously lost vLLM’s hash_block_size when replacing the native block pool, causing physical blocks to be cached under incorrect prefix hashes.

- Preserves the native vLLM hash_block_size.
- Reuses vLLM’s [BlockHashListWithBlockSize](https://github.com/vllm-project/vllm/blob/main/vllm/v1/core/kv_cache_utils.py#L2237) for hash conversion. 

Testing
- 40 passed in targeted prefix-cache and coordinator tests.
- Tested `google/gemma-4-12B-it` with vLLM 0.24.0.

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -400,6 +400,25 @@ def _set_block_hash(block: Any, key: Any) -> None:
         block.block_hash = key
 
 
+def _convert_block_hashes(
+    block_hashes: Any,
+    hash_block_size: int,
+    target_block_size: int,
+) -> Any:
+    if target_block_size == hash_block_size:
+        return block_hashes
+
+    import importlib
+
+    kv_cache_utils = importlib.import_module("vllm.v1.core.kv_cache_utils")
+    converter = getattr(kv_cache_utils, "BlockHashListWithBlockSize", None)
+    if converter is None:
+        raise RuntimeError(
+            "This vLLM version does not support heterogeneous block-hash conversion"
+        )
+    return converter(block_hashes, hash_block_size, target_block_size)
+
+
 class ElasticBlockPoolPatch(VersionAwarePatch, BasePatch):
     """Inject ElasticBlockPool into vLLM's block pool module"""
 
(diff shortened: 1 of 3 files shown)


RixinLiu:
LGTM! Thanks for finding the root cause of this bug!

PR #452, opened 2026-08-18
fix(kv-cache): size pages via get_block_range in _get_num_alloced_blocks

SuperMarioYL:
# PR Draft — Round 1

**Title:** `fix(kv-cache): size pages via get_block_range in _get_num_alloced_blocks`

**Base:** `main` ← **Branch:** `fix/page-capacity-get-block-range`

---

## Summary

`KVCacheManager._get_num_alloced_blocks` and `available_size` sized every page with
`InternalPage.get_num_blocks(page_size, block_mem_size)` — the theoretical
`page_size // block_mem_size`. But `get_page_occupancy` already uses the
boundary-aware `InternalPage.get_block_range(page_id, page_size, block_mem_size)`,
and the comment there explains why: blocks straddling a page boundary belong to
neither page, so a page's capacity comes from its own `get_block_range`, not from
`page_size // block_mem_size`.

When `block_mem_size` does not evenly divide `page_size` (HYBRID_LINEAR / Mamba GDN
per-block state — the exact case the `_alloc` parking comment at
`kv_cache_manager.py:335` names), some page ids yield **zero** usable blocks while
`get_num_blocks` reports one or more. Those 0-block pages are parked in
`full_pages` (so they're not re-handed-out) but were still counted by
`get_num_blocks`, inflating `_get_num_alloced_blocks`.

That inflation is observable, not cosmetic: `free()`'s lazy-shrink completion gate
`_get_num_alloced_blocks() <= target_num_blocks` reads the inflated count, so an
operator's `kvctl limit` shrink never crosses the threshold and physical GPU memory
is not returned (the elastic-shrink-stalls failure).

**Fix:** add a module-level `_page_capacity(page_id)` helper that mirrors
`get_page_occupancy`'s `get_block_range` usage, and use it in
`_get_num_alloced_blocks` (per tracked page id in `full_pages` / `avail_pages`) so a
parked 0-block page contributes 0 instead of the theoretical count. The helper is
module-level (not a staticmethod) so it is unit-testable without the compiled
`kvcached.vmm_ops` extension or a GPU, matching the `_get_max_cached_blocks` /
`_make_cache_key` idiom.

**Scope:** `available_size`'s `free_pages` term could not be made per-id precise —
the C++ `PageAllocator` exposes only a *count* of free pages, not their ids, so
the boundary-aware `_page_capacity` (which depends on `page_id`) can't be applied
there without a C++ change. The `avail_blocks` term in `available_size` was already
boundary-aware (`num_avail_blocks` is maintained from `page.num_free_blocks()`, which
`InternalPage::init` populates from `get_block_range`). An inline note documents the
remaining upper-bound term and the precise-fix prerequisite.

## Test

`tests/test_block_count_consistency.py` (registered in `tests/manifests/cpu.txt`)
builds a real `KVCacheManager` via `object.__new__` (per `test_resize_reserved_order.py`)
with `page_size=4, block_mem_size=3` (3 does not divide 4) and a stub `InternalPage`
exposing the real static `get_num_blocks` / `get_block_range` formulas. It asserts
`_get_num_alloced_blocks()` returns the `get_block_range`-based count, not the
inflated `get_num_blocks` count. The test is **red on `main`** (concrete value
failures `assert 1 == 0` / `assert 2 == 1`, not import errors, and does not mock the
function under fix) and **green on this branch**.

CPU suite: `142 passed` (was 138; +4 from the new test); the 10 errors are
pre-existing macOS `/dev/shm` environment failures in unrelated files.

## Why not just `available_size`

Addressed above — `available_size`'s count-based free-page term is an upper bound
until the allocator exposes free page ids; the load-bearing accounting
(`_get_num_alloced_blocks`, used by the shrink gate) is fully fixed here.

### DIFF
--- a/kvcached/kv_cache_manager.py
+++ b/kvcached/kv_cache_manager.py
@@ -54,6 +54,30 @@ def synchronized(method):
     return synchronized_method
 
 
+def _page_capacity(page_id: int, page_size: int, block_mem_size: int) -> int:
+    """Return the number of usable blocks on a page.
+
+    Blocks straddling a page boundary belong to neither page (see the
+    comment in ``get_page_occupancy``), so a page's capacity comes from
+    its own ``get_block_range`` rather than from the theoretical
+    ``page_size // block_mem_size`` that ``InternalPage.get_num_blocks``
+    returns. When ``block_mem_size`` does not evenly divide ``page_size``
+    (e.g. HYBRID_LINEAR / Mamba GDN per-block state — the case the
+    ``_alloc`` 0-usable-block parking comment at kv_cache_manager.py:310
+    names), some page ids yield *zero* usable blocks while
+    ``get_num_blocks`` reports one or more; counting those pages with
+    ``get_num_blocks`` inflates both ``available_size`` and the
+    lazy-shrink completion gate ``_get_num_alloced_blocks``.
+
+    Module-level (rather than a staticmethod) so it is unit-testable
+    without the compiled ``kvcached.vmm_ops`` extension or a GPU,
+    matching the ``_get_max_cached_blocks`` / ``_make_cache_key`` idiom.
+    """
+    start, end = InternalPage.get_block_range(page_id, page_size,
+                                               block_mem_size)
+    return end - start
+
+
 class KVCacheManager:
 
     def __init__(
@@ -502,6 +526,16 @@ class KVCacheManager:
             physical_free_pages = self.page_allocator.get_avail_physical_pages(
             ) + self.page_allocator.get_num_reserved_pages()
             free_pages = min(virtual_free_pages, physical_free_pages)
+            # The allocator exposes only a COUNT of free pages, not their ids,
+            # so this term can't use the boundary-aware _page_capacity
+            # (capacity depends on page_id; some ids yield zero usable blocks
+            # when block_mem_size does not divide page_size — see _alloc's
+            # 0-block parking at kv_cache_manager.py:310). get_num_blocks is
+            # the theoretical page_size // block_mem_size, so this is an UPPER
+            # BOUND; the precise accounting in _get_num_alloced_blocks and
+            # get_page_occupancy uses _page_capacity / get_block_range. A
+            # precise fix here needs page-id enumeration from the allocator
+            # (a C++ change, out of scope).
             blocks_from_free_pages = free_pages * InternalPage.get_num_blocks(
                 self.page_size, self.block_mem_size)
         return avail_blocks + blocks_from_free_pages
@@ -628,13 +662,19 @@ class KVCacheManager:
         try_to_reserve() obtains them via alloc(), so they have already left
         their pages. They are deliberately NOT added a second time below.
         """
-        # Blocks from fully allocated pages
-        blocks_from_full_pages = len(self.full_pages) * InternalPage.get_num_blocks(
-            self.page_size, self.block_mem_size)
+        # Blocks from fully allocated pages. Capacity is per-page-id because
+        # blocks straddling a page boundary belong to neither page (see
+        # get_page_occupancy); a parked 0-block page (the _alloc branch at
+        # kv_cache_manager.py:310) contributes nothing here, whereas the
+        # previous len(self.full_pages) * get_num_blocks(...) inflated it.
+        blocks_from_full_pages = sum(
+            _page_capacity(page_id, self.page_size, self.block_mem_size)
+            for page_id in self.full_pages)
         # Blocks from partially allocated pages. num_avail_blocks is the number
         # of free blocks in the partially allocated pages so the number of
         # allocated blocks is the total number of blocks in the partially
         # allocated pages minus the number of free blocks.
-        blocks_from_avail_pages = len(self.avail_pages) * InternalPage.get_num_blocks(
-            self.page_size, self.block_mem_size) - self.num_avail_blocks
+        blocks_from_avail_pages = sum(
+            _page_capacity(page_id, self.page_size, self.block_mem_size)
+            for page_id in self.avail_pages) - self.num_avail_blocks
         return blocks_from_full_pages + blocks_from_avail_pages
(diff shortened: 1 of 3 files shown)


rishabhsinha17 (commented):
Verified this end to end against the C++ the mirror claims to copy. `RichInternalPage.get_block_range` matches InternalPage::get_block_range exactly (page_allocator.cpp:91-98, ceil start, floor end) and `get_num_blocks` the theoretical division (:100-103), and InternalPage::init populates the free list from the same range (:47-48), which is what makes the avail_pages term coherent: num_avail_blocks was already boundary-aware, so the sum of _page_capacity minus num_avail_blocks is exact where the old theoretical minuend was inflated. The stall mechanism is real on main. Parked 0-block pages sit in full_pages forever (free() only visits pages whose indices are being freed, and these have none), each counted as get_num_blocks by the old math, and the lazy-shrink gate at kv_cache_manager.py:512 compares that inflated count. Ran the new file together with the sibling cpu tests that also stub vmm_ops (test_alloc_rollback, test_bestfit_page_selection, test_get_max_cached_blocks) in one pytest invocation on 2cd28a3, 31 passed, so the autouse rebinding does not leak across modules; reverted kvcached/kv_cache_manager.py to main with your tests kept and all four fail there. The get_page_occupancy precedent and the honest upper-bound scope note on available_size both check out.

Two findings, the first worth handling in this PR.

1. Passing the gate now exposes the unchecked resize on the completion path. free() calls page_allocator.resize(target_num_blocks * block_mem_size) and clears in_shrink without looking at the return (kv_cache_manager.py:512-516 on main, unchanged here). The parked 0-block pages this PR stops counting are still in-use pages to the allocator (alloc_page() took them off the free list and nothing returns them), and the C++ resize() refuses whenever the target page count is below in-use (page_allocator.cpp:331-333). So for a shrink aggressive enough that the target sits below parked plus live pages, behavior moves from "gate never passes, shrink stalls visibly" to "gate passes, resize refuses, in_shrink cleared, shrink silently dropped", and available_size stops suppressing the free-page term as well. That path becomes reachable exactly in the scenario this PR fixes, so I would check the return and keep in_shrink and target_num_blocks when the allocator refuses, or at least log the refusal.

2. Follow-up, not this PR. The parked pages themselves are the deeper problem: each is a mapped physical page serving zero blocks, held in full_pages forever and reported as 0 by get_page_occupancy, so no eviction path will ever release it. Counting them correctly is right; the cure is for the allocator to skip page ids whose block range is empty (or for _alloc to hand them back), which is a C++-side change. Worth filing so it does not get lost behind the accounting fix.

Small one. The hardcoded kv_cache_manager.py:335 in the _page_capacity docstring, the available_size comment, and the test module docstring is already :387 on current main after #448/#414/#461 landed (it was right when you branched). Naming the branch, say "the 0-usable-block parking in _alloc", keeps it from rotting again.

SuperMarioYL:
Thanks for the careful trace against the C++ mirror and for catching the path this fix exposes.

Finding 1 addressed in ae0b881 (worth handling here, as you noted):

The `free()` in_shrink completion path now checks `page_allocator.resize()`'s return. On refusal it keeps `in_shrink` and `target_num_blocks` set and logs a warning, so the shrink stays pending and a later `free()` retries instead of being silently dropped. That mirrors the standalone `resize()`'s lazy-shrink-wait semantics (which also leaves `in_shrink`/`target_num_blocks` set on a refused resize). The previously-unreachable "gate passes, resize refuses, in_shrink cleared" path is now the same visible stall it was before the parked-page count was fixed.

Added `test_free_keeps_shrink_pending_when_allocator_refuses_resize`: a `_ShrinkRefusingPageAllocator` whose `resize()` returns False drives `free()` through the gate; red without the fix (`in_shrink` is cleared) and green with it (shrink stays pending, `target_num_blocks` preserved). Ran it together with the sibling vmm_ops-stubbing CPU tests — 32 passed, ruff/mypy clean.

Finding 2 (parked 0-block pages held in `full_pages` forever, each a mapped physical page serving zero blocks) is the deeper problem and, as you scoped it, a follow-up rather than this PR. Happy to open a separate issue/PR for it if that's useful.

rishabhsinha17 (approved):
Verified ae0b881 and d6c8bd5 against the branch plus a local red and green run.

free() now checks the resize return (kv_cache_manager.py:462-480), keeps in_shrink and target_num_blocks on refusal, and warns. Standalone resize() leaves the same pending state when page_allocator.resize() refuses (505-526). Both entry points now agree on the lazy-shrink wait, and available_size() keeps suppressing the free-page term until the retry succeeds (539-540), which closes the second half of the exposure.

Ran the four vmm_ops-stubbing CPU files together on d6c8bd5, 32 passed. Reverting only the free() hunk with your tests kept fails the new test and nothing else.

d6c8bd5 fixes a concrete isolation hole rather than a race. test_memory_limit_control.py calls _load_manager_class() at module level during collection, which pops kvcached.kv_cache_manager, reimports it under a vmm_ops stub whose InternalPage is a SimpleNamespace with only get_num_blocks, restores vmm_ops, and leaves the stub-bound kv_cache_manager in sys.modules (tests/test_memory_limit_control.py:30-42). Your fixture patches the module object it cached at collection while the lazy import inside the test resolves the newer one. CI's full manifest hits that AttributeError on get_block_range. A four-file run never does. Explicit injection removes the module-global lookup, the right cure here. That leftover stale module remains a manifest-wide hazard for any test collected later, pre-existing and separate from this PR.

On finding 2 a separate issue is the right container. The accounting fix stands on its own and should not wait for a C++-side change to how _alloc parks 0-usable-block pages. When you file it, link this thread for the trace, pages held in full_pages forever, reported 0 by get_page_occupancy, invisible to every eviction path. Happy to file it myself if you would rather keep moving.

rishabhsinha17:
Verified end to end against the branch and a reverted control on Sep 10. Head unchanged since and all 11 checks green, marking ready for your pass. Happy to use a label for these ready-for-maintainer handoffs if you prefer one, say the word and I will apply it going forward.

RixinLiu (approved):

PR #456, opened 2026-08-19
perf(kv-cache): cache get_avail_physical_pages in available_size to skip per-alloc cudaMemGetInfo

SuperMarioYL:
## Summary

`KVCacheManager.available_size()` calls `PageAllocator.get_avail_physical_pages()` on every invocation, and that C++ method issues a `cudaMemGetInfo` driver call. `available_size()` runs once per allocation request (`patches.py:792`) and once per scheduler step (`:927`), so the driver call fires on every alloc/step. This PR caches the `get_avail_physical_pages()` result for 100 ms — the `resize_watcher` poll interval — and invalidates it after every physical-pool mutation the manager drives, so the driver call fires at most once per 100 ms window instead of once per call.

## Motivation

`benchmarks/bench_alloc/README.md` documents a 12.5× allocation-throughput win for dropping `cudaMemGetInfo` from `available_size()`. That figure is the README's historical benchmark of the original implementation — it is not a measurement made in this PR, and this PR does not claim it. At HEAD (`ac9680a`) the drop did not happen: the call was relocated from `available_size()` into `get_avail_physical_pages()` (`csrc/page_allocator.cpp:481`), which `available_size()` still calls (`kvcached/kv_cache_manager.py:502`). What this PR restores is the mechanism behind that number — removing the per-call driver read — by caching the call rather than relocating it. The claims this PR makes and tests are: (1) the `cudaMemGetInfo` call count collapses from one per `available_size()` to one per 100 ms window, and (2) the staleness bounds described below. Refs #299 (the "Reduce CUDA Call Overhead in `available_size`" track item); this PR takes the narrower TTL-cache approach rather than the issue's proposed C++ rewrite.

## Changes

- `kvcached/kv_cache_manager.py`: add a TTL cache (value + `time.monotonic()` timestamp, 100 ms) for `get_avail_physical_pages()` in `available_size()`. Invalidated after every physical-pool mutation the manager drives — page mapping in `alloc()`, `free()`, `resize()`, `trim()`, and `clear()` — plus the `in_shrink` completion toggle, so a manager-driven mutation is never served stale physical-free data. `get_num_free_pages()` and `get_num_reserved_pages()` stay uncached (cheap / atomic).
- `tests/test_available_size_cache.py` (new, CPU-only): five tests — the call-count collapse across repeated `available_size()`, refetch after the TTL window expires, and refetch after `resize()` / `alloc()` / `trim()` respectively.

## Reviewer note — staleness bound

The physical free-page count can go stale for up to 100 ms between refreshes. Two classes of staleness are distinguished:

- **Manager-driven mutations** (`alloc`, `free`, `resize`, `trim`, `clear`, and the `in_shrink` completion toggle) invalidate the cache immediately, so no stale value is served across them.
- **Sources the manager cannot invalidate** — the C++ prealloc thread mapping reserved pages in the background, and a sibling pool in a multi-manager process — are bounded by the TTL window and absorbed by the `alloc_page` miss path, the same way cross-process races already are.

The 100 ms figure is the allocator's existing `resize_watcher` cadence (one driver read per watcher window), not a claim that serving generally tolerates 100 ms of capacity staleness; see the `_get_cached_avail_physical_pages` docstring for the full accounting.

## Tests

`python -m pytest tests/test_available_size_cache.py` — 5 passed. The cache test is red on `main` (no cache → call count N) and green on this branch. Existing CPU tests touching `available_size` (`test_prefix_cache`, `test_bestfit_page_selection`, `test_page_aware_eviction`, `test_alloc_rollback`, `test_observability`) still pass; the full CPU manifest is 273 passed on this branch vs 268 on `main` (delta = the 5 new tests). The new test file is registered in `tests/manifests/cpu.txt`; `python tools/check_test_classification.py` passes.

### DIFF
--- a/kvcached/kv_cache_manager.py
+++ b/kvcached/kv_cache_manager.py
@@ -709,10 +709,10 @@ class KVCacheManager:
                 and now - self._avail_physical_pages_ts
                 < _AVAIL_PHYSICAL_PAGES_TTL_S):
             return cached
-        cached = self.page_allocator.get_avail_physical_pages()
-        self._avail_physical_pages_cache = cached
+        result: int = self.page_allocator.get_avail_physical_pages()
+        self._avail_physical_pages_cache = result
         self._avail_physical_pages_ts = now
-        return cached
+        return result
 
     @synchronized
     def get_page_occupancy(self, page_ids: List[int]) -> Dict[int, int]:
--- a/kvcached/kv_cache_manager.py
+++ b/kvcached/kv_cache_manager.py
@@ -705,12 +705,13 @@ class KVCacheManager:
         """
         now = time.monotonic()
         cached = self._avail_physical_pages_cache
-        if (cached is None
-                or now - self._avail_physical_pages_ts
-                >= _AVAIL_PHYSICAL_PAGES_TTL_S):
-            cached = self.page_allocator.get_avail_physical_pages()
-            self._avail_physical_pages_cache = cached
-            self._avail_physical_pages_ts = now
+        if (cached is not None
+                and now - self._avail_physical_pages_ts
+                < _AVAIL_PHYSICAL_PAGES_TTL_S):
+            return cached
+        cached = self.page_allocator.get_avail_physical_pages()
+        self._avail_physical_pages_cache = cached
+        self._avail_physical_pages_ts = now
         return cached
 
     @synchronized
--- a/kvcached/kv_cache_manager.py
+++ b/kvcached/kv_cache_manager.py
@@ -40,6 +40,12 @@ logger = get_kvcached_logger()
 
 KV_TENSOR_WAIT_TIMEOUT: float = 10.0  # seconds
 
+# TTL for the cached get_avail_physical_pages() result in available_size().
+# Matches the C++ resize_watcher poll interval (csrc/page_allocator.cpp:838)
+# so physical-free-page staleness is bounded by the same 100 ms window the
+# allocator already tolerates for virtual-free-page polling.
+_AVAIL_PHYSICAL_PAGES_TTL_S: float = 0.1
+
 
 def synchronized(method):
     """
@@ -55,6 +61,12 @@ def synchronized(method):
 
 
 class KVCacheManager:
+    # Cached get_avail_physical_pages() result + its monotonic timestamp.
+    # Class-level defaults keep the no-__init__ test-stub pattern
+    # (tests/test_alloc_rollback.py) working without each stub knowing
+    # about the cache; available_size() re-fetches when the TTL elapses.
+    _avail_physical_pages_cache: Optional[int] = None
+    _avail_physical_pages_ts: float = 0.0
 
     def __init__(
         self,
@@ -189,6 +201,11 @@ class KVCacheManager:
         self._memory_limit_bytes: Optional[int] = None
         self._memory_limit_effective_bytes: Optional[int] = None
         self._memory_limit_revision = -1
+        # TTL cache for get_avail_physical_pages() (a cudaMemGetInfo driver
+        # call); see _AVAIL_PHYSICAL_PAGES_TTL_S. Invalidated on resize() and
+        # when in_shrink toggles so a resize/shrink is never served stale.
+        self._avail_physical_pages_cache: Optional[int] = None
+        self._avail_physical_pages_ts: float = 0.0
         # NOTE: we use a no-op lock for sync scheduling to avoid overhead
         self._lock = threading.RLock() if async_sched else NoOpLock()
 
@@ -513,6 +530,10 @@ class KVCacheManager:
                 self.page_allocator.resize(self.target_num_blocks *
                                            self.block_mem_size)
                 self.in_shrink = False
+                # Exiting shrink: the resize above changed the physical
+                # footprint and this toggle bypasses resize(), so drop the
+                # cached value so available_size() re-reads the driver.
+                self._avail_physical_pages_cache = None
                 self.target_num_blocks = None
 
     @synchronized
@@ -544,6 +565,10 @@ class KVCacheManager:
         new_mem_size: the memory size of the K or V tensor in one layer
         """
         self._wait_post_init()
+        # resize() changes the physical footprint (and may toggle in_shrink);
+        # drop the cached avail-physical-pages so the next available_size()
+        # re-reads the driver instead of serving pre-resize data.
+        self._avail_physical_pages_cache = None
         assert new_mem_size >= 0, "new_mem_size must be non-negative"
         if self.page_allocator.resize(new_mem_size):
             if self.in_shrink:
@@ -657,12 +682,36 @@ class KVCacheManager:
             blocks_from_free_pages = 0
         else:
             virtual_free_pages = self.page_allocator.get_num_free_pages()
-            physical_free_pages = self.page_allocator.get_avail_physical_pages(
-            ) + self.page_allocator.get_num_reserved_pages()
+            physical_free_pages = (
+                self._get_cached_avail_physical_pages()
+                + self.page_allocator.get_num_reserved_pages())
             free_pages = min(virtual_free_pages, physical_free_pages)
             blocks_from_free_pages = free_pages * InternalPage.get_num_blocks(
                 self.page_size, self.block_mem_size)
         return avail_blocks + blocks_from_free_pages
+
+    def _get_cached_avail_physical_pages(self) -> int:
+        """Return get_avail_physical_pages(), TTL-cached for available_size().
+
+        The underlying call fires cudaMemGetInfo (csrc/page_allocator.cpp:481)
+        on every available_size(), which runs per alloc
+        (kvcached/integration/vllm/patches.py:792) and per scheduler step
+        (:927); the TTL window collapses those to one driver read. The cache
+        is invalidated on resize() and when in_shrink toggles, so a
+        resize/shrink is never served stale physical-free data.
+        get_num_free_pages() and get_num_reserved_pages() stay uncached (cheap
+        / atomic). Called under available_size()'s @synchronized lock, so the
+        cache read/write here is already serialized.
+        """
+        now = time.monotonic()
+        cached = self._avail_physical_pages_cache
+        if (cached is None
+                or now - self._avail_physical_pages_ts
+                >= _AVAIL_PHYSICAL_PAGES_TTL_S):
+            cached = self.page_allocator.get_avail_physical_pages()
+            self._avail_physical_pages_cache = cached
+            self._avail_physical_pages_ts = now
+        return cached
 
     @synchronized
     def get_page_occupancy(self, page_ids: List[int]) -> Dict[int, int]:
@@ -769,6 +818,9 @@ class KVCacheManager:
 
         self.target_num_blocks = None
         self.in_shrink = False
+        # clear() freed every page and trimmed; drop the cached value so
+        # the next available_size() re-reads the post-clear physical state.
+        self._avail_physical_pages_cache = None
         self.num_avail_blocks = 0
 
         # Possibly reserve the first block as null block for padding tokens
--- a/kvcached/kv_cache_manager.py
+++ b/kvcached/kv_cache_manager.py
@@ -395,6 +395,10 @@ class KVCacheManager:
                 try:
                     page = self.page_allocator.alloc_page()
                     page.init(self.block_mem_size)
+                    # alloc_page() mapped a new physical page, shrinking the
+                    # driver's free pool; drop the cached count so the next
+                    # available_size() re-reads instead of serving stale data.
+                    self._avail_physical_pages_cache = None
                 except RuntimeError as e:
                     self._rollback_partial_alloc(ret_index, num_from_reserved)
                     logger.warning(
@@ -523,6 +527,9 @@ class KVCacheManager:
 
         if pages_to_free:
             self.page_allocator.free_pages(pages_to_free)
+            # free_pages() returned physical pages to the driver, growing the
+            # free pool; drop the cached count so available_size() re-reads.
+            self._avail_physical_pages_cache = None
 
         if self.in_shrink:
             assert self.target_num_blocks is not None
@@ -804,6 +811,9 @@ class KVCacheManager:
             pages_to_free.append(page.page_id)
         if pages_to_free:
             self.page_allocator.free_pages(pages_to_free)
+            # free_pages() returned physical pages to the driver, growing the
+            # free pool; drop the cached count so available_size() re-reads.
+            self._avail_physical_pages_cache = None
         self.avail_pages.clear()
         self.full_pages.clear()
 
--- a/tests/manifests/cpu.txt
+++ b/tests/manifests/cpu.txt
@@ -1,4 +1,5 @@
 tests/test_alloc_rollback.py
+tests/test_available_size_cache.py
 tests/test_bestfit_page_selection.py
 tests/test_get_max_cached_blocks.py
 tests/test_ipc_name.py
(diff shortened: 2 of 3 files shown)


jeff3071:
Thanks for your contribution!

I think this change introduces a correctness issue. `physical_free_pages` now uses a cached value:
```python
physical_free_pages = (
                self._get_cached_avail_physical_pages()
                + self.page_allocator.get_num_reserved_pages())
```

But the cached pages do not change after alloc that cause  cached pages number is dirty.
That would cause `available_size` return wrong number.

However, the cache is not updated or invalidated after alloc() maps new physical pages. As a result, subsequent calls to available_size() within the TTL may use a stale physical-page count and overestimate the available KV-cache capacity.

This can be reproduced by changing `n_blocks` to 1024 in tests/test_kvcache_manager.py::test_basic_alloc_free:
```python
def test_basic_alloc_free(setup_kvcache):
    # instantiate a kv cache manager with known size
    manager = setup_kvcache

    # initial available blocks
    initial_available = manager.available_size()

    # allocate some blocks
    # change this
    n_blocks = 1024 
    handle = manager.alloc(n_blocks)
    after_alloc = manager.available_size()
    assert after_alloc + n_blocks == initial_available

    # free the allocated blocks
    manager.free(handle)
    after_free = manager.available_size()
    assert after_free == initial_available
```

Result
```
>       assert after_alloc + n_blocks == initial_available
E       assert (20992 + 1024) == 21632

tests/test_kvcache_manager.py:99: AssertionError
```

SuperMarioYL:
Thanks for catching this — you are right. The cache served stale physical-free data after `alloc()` mapped new pages (and after `free()`/`clear()` returned pages), so `available_size()` could overestimate KV-cache capacity within the TTL window. Fixed in 2395135 by dropping the cached value at each physical-pool mutation, consistent with the existing `resize()`/`in_shrink` invalidation:

- `alloc_page()` maps a new physical page → invalidate (in `alloc`)
- `free_pages()` returns pages to the driver → invalidate (in `free` and `clear`)

The per-scheduler-step hot path (`available_size()` with no pool mutation between calls) still collapses to one driver read per TTL window; only the page-map/return paths now force a re-read, which is the correct place to re-read since capacity actually changed there.

Since `test_basic_alloc_free` needs a GPU, I added a CPU-only regression test (`tests/test_available_size_cache.py::test_available_size_refetches_after_alloc`) that stubs the page allocator: `alloc` maps a page and `available_size()` must drop by one page's worth of blocks instead of replaying the pre-alloc cached value. It is red on the previous commit and green on this one.

SuperMarioYL:
Rebased onto latest main to resolve a merge conflict in `kv_cache_manager.py`. Two upstream commits (#448, #414) added memory-limit tracking and refactored null-block reservation in the same `__init__` region where this PR adds the avail-physical-pages TTL cache. Both sets of instance variables are kept; the cache invalidation in `resize()` now adopts upstream's `>= 0` assertion. All 12 tests pass and mypy is clean.

rishabhsinha17 (commented):
Verified end to end, including both directions of the tests. The call chain the motivation rests on is real at HEAD: available_size() reaches get_avail_physical_pages() (the cached read is at kv_cache_manager.py:693 on this branch), which fires gpu_vmm::mem_get_info, i.e. cudaMemGetInfo, on every call (csrc/page_allocator.cpp:479-481, gpu_vmm.hpp:179), while benchmarks/bench_alloc/README.md:23 says the C++ path skips it, so the README's 12.5x is currently aspirational and caching is a legitimate way to land it. The invalidation set now covers every physical-pool mutation this manager performs: alloc_page success including the parked 0-block branch (:401 runs before the parking check), free()'s free_pages call and the in_shrink completion toggle (:532, :543), resize() entry, which also covers set_memory_limit since that path goes through resize() (:578), and both clear() sites (:816, :834). The class-level defaults keep the object.__new__ stub pattern working, monotonic time is the right clock, and the cache is only touched under available_size()'s lock. Ran tests/test_available_size_cache.py together with the sibling vmm_ops-stubbing cpu tests in one invocation on faa9c7d, 19 passed; with kvcached/kv_cache_manager.py reverted to main and the tests kept, the TTL test and the alloc-staleness regression from jeff3071's catch fail as intended, and the resize test passes both ways as a spec test. cpu.txt is updated so CI runs the file, and CI is green.

Three small things, none blocking.

1. trim() is the one physical-pool mutation without an invalidation. It unmaps every reserved page, so right after it the live reserved term drops to zero while the cached avail stays low-stale, and available_size() underestimates by the trimmed amount for up to one TTL window. Safe direction, but free() got the invalidation for the same grow-direction mutation, so trim() should too for consistency, and engines call trim() on idle transitions where a capacity dip immediately after is at least confusing.
2. The invalidation story in the docs lags the fix. The __init__ comment and the _get_cached_avail_physical_pages docstring still say "invalidated on resize() and when in_shrink toggles", wording from before 2395135 added the alloc/free/clear sites. Worth refreshing so a reader does not conclude alloc can serve stale data.
3. The staleness note covers other processes on the GPU but not the two same-process sources: the C++ prealloc worker maps reserved pages in the background, so stale avail plus live reserved double counts the just-mapped pages for up to one window, and a sibling pool in a multi-manager process invalidates only its own cache. Both are the same bounded overestimate, absorbed by the alloc_page miss path the way cross-process races already are, so no code change needed; naming them in the docstring keeps the bound honest.

The steady-state hit rate survives the per-mutation invalidation because most allocs are served from already-mapped pages and only a page-mapping alloc drops the cache, so the per-step and per-alloc callers in patches.py still coalesce to roughly one driver read per window.

SuperMarioYL:
Thanks for the deep end-to-end verification, @rishabhsinha17 — really appreciate you confirming the call chain at HEAD and running the file against both the branch and a reverted `kv_cache_manager.py`.

All three nits addressed in f1f5e89:

1. **trim() invalidation** — added `self._avail_physical_pages_cache = None` after `page_allocator.trim()`, so `available_size()` re-reads instead of underestimating by the trimmed amount for one TTL window. Same grow-direction mutation `free()` already invalidates, so `trim()` now does too. `clear()` calls `trim()` and still re-invalidates afterward (harmless redundancy kept).
2. **Invalidation docs** — refreshed the `__init__` comment and the `_get_cached_avail_physical_pages` docstring to list every site (alloc mapping a page, `free()`, `resize()`, `trim()`, `clear()` + the in_shrink completion toggle) instead of the stale "resize() and in_shrink toggles" wording, so a reader won't conclude alloc can serve stale data.
3. **Same-process staleness sources** — named both in the docstring: the C++ prealloc thread mapping reserved pages in the background (stale avail + live reserved double-counts them for up to one window) and a sibling pool in a multi-manager process whose own invalidation doesn't reach this cache. Both are the same bounded overestimate, absorbed by the `alloc_page` miss path the way cross-process races already are — no code change, just making the bound honest as you noted.

Added `test_available_size_refetches_after_trim` (a `TrimFreeingPageAllocator` whose `trim()` grows the driver free pool); it's red without the trim invalidation (serves the pre-trim cached 400 instead of 600) and green with it. Ran it together with the sibling vmm_ops-stubbing CPU tests in one invocation — 32 passed, ruff/mypy clean.

rishabhsinha17 (approved):
Verified the trim invalidation, the doc refresh, and the staleness naming against f1f5e89. The trim() invalidation is right after page_allocator.trim() (kv_cache_manager.py:604-609). That makes seven invalidation sites on this branch (403, 534, 545, 580, 609, 831, 849), and the refreshed __init__ comment and docstring name that exact set plus the two same-process staleness sources. Ran the four-file set on f1f5e89, 32 passed. With only the trim invalidation removed, test_available_size_refetches_after_trim fails alone, and TrimFreeingPageAllocator models the right direction, unmap grows the driver pool and the stale cached value underestimates. clear() re-invalidating after its internal trim() is redundant now but harmless. Approving.

rishabhsinha17:
Verified end to end against the branch and a reverted control on Sep 10. Head unchanged since and all 11 checks green, marking ready for your pass.

SuperMarioYL:
Rebased onto latest main to clear a conflict with #452 (the `available_size` comment block landed there). While resolving I also re-bound the cache invalidation onto the newer main-side structures that now sit in the same paths:

- `free()` in-shrink completion: invalidation now lives in the `if resized:` branch of the `QuarantinedResizeError`-aware resize; the quarantine-reject and refuse branches exit without a completed resize, so no stale-data window is introduced by leaving them uncached.
- `alloc_page()` page-mapping path: the cache drop stays inside the try block alongside the new `StateConsistencyError` re-raise.

CPU manifest is unchanged relative to main (the one `test_tp_listener_device` failure and the ten `/dev/shm` errors reproduce on main here; our 5 cache tests pass).

shipiyouniao (commented):
Rechecked `03ba7ff`, including the #418/#452 rebase. I did not find a new correctness blocker in the cache invalidation paths.

Validation on my side:
- 44 focused CPU tests passed, including two extra checks for stale capacity and fatal-pool propagation.
- 30 native VMM transaction/failure-policy tests passed on a T4.
- A separate CUDA process allocated/freed 128 MiB. With a controlled TTL clock, the cached count stayed stale inside the window and refreshed in both directions at expiry.
- Ten injected native physical-allocation failures returned allocation misses; all ten retries succeeded with CUDA readback.

The native extension was reused only after verifying that its C++ and build sources match this head. These are single-GPU contract/fault checks with background preallocation disabled, not a colocated-serving performance benchmark.

Could you add a small before/after serving comparison with two colocated instances, including throughput, TTFT and allocation-miss counts? The PR description still says this delivers the README's 12.5x win, but that historical number is not a measurement of this implementation. It also still lists only the original three tests/invalidation sites. Updating those claims and showing the effect of the 100 ms window under real serving pressure would help finish the performance review; the resize watcher's identical polling interval alone does not establish an equivalent capacity-staleness tolerance.

SuperMarioYL:
@shipiyouniao thanks for the thorough recheck — you're right on both description points, and I've updated the PR body:

- The 12.5× figure is now attributed explicitly as the README's historical benchmark of the original implementation, not a measurement made in this PR. The claims this PR makes and tests are the call-count collapse (one `cudaMemGetInfo` per 100 ms window) and the staleness bounds.
- The changes/tests sections now reflect the post-rebase state: invalidation after every manager-driven mutation (`alloc` / `free` / `resize` / `trim` / `clear` + the `in_shrink` completion toggle), and the five CPU tests (call-count, TTL expiry, refetch after `resize` / `alloc` / `trim`).

On the staleness-tolerance point, agreed that the identical polling interval by itself doesn't establish tolerance. The load-bearing argument is narrower: every mutation this engine drives invalidates immediately, so staleness only arises from sources the manager cannot see — the C++ prealloc thread mapping reserved pages, and a sibling pool in a multi-manager process. Those are bounded by the TTL window and absorbed by the `alloc_page` miss path the same way cross-process races already are (also spelled out in the `_get_cached_avail_physical_pages` docstring).

On the colocated before/after serving comparison: happy to — with one constraint, that the machines I develop on are CPU-only, so I can't produce the two-colocated-instance numbers myself. Two ways forward:

1. I add a self-contained benchmark script (two colocated instances; throughput, TTFT, allocation-miss counters) sized for a T4-class rig like the one you used, and you run it for the numbers.
2. If the repo has a preferred colocated-serving harness or config, point me at it and I'll match it in the script.

If neither works, tell me what shape of evidence would be most useful and I'll aim for that.

shipiyouniao:
Thanks for updating the description and clarifying the staleness argument. We can take the GPU comparison on our T4 and reuse our existing serving setup, so no need to add a separate benchmark script just for this.

I'll compare the same source revision with and without the TTL-cache change, using two colocated instances under light load and near the physical-capacity limit. We'll keep the model/runtime/settings fixed and record throughput, TTFT, allocation misses, and completion/recovery checks. That should tell us whether fewer driver reads translate into a serving benefit, and whether stale capacity causes extra retries. I'll post the exact setup and results once the comparison is complete.

shipiyouniao (approved):
Finished the colocated T4 comparison on `03ba7ff`. The cache does remove the repeated capacity queries, with effectively unchanged serving performance in this setup. This addresses the serving comparison I asked for.

I used A/B/B/A, with two independent Qwen2.5-0.5B-Instruct FP16 instances on one T4: vLLM 0.28.0 V1, PyTorch 2.13.0+cu129, TRITON_ATTN, synchronous scheduling, CUDA graphs, 2 MiB pages, prefix caching capped at 4096 tokens, and background preallocation disabled. Both arms used the same native extension and source tree; only `kv_cache_manager.py` was restored to parent `73fa57ce` for the control.

Both arms used native `kv_cache_memory_bytes=10 GiB` to avoid the separate vLLM 0.28 startup-profiling contract on this branch; physical KV allocation remained dynamic. This does not validate the default startup-profiling path.

Native `vllm bench serve`, per instance, temperature 0, fixed seeds, three warmups:
- Light: I256/O128, N32/C4, 4 req/s.
- Capacity pressure: I4096/O128, N64/C32, unlimited request rate. A separate CUDA process held a fixed allocation, leaving about 3.1 GiB physically free before the workload; KVCached utilization remained 0.9.

Throughput below is the sum of both instance rates, averaged over two runs per version. Queries/preemptions are totals over those runs.

| Metric | Light: control / PR | Capacity pressure: control / PR |
| --- | ---: | ---: |
| Output throughput (tok/s) | 546.07 / 545.21 (-0.16%) | 81.14 / 81.07 (-0.08%) |
| Mean TTFT (ms) | 48.19 / 48.83 | 47500.06 / 47702.63 |
| Mean TPOT (ms) | 13.69 / 13.71 | 367.03 / 366.04 |
| Physical-capacity queries | 26693 / 629 | 47630 / 1382 |
| Manager allocation misses | 0 / 0 | 0 / 0 |
| vLLM preemptions | 0 / 0 | 2 / 2 |

All 768 benchmark requests completed with all 98,304 requested output tokens. Startup and recovery probes returned `OK`, final running/waiting counts were zero, and there were no runtime CUDA or EngineCore failures. One API process logged `EngineDeadError` after the harness sent SIGTERM and vLLM forcibly tore down its child; that shutdown diagnostic is retained separately, not counted as a serving failure.

The query counters cover the client invocation, including its probe/warmups; the throughput and latency values come from the client's measured interval. The hooks count Python calls to the native capacity query, not all driver calls made internally by C++.

So the supported claim here is roughly 97% fewer capacity queries without an observed serving regression, not a 12.5x serving speedup. No native allocation failure was triggered by this workload; it complements, rather than repeats, the fault/cache-expiry checks in my previous review. TP>1, MPS and background preallocation remain outside these measurements.

One small wording follow-up: the comment above `_AVAIL_PHYSICAL_PAGES_TTL_S` still says the allocator already tolerates this window. It would be good to mirror the more careful wording now in the PR body there, and qualify one-read-per-window as applying between invalidations.

@RixinLiu the requested T4 comparison is complete; ready for your pass. I haven't merged it.

SuperMarioYL:
Thanks for running the colocated comparison — that A/B/B/A setup with the capacity-pressure arm is a much better validation than anything we could produce from CPU-only machines, and the 97%-fewer-queries-without-serving-regression readout is the right characterization.

The wording follow-up is done in b6f756e: the comment above `_AVAIL_PHYSICAL_PAGES_TTL_S` no longer claims the allocator "already tolerates" the window. It now mirrors the staleness accounting in the PR description — one driver read per 100 ms window applies *between invalidations* only; manager-driven mutations (alloc/free/resize/trim/clear and the in_shrink completion toggle) invalidate immediately, and the TTL bounds just the two sources the manager cannot invalidate (the C++ prealloc thread and a sibling-manager pool), which the `alloc_page` miss path absorbs. The focused cache tests still pass (5/5).

RixinLiu (approved):
Tested on A100 using Qwen2.5-7B at almost full load, this cache mechanism does incur visible throughput overhead, and the cudaMemGetInfo indeed get called much fewer times. 
I think the negligible overhead is because the refresh mechanism inside this PR helps, but it also means the overhead introduced by cudaMemGetInfo is not the bottleneck.
This PR can be merged, but i will keep this in mind when i found performance issue in the future.

PR #457, opened 2026-08-20
feat: add kvweb rest api

allen3325:
## Summary

`kvctl` and `kvtop` only work in terminal. this PR exposes the same `/dev/shm` KV cache stats over HTTP, plus the limit changes `kvctl limit` performs, so a
segment can be inspected or resized from HTTP. Start it with `kvweb` or `kvctl web`; FastAPI generates the docs at `/docs`.

`GET /api/status`, `/api/ipcs`, `/api/ipcs/{name}`, `POST .../limit`, `POST .../limit-percent`, `DELETE .../{name}`, and `GET /api/stream` (SSE).
Details in `docs/WEB_API.md`.

## Notes for review

- **Kept things secure by default.** The endpoints can shrink or delete a running engine's KV cache, so the server binds to `127.0.0.1`, sends no CORS headers, and takes an optional API key via `KVCACHED_WEB_API_KEY`. Binding elsewhere without a key prints a warning.
- **Prevented ghost segments.** `RwLockedShm` silently creates missing files when updating limits. We added an upfront existence check that throws a 404 so a typo doesn't spawn random dummy segments in kvtop. (`kvctl limit` has the same behaviour today; not touched here.)
- `fastapi`/`uvicorn` are an optional `web` extra, and `kvctl` imports `kvweb` lazily, so nothing changes for users who do not install it.
- `_parse_size()` in `kvctl.py` is renamed `parse_size()` and reused, rather than copied into the new module.
- Fixes a `kvctl` shell bug this touched: `web` was in `COMMANDS` and the help text but had no dispatch branch, so it fell through to the `os.system()` fallback.

I'll follow up with a web dashboard PR on top of this API.

## Testing

`tests/test_kvweb_api.py` covers the endpoints, the 404 paths, and the API key.
Full CPU suite passes (176), as does `pre-commit run --all-files`.
Adds [fastapi, uvicorn, pydantic, httpx] to the CPU CI deps so the new tests actually run


RixinLiu:
Hi, thanks for the contribution, however, i dont think we need this in kvcached, it could be something in the production layer like discussed in #375

shipiyouniao:
Thanks for putting this together. Following Rixin's scope guidance above, would you be happy to close this PR here and keep the HTTP/dashboard layer in a separate integration?

We'd welcome your help with the public observability interfaces in #375. A small example consuming the public snapshots and capabilities, or consumer-level contract tests, would be useful next steps. The capability, lifecycle and snapshot-history work already has active PRs (#474, #478 and #485), so we can coordinate there rather than duplicate them.

I've refreshed the tracking comment in #375 with the current status and remaining work: https://github.com/ovg-project/kvcached/issues/375#issuecomment-5078876572. Feel free to suggest a scope that interests you.

shipiyouniao:
Closing this based on Rixin's scope decision above. The HTTP API, authentication, and dashboard serving belong in a separate production integration rather than in kvcached itself. That integration can consume the public snapshot and memory-limit APIs.

Thanks for the work here. We'd still welcome your help with #375, especially a small consumer example or contract tests for the public observability interfaces.

PR #461, opened 2026-08-21
fix(vllm): preserve inference mode during memory profiling

### LINKED ISSUE #460: fix: #448 drops inference mode during vLLM memory profiling
## Problem

#448 replaces vLLM's `GPUWorker.determine_available_memory()` but does not
preserve the original method's `@torch.inference_mode()` decorator.

In vLLM 0.22.1, `GPUModelRunner.profile_run()` is not independently decorated
with `@torch.inference_mode()`; it relies on the outer worker method to establish
inference mode. As a result, the patched path runs the multimodal dummy profile
with autograd enabled, retains the vision encoder's backward graph and
intermediate activations, and can turn startup into a real CUDA OOM.

This is a regression introduced by #448 rather than an unavoidable multimodal
profiling peak.

## Evidence

An A/B test used the same image, model, startup arguments and profiling inputs.
Only `kvcached/integration/vllm/patches.py` changed.

Both paths received the same inputs:

- V1 model runner
- `pixel_values=(65536, 1536)`, `bfloat16`
- `image_grid_thw=(1, 3)`
- identical multimodal budgets and item counts

Memory observed around the same `model.embed_multimodal()` call:

| Path | Before embed | After embed | PyTorch peak | Process footprint |
| --- | ---: | ---: | ---: | ---: |
| Before #448 | 0.867 GiB | 0.944 GiB | 1.943 GiB | about 2.9 GiB |
| With #448 | 0.867 GiB | 13.13 GiB | 13.62 GiB | about 14.06 GiB |

The increase occurs inside the vision encoder forward, before
`determine_available_memory()` returns and before KV cache configuration,
`num_blocks`, or `_allocate_kv_cache_tensors()` runs. This rules out eager KV
allocation, Mamba cache allocation, changed dummy inputs, and image resizing as
the cause.

The relevant execution contract is:

```python
# vLLM 0.22.1
@torch.inference_mode()
def determine_available_memory(self) -> int:
    ...
    self.model_runner.profile_run()
```

The replacement introduced by #448 calls the same profiling methods without
that decorator. The expected mode difference is therefore:

- original path: `torch.is_grad_enabled() == False` and
  `torch.is_inference_mode_enabled() == True`
- patched path: `torch.is_grad_enabled() == True` and
  `torch.is_inference_mode_enabled() == False`

## Proposed fix

Preserve vLLM's original execution contract by decorating the replacement with
`@torch.inference_mode()` (or entering the equivalent context) for the complete
profiling path.

This must not skip the model forward, multimodal profiling, CUDA graph
profiling/capture, or real CUDA error propagation. It only prevents autograd
state from being retained during an inference-only startup profile.

Regression tests should assert that both `profile_run()` and
`profile_cudagraph_memory()` execute with inference mode enabled when called
through the patched `determine_available_memory()`.

## Expected behavior

The patched worker should retain #448's logical KV-capacity behavior while
matching vLLM's original inference-mode semantics and avoiding the additional
vision-encoder activation peak.

shipiyouniao:
## Summary

- Preserve vLLM's original inference-mode contract when replacing `GPUWorker.determine_available_memory()`.
- Verify that both model profiling and CUDA Graph profiling run with autograd disabled and inference mode enabled.

## Root cause

vLLM 0.22.1 decorates `GPUWorker.determine_available_memory()` with `@torch.inference_mode()`. Its V1 `GPUModelRunner.profile_run()` relies on that outer method for inference mode.

#448 replaced the worker method without preserving the decorator. Multimodal dummy profiling therefore retained the vision encoder's backward graph and intermediate activations. The A/B trace in #460 measured the process footprint increasing from about 2.9 GiB to 14.06 GiB with identical inputs.

This patch restores the original execution contract without changing the logical KV-capacity formula, skipping profiling, or hiding real CUDA errors.

Sorry for missing this contract in #448. The earlier small text-model validation did not exercise the multimodal profiling path where the regression becomes visible.

## Validation

- C++ unit tests
- pre-commit
- mypy on Python 3.9, 3.10, 3.11, 3.12 and 3.13
- CPU pytest on Python 3.9, 3.10, 3.11, 3.12 and 3.13: 186 passed per version

Fixes #460

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -2055,8 +2055,11 @@ class GPUWorkerPatch(VersionAwarePatch, BasePatch):
             self.logger.debug("Worker.determine_available_memory already patched")
             return True
 
+        import torch
+
         logger = self.logger
 
+        @torch.inference_mode()
         def _patched_determine_available_memory(
             self, *args: Any, **kwargs: Any
         ) -> int:
--- a/tests/test_vllm_virtual_kv_capacity.py
+++ b/tests/test_vllm_virtual_kv_capacity.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import contextlib
+import functools
 import importlib
 import sys
 import types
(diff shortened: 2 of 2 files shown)


RixinLiu:
LGTM, this indeed is a big problem, merging it now

PR #465, opened 2026-08-26
fix: split page create from map to avoid a remap race on Volta

andrewleech:
With `KVCACHED_CONTIGUOUS_LAYOUT=true` and page prealloc on, the background prealloc thread
does the page-table edit (`mem_unmap` the shared zero page, then `cuMemMap` a real page) over a
big compound page while compute kernels are still running. On Volta (V100 / sm_70) that edit
faults the running kernel and comes back as a `CUDA error: an illegal memory access` (700). It
only bites hybrid GDN/Mamba models under concurrency, and `KVCACHED_CONTIGUOUS_LAYOUT=false` has
been the workaround so far.

The fix splits `FTensor::map` into two halves so the two ends can run on different threads:

- `prepare()` does just the physical allocation (`cuMemCreate`) and stashes the page. It touches
  no mapped VA, so it's fine to run alongside live kernels.
- `commit()` does the VA edit (`mem_unmap` zero page + `cuMemMap` real), which has to happen while
  the GPU is idle.

The prealloc thread now only prepares. The VA edit moves to the main thread at the `alloc_page`
handout via `commit_pages()`, with a `device_synchronize()` in async mode. The expensive
`cuMemCreate` stays off the critical path, so elasticity and throughput are unchanged. `map()` =
`prepare()` + `commit()` is kept for existing callers (the worker IPC map path is untouched).
`commit()` is a no-op if the page is already mapped (a page can sit in the reserved pool still
mapped after a free), and `unmap()` just drops a prepared-but-uncommitted page with no VA edit.

## Testing

Validated on 2x V100-SXM2-16GB (the 1Cat sm_70 vLLM fork, OvisOCR2, TP1, contiguous layout,
prealloc on). The config that used to crash within a couple of reps (~33s) now runs 96/96 clean,
the output is numerically identical to the `CONTIGUOUS_LAYOUT=false` reference (16/16 pages), the
non-contiguous path still passes, and throughput is unchanged (88.8s vs an 84.8s no-reclaim
baseline, well within noise). Added `tests/test_create_map_split.py` covering the split
accounting: distinct physical pages after prepare/commit, data intact across free then realloc,
and `trim()` dropping uncommitted reserved pages.

Also validated on stock upstream vLLM 0.26.0 (kvcached `ac9680a`, this PR's base): rebuilt from
this branch in that image and ran the same repro on a Quadro RTX 8000 (SM75), contiguous=true, TP1:
48/48 clean on the config that host has always corrupted on. So the fix holds on stock upstream and
a second GPU arch.

## Caveats

A few things to flag though:

- This is the single-process map path only. The multi-process TP-worker-IPC map path is unchanged
  here.
- There's a separate TP2 crash (illegal access that surfaces in the NCCL watchdog) that I looked
  into, it's a different root and this PR doesn't touch it.


csrc/allocator.cpp
```diff
@@ -256,6 +256,57 @@ bool FTensorAllocator::unmap_from_kv_tensors(
...
+  if (num_layers_ == 0) {
+    LOGGER(ERROR, "try to map to KV tensors when KV tensors are not created");
+    return false;
+  }
+
+  if (contiguous_layout_) {
+    // Single contiguous tensor; each offset covers all layers.
+    auto ftensor = contiguous_kv_tensor_.get();
+    for (auto offset : offsets) {
+      op(ftensor, offset);
```
shipiyouniao, line 271: op() returns bool, but this result is discarded here and in the unified/split-buffer branches; for_each_mapping_ then returns true unconditionally. Its prepare_kv_tensors()/commit_kv_tensors() callers therefore cannot report a negative result to PageAllocator. If an already-mapped prepare is intentionally successful/idempotent, please express that explicitly rather than ignoring every false result. Add negative-result and exception cases so the two failure channels remain well-defined.

shipiyouniao (changes requested):
I checked 32789bc and found two items to address.

First, the new for_each_mapping_() discards the bool returned by prepare()/commit(). A CPU-only C++ harness using the exact visitor method returned true in all three layouts when its operation returned false. Please make the success/failure contract explicit; details inline. This is a return-contract test, not a GPU OOM reproduction, and exceptions still propagate separately.

Second, the test-classification check fails on this head:

```text
$ python tools/check_test_classification.py
test classification invalid: unclassified tests: tests/test_create_map_split.py
```

Please register the new module in tests/manifests/gpu.txt, since it requires the compiled extension and a GPU. The existing device tests exercise successful allocation/reuse/trim, so a focused failure-propagation test would be useful alongside them.

shipiyouniao:
One more point from revisiting the performance evidence: the V100/SM75 correctness runs are useful, but 88.8s versus an 84.8s no-reclaim baseline doesn't yet establish unchanged throughput. Reclaim behavior differs, and no repeat-to-repeat spread is given.

Since commit_pages() adds device_synchronize() on the async handout path, could you compare matched configurations on a layout/workload where both versions can finish, with repeated runs and growth/reclaim counts? The crashing configuration can remain separate correctness evidence. This isn't a claim of a measured regression; I'd just qualify the "throughput unchanged" wording until that comparison is available.

Issue #437, opened 2026-08-08
SIGABRT in FTensor::map() — mem_unmap fails with "invalid argument" under repeated large alloc/free churn


jiarong0907:
## Problem

A loop of large allocations with periodic bulk frees aborts the process within a few dozen iterations. Reproduces 3/3 on clean `main` (666671b), single L4, no engine integration involved.

```
csrc/ftensor.cpp:112 gpu_vmm::mem_unmap(vaddr, page_size_) failed in CUDA driver (1): invalid argument
Aborted (core dumped)     # exit 134
```

The call site is inside `FTensor::map()`, which unmaps before establishing a new mapping (`csrc/ftensor.cpp:100-118`):

```cpp
page_id_t page_id = offset / page_size_;
if (mapping_.find(page_id) != mapping_.end()) { ... return false; }
auto vaddr = ...;
if (dev_.is_cuda()) {
  CHECK_GPU(gpu_vmm::mem_unmap(vaddr, page_size_));   // :112
}
```

`CHECK_GPU` routes to `gpu_vmm::check()`, which calls `std::abort()` on any non-success status (`csrc/inc/gpu_vmm.hpp:148-156`). A driver error here takes down the whole engine process rather than surfacing as a Python exception.

`CUDA_ERROR_INVALID_VALUE` on `cuMemUnmap` for a page the code has just confirmed is *absent* from `mapping_` points at `mapping_` drifting out of sync with the device's actual mappings, rather than at OOM.

## Reproducer

Aborts after `iter 0` in all three runs, well before `iter 50` (~10s).

```python
import os, sys, time, torch
from kvcached.integration.vllm.interfaces import (alloc_kv_cache, get_kv_cache_manager, init_kvcached)
from kvcached.vmm_ops import kv_tensors_created

os.environ["KVCACHED_CONTIGUOUS_LAYOUT"] = "true"
torch.cuda.set_device(0)
init_kvcached(tp_rank=0, world_size=1, is_worker=True, async_sched=False)
alloc_kv_cache(kvcache_shape=(2, 65536, 16, 8, 64), block_size=16,
               dtype=torch.float16, device="cuda:0", num_layers=16)
t0 = time.time()
while not kv_tensors_created():
    if time.time() - t0 > 10: sys.exit("no tensors")
    time.sleep(0.1)

m = get_kv_cache_manager(65536, 16, 1024, 16, group_id=0)
m._post_init_done.wait(15)

held = []
for i in range(400):
[...]

PR #466, opened 2026-08-26
[DO NOT MERGE] fix: compare resize target against current allocator capacity

jeff3071:
## Summary

Fix the resize watcher failing to restore an allocator to its initial capacity after shrinking.

`mem_size_per_layer_` stores the startup capacity and is not updated by `resize()`. The watcher therefore treated the initial limit as unchanged even when the allocator was currently smaller.

Use `num_total_pages_ * page_size_` as the current allocator capacity instead.

## Test

Added a regression test covering:

1. Shrink from the initial capacity.
2. Restore the original limit.
3. Verify the watcher produces the expansion target.
4. Verify the allocator returns to its initial page count.

```bash
python -m pytest -v  tests/test_kvcache_manager.py::test_resize_watcher_can_restore_initial_capacity
```

### DIFF
--- a/csrc/inc/page_allocator.hpp
+++ b/csrc/inc/page_allocator.hpp
@@ -93,7 +93,8 @@ public:
 
   // Poll the shared-memory MemInfoStruct to see if an external controller
   // (e.g. `kvctl limit`) has written a new total_size. Returns the new
-  // per-layer mem_size if it differs from current_mem_size, otherwise -1.
+  // per-layer mem_size if it resolves to a different page count, otherwise
+  // -1.
   int64_t check_and_get_resize_target(int64_t current_mem_size) const;
 
   // Fast atomic read of the resize target (updated by background watcher
--- a/csrc/page_allocator.cpp
+++ b/csrc/page_allocator.cpp
@@ -502,8 +502,20 @@ PageAllocator::check_and_get_resize_target(int64_t current_mem_size) const {
   if (!mem_info_tracker_) {
     return -1;
   }
-  return mem_info_tracker_->check_and_get_resize_target(
+  const int64_t target = mem_info_tracker_->check_and_get_resize_target(
       current_mem_size, num_layers_, num_kv_buffers_);
+  if (target < 0) {
+    return -1;
+  }
+
+  const int64_t current_num_pages =
+      num_total_pages_.load(std::memory_order_relaxed);
+  // resize() floors byte targets to whole pages, so an unaligned target is
+  // satisfied once its page count matches the allocator's live capacity.
+  if (target / page_size_ == current_num_pages) {
+    return -1;
+  }
+  return target;
 }
 
 std::unordered_map<page_id_t, std::vector<int64_t>>
@@ -842,8 +854,7 @@ void PageAllocator::resize_watcher() {
     if (mem_info_tracker_) {
       const int64_t current_mem_size =
           num_total_pages_.load(std::memory_order_relaxed) * page_size_;
-      int64_t target = mem_info_tracker_->check_and_get_resize_target(
-          current_mem_size, num_layers_, num_kv_buffers_);
+      int64_t target = check_and_get_resize_target(current_mem_size);
       resize_target_.store(target, std::memory_order_relaxed);
     }
   }
--- a/csrc/page_allocator.cpp
+++ b/csrc/page_allocator.cpp
@@ -840,8 +840,10 @@ void PageAllocator::resize_watcher() {
       break;
     }
     if (mem_info_tracker_) {
+      const int64_t current_mem_size =
+          num_total_pages_.load(std::memory_order_relaxed) * page_size_;
       int64_t target = mem_info_tracker_->check_and_get_resize_target(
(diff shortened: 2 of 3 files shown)


rishabhsinha17 (commented):
Verified the failure mode on main before reading the fix. `mem_size_per_layer_` is set once in the constructor (page_allocator.cpp:112) and `resize()` only ever moves `num_total_pages_` (:362, :385, :412), so the watcher at :843-844 compares the shm-derived size against the startup capacity forever. Shrink via `update_kv_cache_limit`, resize, then restore the original limit, and `check_and_get_resize_target` (mem_info_tracker.hpp:191) sees new_mem_size == mem_size_per_layer_, returns -1, and the expand never gets posted, so `_alloc`'s poll (kv_cache_manager.py:344-346) never restores capacity. Reading the live capacity as `num_total_pages_ * page_size_` is the right baseline and the units line up, num_total_pages_ is initialized as mem_size_per_layer / page_size (:119) and resize() converts with the same page_size_ (:324); the relaxed atomic load matches every other reader of that counter. The regression test drives the real shm path and asserts both the posted target and the restored page count, and it correctly waits for -1 after the first resize so the second target cannot be a leftover. Note the test needs CUDA (the fixture pins cuda:0) and tests/test_kvcache_manager.py is not in tests/manifests/cpu.txt, so the green CPU CI on this PR never ran it; I reviewed it by analysis only.

Two interactions to sort out, the first one before this merges, because the change alters when the watcher speaks and not just what it says.

1. It makes the shm limit continuously authoritative, which unwinds revisioned limits from #414. `set_memory_limit` resizes through the same allocator but never writes the shm limit (kv_cache_manager.py:597-601), while MemInfoTracker init writes the startup total into the shm (mem_info_tracker.hpp:166). On main the watcher stays silent in the default state since baseline and shm both hold the startup value. With this change, apply a revisioned shrink and within 100ms the watcher sees shm != live capacity and posts an expand target, and the next `_alloc` resizes back up, undoing the applied shrink (and a lazily completed one as soon as it lands). Main already had this fight whenever the shm limit was explicitly changed away from startup, but this PR extends it to the untouched-shm default, so the precedence between the two limit channels needs to be stated, either the revisioned path mirroring its limit into the shm or the watcher respecting the manager's effective limit.

2. Comparing at page granularity would avoid a perpetual phantom target. The manager's mem_size is num_blocks * block_mem_size (kv_cache_manager.py:121), not necessarily a PAGE_SIZE multiple, and num_total_pages_ floors it (:119). The shm-derived size at startup is exactly mem_size_per_layer, so whenever that value is not page-aligned the new baseline can never equal it and an idle engine posts a target every 100ms, with every `_alloc` running a no-op resize and observability's resize_target_bytes pinned non-negative. Comparing `target / page_size_` against num_total_pages_ (or passing the page count into the check) makes satisfied-detection exact. Main has the same defect after any resize away from startup; here it becomes the default for unaligned sizes, so it is worth folding in.

One more, fine as a follow-up rather than in this PR. Now that expand-after-shrink actually works, the watcher path can also be the thing that expands past the FTensor's reserved virtual range if the shm limit is ever raised above the startup total, since nothing clamps the posted target and page ids past the reservation abort inside FTensor::map (the #437 failure class; #469 guards the construction seam, not this resize seam). Clamping the target to mem_size_per_layer_, which is exactly the startup VA reservation, would close that, and this PR conveniently keeps that member around as the natural cap. Pre-existing on main, so not blocking here.

jeff3071:
Thanks for the detailed review. I’ve addressed the page-granularity issue in c71e642: resize targets are now considered satisfied when their floored page count matches the allocator’s live page count. The regression test now uses an intentionally unaligned target.

The interaction with revisioned limits needs a separate design decision. I’ll mark this PR as [DO NOT MERGE] until that dependency is resolved. The virtual-address-cap clamp can also remain a follow-up as suggested.

rishabhsinha17:
c71e642 holds. The wrapper now floors the tracker's byte target and returns -1 when target / page_size_ equals the live page count (page_allocator.cpp:505-517), the watcher calls the wrapper instead of the raw tracker method (:857), and the header comment matches the new behavior. The regression test's target of initial minus one page plus a half page floors to the shrunken capacity, the case that used to phantom-post forever. Same caveat as before, the test needs CUDA and cpu.txt does not run it, so I checked it by reading.

On precedence, the smaller change is the revisioned path mirroring its limit into the shm at set_memory_limit time. set_memory_limit already computes effective_limit_bytes (kv_cache_manager.py:599) next to the resize call (:601); writing that value into the segment there, through the existing update_kv_cache_limit writer (kvcached/cli/utils.py:130) or a small tracker-side setter, gives the watcher one source of truth. After a revisioned shrink the shm-derived per-layer size equals the new live capacity and the watcher posts nothing, while the revision bookkeeping stays in-process where it already lives. Making the watcher respect the manager's effective limit instead needs a new shm field or C++ plumbing for in-process state, a bigger change for the same behavior. A later kvctl write still replaces the mirrored value, though the shm channel already works that way today. Which way to go is the maintainers' call, this only maps the cost of each option.

jeff3071:
https://github.com/ovg-project/kvcached/pull/499 resolve

Issue #470, opened 2026-08-27
Editable install does not install kvcached_autopatch.pth; KVCACHED_AUTOPATCH silently no-ops


rishabhsinha17:
Found while producing the MLA rows for #425.

**Repro**: `pip install -e . --no-build-isolation` into a fresh venv, set `ENABLE_KVCACHED=true KVCACHED_AUTOPATCH=1`, run vLLM 0.24. The engine runs pure vanilla: zero kvcached log lines, no error, no warning.

**Mechanism**: the wrapt import hooks live in `kvcached_autopatch.pth`, which only fires when it sits in site-packages, where the interpreter executes .pth files at startup. Wheel and regular installs put it there (`BuildPyWithPth` / `InstallWithPth`, setup.py:116/128). Editable installs do not: setup.py:139 defines `DevelopWithPth` for exactly this case, but with `build-backend = "setuptools.build_meta"` and setuptools>=64, `pip install -e .` goes through the PEP 660 `build_editable` hook and the legacy `develop` command never runs. So the .pth is never copied, nothing imports `kvcached.integration.*.autopatch`, and since `KVCACHED_AUTOPATCH` is only read inside those modules (kvcached/integration/vllm/autopatch.py:30), setting it has no effect.

**Workaround**: `cp kvcached_autopatch.pth <venv>/lib/python3.11/site-packages/`. After that all 9 vLLM patches apply normally.

**Why file this**: the failure is silent. An A/B comparison "with vs without kvcached" run from an editable install trivially passes while never exercising kvcached. Distinct from #320: there the .pth ran but the env var was set too late; here the .pth is absent entirely.

**Options**: install the .pth on the PEP 660 path too (a `build_editable` override, or packaging it so the editable wheel carries it); or warn loudly when `ENABLE_KVCACHED` is set but no autopatch module ever loaded; or document the copy step for editable installs. Happy to PR whichever you prefer.

PR #472, opened 2026-08-27
fix(vllm): bind cross-layer KV sharing layers to their target's cache

### LINKED ISSUE #417: Need support for Gemma model which has cross-layer kv sharing
## kvcached crashes with a bare `KeyError` on models with cross-layer KV sharing

`google/gemma-4-E2B-it` fails at KV-cache init under kvcached, on both vLLM 0.22.1 and 0.24.0. The same model runs fine with kvcached disabled.

```
KeyError: 'language_model.model.layers.15.self_attn.attn'
  File "kvcached/integration/vllm/patches.py", line 1285, in _allocate_kv_cache_from_kvcached
    tensor_cfg = layer_to_tensor_cfg[layer_name]
```

rishabhsinha17:
## Summary

Fixes #417. Models with cross-layer KV sharing (gemma E2B) crashed at startup with `KeyError: '...self_attn.attn'` in kvcached's layer-to-tensor binding. Skip the sharing layers wherever a layer is resolved against `kv_cache_tensors`, then alias each one to its target layer's cache, mirroring vanilla vLLM's `initialize_kv_cache_tensors`.

## Root cause

vLLM's `maybe_add_kv_sharing_layers_to_kv_cache_groups` appends sharing-layer names to `kv_cache_groups[*].layer_names` and records them in the runner's `runner_only_attn_layers` WITHOUT adding them to any `kv_cache_tensors[*].shared_by` (verified identical on 0.22.1/0.23.0/0.24.0, `v1/worker/utils.py`). Vanilla vLLM copes by skipping `runner_only_attn_layers` in `_allocate_kv_cache_tensors` / `_reshape_kv_cache_tensors` and aliasing `kv_caches[layer] = kv_caches[target]` after allocation (`gpu_model_runner.py:7064/:7115/:7266` at v0.24.0). kvcached's `_allocate_kv_cache_from_kvcached` did neither and direct-indexed every `grp.layer_names` entry, which is the KeyError site.

The appended names also inflated `_get_group_size`, so the worker would have allocated more pools than the scheduler-side `num_layers` (the scheduler's config never contains the appended names, the append is runner-side only).

## Changes

`kvcached/integration/vllm/patches.py`:
- New helpers `_get_runner_only_attn_layers`, `_tensor_backed_layer_names`, `_alias_shared_kv_layers`; `_get_group_size` takes an optional runner-only set (default keeps the old behavior; the scheduler-side call is unchanged).
- `_allocate_kv_cache_from_kvcached`: resolve only tensor-backed layer names in the validation loop, representative-layer selection, pool count, and the per-group hetero views.
- `_reshape_kv_cache_tensors_from_kvcached`: bind pools to tensor-backed names only, then alias sharing layers to their target's cache (fail-loud if the target has no cache). On vLLM releases whose vanilla `initialize_kv_cache_tensors` also aliases after the patched reshape returns, the second pass re-assigns the same objects.
- The V8 path (`kv_cache_config.tensors` schema) predates cross-layer sharing and is untouched.

`tests/test_shared_kv_layer_binding.py` (added to `tests/manifests/cpu.txt`): 14 CPU tests over the helpers, including a scenario shaped like the #417 report that reproduces the KeyError without the filter and a full binding+aliasing pass with it.

## Validation

- `python3.11 -m pytest tests/test_shared_kv_layer_binding.py`: 14 passed.
- Full CPU manifest: 190 passed (176 on main + 14 new); the 10 local errors are the pre-existing macOS /dev/shm + posix_ipc environment issues, unchanged from main.
- `tools/check_test_classification.py`: valid (cpu=23).
- ruff 0.11.7, isort 6.0.1 clean; `mypy --python-version 3.10` introduces no new errors vs main (error set identical).
- `git merge-tree --write-tree`: merges clean against open PRs #418/#455/#465 (2026-08-26).
- GPU, 1x L40S, vLLM 0.24.0, transformers 5.14.1, google/gemma-4-E2B-it, autopatch engaged (all 9 patches logged; note the issue's earlier PASS reports were measured with autopatch silently inactive, see #470):
  - main + a diagnostic log at the KeyError site: `sharing_active=True num_shared=20`, `group_layer_names=35 tensor_backed(shared_by)=15`, 20 layer names in groups but absent from every `shared_by`, then startup dies with `KeyError: 'language_model.model.layers.15.self_attn.attn'` in EngineCore.
  - this branch, same env and run: starts and completes the generation cell cleanly, zero KeyErrors.
  - Byte-level output parity vs vanilla was not re-measured in this session (main cannot start at all, so there is no kvcached-on baseline to compare before this fix); happy to add a parity run if wanted.

### DIFF
--- a/kvcached/integration/vllm/patches.py
+++ b/kvcached/integration/vllm/patches.py
@@ -11,7 +11,7 @@ import inspect
 import math
 import types
 from collections import OrderedDict
-from typing import TYPE_CHECKING, Any, Iterable, Optional
+from typing import TYPE_CHECKING, Any, Collection, Iterable, Mapping, Optional
 
 from kvcached.integration.patch_base import BasePatch, enable_kvcached
 from kvcached.integration.version_utils import VersionAwarePatch, VersionRange, version_range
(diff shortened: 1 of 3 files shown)


RixinLiu:
@shipiyouniao can you help me to review this PR?

shipiyouniao (commented):
@RixinLiu, I went through `f3b8a00`. I haven't found a new functional bug in this change. Filtering the runner-only layers before counting/binding pools, then aliasing them to their targets, matches vLLM's handling of cross-layer KV sharing.

The 14 new tests pass. I also exercised the actual patched allocation/reshape entry points with a small review harness: 12 cases passed on CPU and the same 12 passed with T4 CUDA tensors. These cover homogeneous/heterogeneous groups, both layouts, exact target aliasing and readback, a missing target followed by recovery, and delegation when kvcached is disabled. The harness substitutes the allocator/backend; it is not a native VMM or Gemma serving test.

Two follow-ups would make this much easier to sign off:

- Could you include tests that call the patched allocation/reshape methods? The current tests rebuild the binding loop around the helpers, so they would still pass if a helper stopped being used at a call site. In particular, assert that the pool count matches the scheduler's pre-append group size and that a sharing layer gets its target's exact tensor/view in the heterogeneous path.
- Could you add the Gemma output comparison you offered? The reference can be vanilla vLLM with kvcached disabled; main's startup failure doesn't prevent that comparison. A small fixed greedy prompt set, including repeated-prefix requests and more than one decode step, would check that fixing the startup exception hasn't left incorrect KV bindings. Please record the versions/settings and token outputs, and investigate any mismatch rather than requiring only successful generation.

These are validation gaps, not a reproduced correctness defect. I'm leaving this as a comment pending that final model-level check; no code redesign requested.

RixinLiu:
LGTM. I run the two layouts on my A100, the output matches exactly with vanilla vLLM.

RixinLiu (approved):

Issue #471, opened 2026-08-27
MLA models on vLLM 0.24: KVCacheManager post-init cannot reach worker IPC socket (Errno 2), alloc_page retry storm


rishabhsinha17:
Found while producing MLA validation rows for #425 (after fixing #470 so autopatch actually engages).

Env: vLLM 0.24.0, kvcached main 60cad94 (editable install + `kvcached_autopatch.pth` copied into site-packages per #470), 1x L40S, driver 570.124.06, torch 2.11.0+cu130 with the cuda-compat-13 shim, `ENABLE_KVCACHED=true KVCACHED_AUTOPATCH=1`.

Repro: serve deepseek-ai/DeepSeek-V2-Lite-Chat (either `KVCACHED_CONTIGUOUS_LAYOUT` value). Patches apply (`Applying 9 patches for vllm`, `Successfully patched vllm: ...`), PageAllocator initializes (`num_layers=27 ... num_kv_buffers=1`), then post-init fails:

```
[kvcached][INFO] [kv_cache_manager.py:173] Set up broadcast callbacks for multi-process (world_size=1, use_worker_ipc=True)
[kvcached][ERROR][kv_cache_manager.py:234] Error during KVCacheManager post-initialization: Worker 0 failed to check KV tensors created: [Errno 2] No such file or directory
```

After that every allocation retries forever: `alloc_page() failed after partially allocating 0/2 blocks; rolled back: Failed to map page 0: RuntimeError: Worker 0 failed to map: [Errno 2] No such file or directory` (`tp_ipc_util.py:217 _broadcast_map_to_kv_tensors`) alternating with `patches.py:2213 Shared physical KV pool is exhausted; reporting a scheduling miss so the engine can preempt and retry`. The loop wrote >100 MB of warnings in ~8 min; zero requests ever complete.

Control in the same session and env: Qwen/Qwen2.5-0.5B-Instruct serves normally (RC 0, generation fine), so this is not general env breakage. The 0.5B log does not show the `use_worker_ipc=True` broadcast setup, so my read is the worker-IPC broadcast path itself is what fails: whatever should start the worker-side listener socket does not (or the manager computes a different socket path), and MLA (`num_kv_buffers=1`) is the path that requires it. Same failure with `VLLM_ENABLE_V1_MULTIPROCESSING=0`, so it is not the engine-process split.

Full logs kept; happy to bisect or test candidate fixes.

Issue #412, opened 2026-07-27
Partial KV tensor mapping failures can leave inconsistent VMM state


shipiyouniao:
## Problem

`FTensorAllocator::map_to_kv_tensors()` and `unmap_from_kv_tensors()` perform a batch of VMM operations across offsets, layers, and K/V regions, but the batch has no defined transactional behavior.

The allocator currently ignores each `FTensor::map()` / `unmap()` return value and reports success after the loop. If an operation fails after earlier operations have succeeded, the process can be left with a partially mapped or partially unmapped KV layout while higher layers continue with inconsistent state.

This is separate from:

- #372, which propagates worker and TP/PP failures to the caller;
- #381, which makes repeated page operations idempotent.

Error propagation and idempotency do not restore a batch that has already partially changed state.

There is also a lower-level contract issue: CUDA/HIP VMM failures currently pass through `CHECK_GPU` and abort the process. A rollback design is unreachable for those failures unless the low-level API exposes a recoverable status or exception. If process termination is intentionally the only safe behavior for some VMM failures, that boundary should be explicit rather than mixed with recoverable `bool` results.

## Expected behavior

The mapping API should define and enforce atomic or explicitly fail-closed batch semantics:

- Distinguish a newly mapped page from an idempotent already-mapped page.
- On a recoverable map failure, roll back only mappings created by the current batch, in reverse order.
- Define equivalent semantics for partial unmap failures; do not silently leave a half-unmapped KV layout.
- Treat rollback failure as a first-class state-inconsistency error and fail loud.
- In TP/PP mode, compensate ranks that succeeded if another rank fails, or explicitly terminate the whole instance if distributed rollback cannot be made reliable.
- Preserve the original failure and any rollback failure in logs/errors.

## Required tests

- Failure injection at every operation position for contiguous, unified, and per-layer K/V layouts.
- A batch containing pre-existing idempotent mappings, proving rollback does not remove them.
- Rollback-failure injection and fail-closed behavior.
- TP and PP partial-rank success followed by one-rank failure.
- Real CUDA/HIP VMM validation in addition to CPU/unit mocks.

## Scope

This issue tracks allocator/VMM correctness only. It does not add quota, policy, or control-plane behavior. No implementation is attached yet; the intent is to agree on the failure contract before changing the low-level API.

PR #478, opened 2026-09-02
feat: lifecycle readiness and failure propagation for KVCacheManager (#375 item 5, phase 1)

rishabhsinha17:
## Summary

Part of #375, item (5) phase 1. Stacked on #474; will rebase once it merges
(the first three commits here are #474's head, the rest are this change;
rebased onto its 09-03 update). This is the progress update @shipiyouniao
asked for on 08-27, and it follows the split @ryanManocha agreed to: item
(5) lands the one lifecycle/readiness key under the `features` record, the
rest of item (3) is his and is now in #474.

`KVCacheManager._post_init()` runs in a daemon thread. When it raises (KV
tensors never created, or the worker IPC check itself failing as in #471,
"Exception in thread Thread-2 (_post_init)") the error dies with the thread
while `_post_init_done` still opens the gate, so callers walk into a pool
with no null block and no prealloc thread, and nothing queryable says why.
This adds a poll-only lifecycle phase per pool, a `wait_ready()` gate that
re-raises the captured error, the phase as an additive snapshot field, and
the capability key.

## Design

Phases: `INITIALIZING`, `READY`, `DEGRADED` (still serving, accounting
suspect), `FAILED`. State is the surface; there are no callbacks, so nothing
re-enters consumer code from an allocator thread (#371). `DEGRADED` and
`FAILED` are sticky; recovery stays a control-plane decision.

The two rules from the thread, as implemented:

1. Broadcast failures (revised per review). An unmap broadcast that fails
   moves the pool from `READY` to `DEGRADED`, a timeout included: the block
   ledger is updated before `free_pages()` reaches the broadcast, no caller
   rolls it back, and without per-rank delivery tracking the outcome across
   TP/PP ranks is unknown. A map broadcast that fails changes no lifecycle
   state in phase 1: it is routinely the expected co-tenancy capacity miss
   (#453) that `alloc_page()` rolls back and `_alloc()` reports as a
   scheduling miss, and `tp_ipc_util` raises one `RuntimeError` shape for
   every per-rank failure, transport or application, naming only the first
   failing rank, so this layer cannot tell that recoverable miss from an
   unknown or partial outcome without matching message text. #373 has
   landed and still raises that flat `RuntimeError` shape (renaming the
   target `pp{p}/rank{r}` and propagating worker-side failures that used
   to report success). Rather than publish a sticky
   false `DEGRADED` for normal memory pressure, phase 1 records nothing on
   map failures; per-rank classification (safe miss stays `READY`, unknown
   `DEGRADED`, confirmed partial `FAILED`) is phase 2, once the broadcast
   returns structured per-rank results. Phase 1 still does not touch
   `tp_ipc_util.py`. Both callback wrappers re-raise unchanged, so
   `alloc_page()` and the prealloc thread see exactly what they saw before.
2. `wait_ready(timeout=None)` blocks until the phase leaves `INITIALIZING`,
   returns on `READY` or `DEGRADED`, re-raises the `_post_init` exception
   object (traceback preserved) on `FAILED`, and raises `TimeoutError` if
   still initializing after `timeout`. `_post_init_done` is set before the
   phase is recorded, so once `wait_ready()` returns the legacy gate is
   already open. `KVCachePoolSnapshot.lifecycle_phase` carries the phase
   through the #385 snapshot contract, so a consumer polling snapshots sees
   `failed` and the cause is available from `lifecycle_error`; wiring
   `wait_ready()` into the engine attach path is a later phase.

Record-only, per the split agreed with @RixinLiu: the existing entry points
(`alloc()`, `free()`, `resize()`, `trim()`, `try_to_reserve()`) and
`_post_init_done` behave exactly as before; a caller that never polls sees
no difference. No new IPC messages, threads, env vars or callbacks.

Transitions are logged at a level matching the phase entered: INFO for
`READY` and `INITIALIZING`, WARNING for `DEGRADED`, ERROR for `FAILED`.

Capability key: `get_capabilities()["features"]["lifecycle_readiness"]` is
`True`, one bool under the `features` record from #474, no schema bump.
`lifecycle_phase` shows up in `pool_snapshot_fields` automatically.

Not in this PR (later phases, as listed in #375): per-rank delivery
classification, the remaining lifecycle record (per-checkpoint readiness,
IPC expected-vs-confirmed, last-error records and counters), integrations
calling `wait_ready()` at attach, the exception taxonomy,
`KV_TENSOR_WAIT_TIMEOUT` as an env knob, `free()` ledger inconsistency and
prealloc-thread death as `DEGRADED`/`FAILED`, group-wide rollback (#412),
and any loud-failure flip of the soft entry points.

## Changes

- `kvcached/lifecycle.py` (new): `LifecyclePhase` (str enum) and
  `LifecycleState`, a condition-variable holder; a degradation seen while
  `INITIALIZING` is applied when that window ends. Imports neither torch nor
  the extension.
- `kvcached/kv_cache_manager.py`: `_post_init()` records `READY`/`FAILED`;
  `clear()` re-enters `INITIALIZING` for its window and fails the pool if it
  raises; the unmap callback wrapper records the failure and re-raises (it
  stores a weak reference to the state holder, and the manager is not
  reachable from the callbacks the C++ allocator owns through
  `std::function`, even through the saved error's traceback), while the
  map wrapper is
  a plain signature adapter per rule 1; `wait_ready()`, `lifecycle_phase`,
  `lifecycle_error`.
- `kvcached/observability.py`: `KVCachePoolSnapshot.lifecycle_phase`
  (`Optional[str]`, default `None`, so manager-like objects without a
  lifecycle keep working) and the `lifecycle_readiness` key.
- `tests/test_lifecycle_readiness.py` (new, CPU manifest): transitions,
  re-raise of a simulated `_post_init` failure (same object, `_post_init` in
  the traceback) and of #471's IPC-check failure, `wait_ready()` timeout and
  lock behavior, unmap failure to `DEGRADED`, map failure leaving the phase
  alone, the #453 recoverable capacity miss through the real `_alloc()`
  path (the miss returns `None`, the phase stays `READY`, and the next
  attempt allocates once pressure clears), an end-to-end
  `KVCACHED_IPC_TIMEOUT` case through the real broadcast path against a
  unix socket that never answers (unmap degrades, map does not), the
  `clear()` window, `LifecycleState` on its own, the snapshot field, the
  capability key, and the callback-lifetime pair from the 09-15 review (the
  saved unmap error must not make the manager uncollectable through the
  natively owned callbacks, and a callback invoked after its pool is
  gone still re-raises). Four lines in `tests/test_observability.py`.
- `tests/test_lifecycle_readiness_native.py` (new, GPU manifest): the same
  lifetime property against the compiled extension and real VMM mapping,
  three healthy and three failed pools in a subprocess.

## Validation

All numbers below are from b688d630.

- `tests/test_lifecycle_readiness.py`: 25 passed;
  `tests/test_observability.py`: 17 passed. Reverting the
  `kv_cache_manager.py` hunk alone makes
  `test_retained_unmap_callback_does_not_own_failed_manager` fail at the
  collection assert. Restoring it passes.
- `tests/test_lifecycle_readiness_native.py` collects and skips without
  CUDA here. On a T4 it fails at the parent commit and passes with the
  fix (3/3 failed-unmap managers retained before, 0/3 after, per the
  review thread).
- Full CPU manifest (`bash tools/run_cpu_tests.sh`): 225 passed; the 10
  errors are the pre-existing macOS-only fixture failures (`/dev/shm`,
  AF_UNIX path length), and the one failure
  (`test_listener_thread_restores_cuda_device`) is macOS-only as well and
  present at the previous head. Both files are green in CI.
- `python tools/check_test_classification.py`: valid (cpu=28, gpu=8).
- ruff 0.11.7, isort 6.0.1, codespell 2.4.1: clean. mypy 1.11.1
  (`--python-version 3.10`, CI flags): clean on the touched files.
- Upstream CI green on b688d630: CPU pytest (3.9 to 3.13), MyPy (3.9 to
  3.13), pre-commit.


kvcached/kv_cache_manager.py
```diff
@@ -157,14 +162,30 @@ def __init__(
...
+                lifecycle = self._lifecycle
+
                 # Wrap Python functions to match C++ callback signature
                 def map_callback(world_size: int, offsets: List[int], pp_rank: int = 0, group_id: int = 0) -> None:
                     """Wrapper for Python broadcast function"""
-                    broadcast_map_to_kv_tensors(world_size, offsets, pp_rank, group_id)
+                    try:
+                        broadcast_map_to_kv_tensors(world_size, offsets, pp_rank, group_id)
+                    except Exception as exc:
+                        lifecycle.record_broadcast_failure("map", exc)
```
shipiyouniao: This catches every map-broadcast exception before the existing allocation path can classify it. Today `KVCacheManager._alloc()` deliberately catches an `alloc_page()` `RuntimeError`, rolls back, and returns `None` for a recoverable allocation miss when a colocated instance consumes the remaining physical capacity. With this wrapper, that expected pressure event first marks the pool permanently `DEGRADED`, even when no rank changed mapping state and the rollback is complete.

Until #373 (or an equivalent structured result) can distinguish `not_delivered` / confirmed no-change from unknown or partial outcomes, I think phase 1 should avoid changing lifecycle state for map failures. Alternatively, land the classification here and keep `READY` for a provably safe allocation miss, `DEGRADED` for unknown outcome, and `FAILED` for confirmed partial state. Please add a test through `_alloc()` rather than invoking the callback directly, so the normal recoverable capacity path is covered.

shipiyouniao (commented):
The poll-only state holder and `wait_ready()` gate are the right direction, and the stacked dependency on #474 is clear. I found one lifecycle-semantic blocker in the map callback below: phase 1 currently cannot distinguish an expected, recoverable capacity miss from an unknown/partial distributed outcome, but it makes both permanently `DEGRADED`. The lifecycle transition should wait for structured outcome classification, or phase 1 should omit the map-failure transition rather than publish a sticky false health state.

kvcached/kv_cache_manager.py
rishabhsinha17: Fixed in 1866110, taking your first option. The map wrapper is now a plain adapter that re-raises unchanged and records nothing (kv_cache_manager.py:195), so the capacity miss reaches _alloc()'s existing rollback-and-return-None path with the phase left READY. I could not find a safe way to classify at this layer: tp_ipc_util raises the same RuntimeError shape for transport and application failures and names only the first failing rank, so telling the recoverable miss from an unknown outcome would mean matching message text you are rewriting in #373. Classification lands in phase 2 on that structured result. Unmap failures still degrade; the block ledger is updated before free_pages() reaches the broadcast and nothing rolls it back.

Added the test through _alloc() you asked for: test_capacity_miss_through_alloc_stays_ready_and_recovers drives a worker map failure inside alloc_page(), asserts the miss returns None with the phase READY, and allocates on the next attempt once pressure clears. The end-to-end timeout test now pins that unmap degrades while map does not. Also rebased onto #474's current head. Thanks for catching this.

kvcached/lifecycle.py
```diff
@@ -0,0 +1,185 @@
...
+            if self._phase is LifecyclePhase.FAILED:
+                return
+            old = self._set(LifecyclePhase.FAILED, reason, error)
+        self._log(old, LifecyclePhase.FAILED, reason)
+
+    def begin_reinit(self) -> None:
+        """READY -> INITIALIZING for the ``clear()`` window. No-op otherwise."""
+        reason = "clear() in progress"
+        with self._cond:
+            if self._phase is not LifecyclePhase.READY:
```
shipiyouniao: A DEGRADED pool returns here, so KVCacheManager.clear() tears down and rebuilds the pool while the phase remains DEGRADED. Since DEGRADED means still serving and wait_ready() treats it as settled, a concurrent consumer can pass the readiness gate during the clear window even after the prealloc thread has stopped and mappings are being released. Stickiness should preserve the degraded cause after reinitialization, not suppress the transient INITIALIZING phase: carry the current cause into _pending_degraded, transition DEGRADED -> INITIALIZING, and let mark_ready() settle back to DEGRADED. Please extend the degraded-clear test to block inside _clear_locked() and verify that wait_ready() does not return until clear settles. FAILED can remain a no-op here.

shipiyouniao (commented):
The previous map-capacity lifecycle issue is fixed: the callback is now a plain adapter, and the new `_alloc()` test covers rollback, `None`, and recovery while remaining `READY`. I found one remaining readiness-state issue around `clear()` on an already degraded pool.

kvcached/lifecycle.py
rishabhsinha17: Fixed in 2bbd755. begin_reinit() (lifecycle.py:155) now leaves DEGRADED as well as READY: it carries the current (reason, error) into _pending_degraded and transitions to INITIALIZING, so the gate holds while _clear_locked() runs. mark_ready() at clear completion settles back to DEGRADED with that cause, not READY. A new degradation inside the window does not displace the carried first cause. FAILED is still a no-op.

test_clear_keeps_a_degraded_pool_degraded now blocks inside _clear_locked() via a hook on reset_free_page_order and checks the window: phase INITIALIZING, wait_ready(0.05) raises TimeoutError, a threaded wait_ready(5) stays blocked, and after release the waiter sees DEGRADED with the original error object. test_begin_reinit_carries_a_degraded_cause_through_the_window pins the carry on LifecycleState alone. Both fail against the previous begin_reinit. Thanks for the clear spec.

rishabhsinha17:
Rebased across #373 and #413. Beyond the mechanical rebase, three composition changes. The unmap wrapper keeps #373's self-bound pp_rank/group_id defaults with the DEGRADED recording inside it. The #471 test now expects the wait-loop TimeoutError naming the last IPC error, since #373's wait loop retries a raising check instead of dying on the first raise, and FAILED records that TimeoutError. Rule 1 is otherwise unchanged, #373 landed still raising one flat RuntimeError per broadcast, so map failures keep recording nothing until per-rank results exist. Also gave #373's new bare-manager post-init test the lifecycle holder it now needs, and added the resolve_gpu_device_index stub to the shim-guard test carried from #474, same as cf46ed7 did for the two factory tests (@ryanManocha, #474 will need that line too). All the touched test files pass locally and the pre-commit pins come back clean.

kvcached/kv_cache_manager.py
```diff
@@ -174,7 +211,11 @@ def unmap_callback(
                     group_id: int = self.group_id,
                 ) -> None:
                     """Wrapper for Python broadcast function"""
-                    broadcast_unmap_from_kv_tensors(world_size, offsets, pp_rank, group_id)
+                    try:
+                        broadcast_unmap_from_kv_tensors(world_size, offsets, pp_rank, group_id)
+                    except Exception as exc:
+                        lifecycle.record_broadcast_failure("unmap", exc)
+                        raise
```
shipiyouniao, line 222: [P1] Avoid retaining the manager indirectly through the saved exception. Although the closure captures LifecycleState rather than self, that state now strongly stores exc. After this exception propagates through free(), its live traceback includes frames holding the manager. The native PageAllocator owns the Python callback through std::function, so the resulting allocator -> callback -> lifecycle -> exception/traceback -> manager -> allocator cycle is not collectable by Python GC. I reproduced this with the compiled extension on T4: 0/3 healthy managers retained versus 3/3 after a real worker-IPC unmap error, even after clear() and gc.collect(); clearing only lifecycle_error.__traceback__ released all three. This pins the allocator and its cleanup resources across pool removal. Please break the native strong-reference edge (for example, capture a weak reference to the lifecycle holder), or retain diagnostics without live manager-owning frames, while preserving the intended wait_ready() error contract. A native callback lifetime test should cover this, not only FakePageAllocator.

shipiyouniao (dismissed):
Re-reviewed cf91ca2, including the rebase onto #373/#413. The previous capacity-miss and degraded-clear findings are fixed, and the callback defaults preserve the PP/group context.

Validation on Linux: lifecycle tests 23 passed; observability tests 17 passed; TP/PP broadcast and listener-device tests 12 passed. I also built this head's native extension and exercised real VMM mapping, tensor writes and worker IPC on one Tesla T4. Five injected map failures recovered without leaving READY; five injected unmap failures entered DEGRADED, and clear() kept the cause while passing through INITIALIZING. This is single-device coverage, not qualification of partial multi-rank mapping outcomes.

The additional native cleanup check found a blocker: retaining the unmap exception keeps the manager alive through the C++ callback. In a separate three-pool control, healthy pools left 0/3 managers alive after clear(), dropping application references and gc.collect(); failed-unmap pools left 3/3 alive. Detaching only the saved exception's __traceback__ then reduced that to 0/3, without changing the allocator or callbacks. The inline comment explains the ownership cycle. This needs a lifetime regression test with the native PageAllocator, since a Python fake does not reproduce the C++ ownership edge.

Also, the description still refers to #373 as future structured-result work. The updated code correctly says it has landed but still raises a flat RuntimeError; please make the description match when revising.

kvcached/kv_cache_manager.py
shipiyouniao: Added b688d630d20a06823cc90d88e3105fd12e519a60 using a weak lifecycle reference, preserving the original exception and traceback.

The new native T4 regression fails on the parent and passes with the fix: 0/3 failed-unmap managers retained after GC, versus 3/3 before; the healthy case remains 0/3. Five map-failure and five unmap-failure injections through real worker IPC also passed, including retry/write/free and degraded-clear behavior.

Full CI (Python 3.9-3.13 CPU/MyPy, pre-commit and C++ checks) passed before pushing. The hardware tests cover single-GPU callback ownership, not distributed TP/MPS recovery.

Please check the appended commit when you have time; your existing commits are unchanged.

rishabhsinha17: Thanks for taking this one yourself, and for the T4 repro that pinned it. b688d630 checks out here, and the weak capture matches the ownership graph in your comment. After the change the std::function edge stores only a weakref, a recorded failure leaves an ordinary manager/lifecycle/traceback cycle for the collector, and the re-raise contract holds, including once the holder is gone. That break also covers a FAILED record from _post_init, whose traceback holds self the same way. Reverting only the kv_cache_manager.py hunk makes test_retained_unmap_callback_does_not_own_failed_manager fail at the collection assert, so the CPU test does pin the leak. Locally at b688d630 the lifecycle file is 25 passed, the native test collects and skips without CUDA, the full CPU manifest is 225 passed with the same pre-existing macOS-only errors, and upstream CI is green. I had an equivalent fix staged locally and dropped it for yours. The description now says #373 landed and still raises the flat RuntimeError shape, and its Changes bullet describes the weak reference instead of the old capture wording.

shipiyouniao (dismissed):
LGTM.
@RixinLiu This looks ready to merge from my side, could you give it a final double-check?

rishabhsinha17:
Head unchanged at b688d630 since the weak-capture fix, all 11 checks green, and shipiyouniao's approval sits on this exact head. Marking ready for your merge pass.

kvcached/lifecycle.py
```diff
@@ -0,0 +1,199 @@
...
+                lambda: self._phase is not LifecyclePhase.INITIALIZING, timeout)
+
+    def raise_if_failed(self) -> None:
+        """Re-raise the recorded error if the pool is FAILED."""
+        with self._cond:
+            phase, error, reason = self._phase, self._error, self._reason
+        if phase is not LifecyclePhase.FAILED:
+            return
+        if error is not None:
+            raise error
```
shipiyouniao: Each `raise error` appends another caller traceback to the same exception held by LifecycleState. With the state object still alive, repeated failed readiness checks therefore retain caller frames and their local objects indefinitely. I reproduced 1 -> 201 retained frames after 100 calls. Please preserve the original failure without growing its stored traceback on every read, and cover repeated wait_ready()/raise_if_failed() calls.

shipiyouniao (changes requested):
I missed a failure-read lifetime issue in my earlier pass, so I've withdrawn my approval. The event-driven lifecycle API still makes sense; this is a specific problem with rethrowing the stored exception, not an objection to the feature.

On b688d63, a CPU probe against the actual LifecycleState records one initialization error and calls raise_if_failed() 100 times, catching the error each time. The retained traceback grows from 1 frame to 201. Repeated failed wait_ready() calls reach this path; merely reading lifecycle_phase does not.

Could you add a repeated-failure regression and avoid accumulating caller tracebacks on the stored error? This is a deterministic Python lifetime check, not a reproduced GPU failure. After rebasing, please also check the failure-state contract against the transaction errors introduced by #418. The earlier T4 unmap/GC results remain useful, but did not cover repeated reads of a failed initialization.

kvcached/lifecycle.py
rishabhsinha17: fixed in df3b64e. the stored failure keeps its record-time traceback and every re-raise rewinds to it first, so reading a failed pool no longer appends the reader's frames to the stored exception (the growth is what retained every caller's locals). your repro is the regression, one recorded error read 100 times through raise_if_failed and wait_ready stays pinned, same object still raised, instead of 1 to 201. the #418 check found a real gap, StateConsistencyError now records FAILED from both broadcast directions, it is a definitive stop-the-engine verdict, with regressions on the unmap wrapper and _alloc path and the ipc-timeout test updated to match. rebased, lifecycle file 30 passed, cpu manifest 339 passed, failure set byte-identical to main.

kvcached/lifecycle.py
```diff
@@ -0,0 +1,230 @@
...
+        the manager calls this for unmap broadcasts only (see the class
+        docstring for why map failures are excluded).
+
+        One typed outcome is stronger than unknown: #418's
+        ``StateConsistencyError`` means the transaction could not establish
+        that every affected address is contained, and its contract is to
+        stop the affected engine, so the pool is FAILED rather than
+        DEGRADED-but-serving.
+        """
+        if isinstance(exc, StateConsistencyError):
```
shipiyouniao, line 235: [P2] Record the native verdict after the callback returns

This check only sees the exception raised inside the Python callback. An untyped failure is recorded as DEGRADED here, but `PageAllocator::unmap_pages()` then calls `fail_pool()` and converts that same failure to `StateConsistencyError`. `KVCacheManager.free()` propagates it without updating lifecycle state. On T4 I reproduced native FAILED + Python DEGRADED + a successful `wait_ready()` in all three injected cases. Even the next alloc leaves it DEGRADED because `available_size()` raises before the new alloc_page try/except. Please record the propagated fatal exception at the manager/native boundary as well, including free/trim and the capacity check, and add a regression that checks native and Python states agree. A confirmed-aborted prepare can also raise an ordinary RuntimeError, so this is not limited to a malformed exception from a test double.

shipiyouniao (changes requested):
Thanks for the traceback fix and the typed failure handling. I reran the CPU manifest at `0d556b4`: 350 passed. The real-native test still fails on T4, though: `tests/test_lifecycle_readiness_native.py` reaches a `StateConsistencyError` in `manager.clear()` after its injected unmap failure. It still expects the pre-#418 degraded-and-clear behavior.

A separate real-VMM probe confirmed the underlying readiness mismatch in three out of three injections:

| Observation after an untyped unmap callback failure | Result |
| --- | --- |
| Exception from `manager.free()` | `StateConsistencyError` |
| Native `get_transaction_state()` | `FAILED` |
| Python `lifecycle_phase` | `DEGRADED` |
| `wait_ready()` | Returns successfully |
| Next `alloc(1)` | Raises `StateConsistencyError`, phase still `DEGRADED` |

The healthy control mapped a page, passed CUDA write/readback, freed it, and remained READY. This used the real C++ allocator with a local callback on one T4; it is not a distributed TP test. I reused the compiled extension only after checking the C++ tree matches this head.

The remaining boundary is detailed inline. Could you also update the native lifetime test to exercise disposal of a FAILED pool, rather than trying to clear it back into service? No need to soften #418's fail-closed behavior; readiness should reflect it.

kvcached/lifecycle.py
rishabhsinha17: fixed in d2ea47e. the manager records the fatal verdict where it propagates out of a native call, free, trim, resize, and the available_size capacity check, plus _alloc's untyped branch, so your 3/3 repro ends FAILED on both sides and wait_ready raises. untyped errors, the confirmed-aborted prepare included, defer to get_transaction_state instead of the exception type, and the phase does not change while that reads HEALTHY, so the #453 miss keeps returning None. cpu regressions replay your sequence against a double mirroring the C++ fail_pool contract and assert phase and native state agree at each boundary, six fail without the fix. the native test's fail arm asserts agreement and fail-closed wait_ready, alloc, and clear, then disposes of the pool instead of clearing it (f19eba2). i cannot run that on gpu here, cpu manifest 346 passed with the failure set byte-identical to main, so the T4 rerun stays yours.

shipiyouniao:
The native lifetime regression now passes on T4: three healthy pools and three injected unmap-failure pools, with Python/native FAILED agreement and successful object collection.

I merged main locally to check the conflicts with #456, but found one remaining issue before pushing: after the injected unmap failure, `manager.lifecycle_phase` is FAILED, while `manager.observability_snapshot()` raises `StateConsistencyError`. The snapshot calls `available_size()` before exporting `lifecycle_phase`; the native free-page getter rejects the failed pool. This reproduced in all three failed pools. Could we keep the failed phase queryable through snapshots without weakening the allocation failure checks? Otherwise the documented snapshot polling path loses the pool exactly when it needs to report failure.

The Python 3.10 CPU run also caught `test_repeated_wait_ready_on_a_failed_pool_keeps_the_traceback_flat`: 361 passed, one failed (`4 == 2`). The initial baseline is sampled after joining the init thread, but the first readiness raise happens before that join and shares the exception object with the thread. Please make that baseline deterministic after the thread finishes. This is not evidence that the old unbounded traceback growth has returned.

The local conflict resolution is ready, but I have not pushed it while these checks are unresolved.

rishabhsinha17:
fixed in 063e434 and 0247973, both additive on f19eba2c, no rebase or merge, your held #456 resolution applies as is. fork CI is green on all 11 jobs at the new head.

the snapshot no longer raises on a failed pool. a failed native pool rejects one read, the free-page count behind available_size, and the builder now catches that verdict and reports available_blocks 0 with lifecycle_phase failed. page counts the allocator still answers keep their values. available_size and the alloc path keep raising, regressions pin both paths plus the background-failure discovery.

the flaky baseline now comes from a fresh wait_ready raise after the init thread joins, the first raise races the thread propagating the same exception object. your 4 == 2 reproduced here, 38 of 60 runs on python 3.10.10 and 35 of 60 on 3.11.2, then 0 of 60 on both after the fix.

shipiyouniao (approved):
Thanks, the snapshot fix and the traceback baseline both check out. I merged main in `1f6aafd`, keeping #456's capacity-cache invalidation and this PR's failure reporting.

I also fixed an init/clear race: the init event could open before READY was published, so the late READY could overwrite a clear already in progress. `clear()` now waits for that publication. The sync/async regressions fail before the fix and pass after it. The native test also checks FAILED snapshots without weakening allocation/readiness errors.

Validation on this head:

- Full pre-commit, GPU-free C++, and CPU/MyPy on Python 3.9-3.13; 393 CPU tests per interpreter.
- T4 native lifetime regression: three healthy pools and three injected unmap failures; Python/native FAILED agreement, readable snapshots, and successful object collection.
- T4 VMM transaction/failure suite: 30 passed. This is single-device coverage, not distributed TP/PP qualification.

Phase 1 still reports recorded state. Direct phase/readiness reads do not poll native background health; an otherwise unobserved preallocation failure needs an allocator/snapshot read to surface. That remains follow-up work, not a guarantee of this approval.

The previous blocking findings are addressed. @RixinLiu, handing this back to you for final review. The description's base and validation section could use a refresh before merge.

shipiyouniao:
Merged the latest main after #482, keeping both sets of tests and the readiness state in the shutdown test fixtures.

The combination exposed one small cancellation gap: stopping initialization before the tensor handshake left readiness at INITIALIZING; stopping during null-block reservation could report READY. Both cases now settle as FAILED with an explicit initialization-cancelled reason, without throwing from the shutdown thread or starting preallocation. The two regressions failed before the fix and pass now.

Validation on `e5c36da`:

- Full private CI: pre-commit, GPU-free C++, and CPU/MyPy on Python 3.9-3.13, with 522 CPU tests per interpreter.
- T4: 19 real-native readiness/lifetime and shared-segment tests, plus 30 VMM transaction/failure tests. The readiness test exercises three healthy pools and three injected unmap failures, checking fail-closed behavior, readable failed snapshots, and object collection. Both startup-cancellation cases also passed against the native allocator, including shared-segment removal.
- Dedicated Codex review completed. Its native-background-health polling finding is the same phase-1 limitation called out in the previous review: direct readiness reads report recorded state; an allocator or snapshot read surfaces a previously unobserved native failure. This merge does not add polling to those accessors.

This is single-device coverage, not new distributed TP/PP qualification. The original commits are preserved; this is an additive merge, not a history rewrite.

Issue #477, opened 2026-09-02
Graceful SIGTERM leaks the /dev/shm segment on vLLM: the only cleanup is the C++ MemInfoTracker destructor, which the forked EngineCore never runs


rishabhsinha17:
Stopping a kvcached-enabled vLLM 0.24 server with SIGTERM leaves `/dev/shm/kvcached_vLLM_<PGID>` (or the explicit `KVCACHED_IPC_NAME`) behind after every process has exited. SIGKILL leaking it is expected; SIGTERM was not. `kvctl list` / `kvtop` keep reporting the dead segments as live pools with their old limits, and once a PGID is reused the exists-probe in `_obtain_default_ipc_name` (`kvcached/utils.py:92-98`) starts handing out `_1`, `_2` suffixes.

Env: vLLM 0.24.0, kvcached main 60cad94 (editable install + `kvcached_autopatch.pth` copied into site-packages per #470), 1x L40S, `ENABLE_KVCACHED=true KVCACHED_AUTOPATCH=1`, TP=1, default `VLLM_WORKER_MULTIPROC_METHOD` (fork).

Repro:

1. `ENABLE_KVCACHED=true KVCACHED_AUTOPATCH=1 setsid vllm serve Qwen/Qwen2.5-0.5B-Instruct --port 20000 &`, wait for `/health`; `kvctl list` shows `kvcached_vLLM_<PGID>`.
2. `kill -TERM -- -<PGID>` (whole group; `kill -TERM <api server pid>` behaves the same), wait until no `EngineCore` process is left.
3. `ls /dev/shm | grep kvcached_`; `kvctl list`.

Observed (Aug 27): after `kill -TERM` on the process groups (37957, 40274) of two engines launched with `KVCACHED_IPC_NAME=kvc_m1|kvc_m2`, all processes exited (a vanilla vLLM pair re-bound the same two ports 90 s later) and `/dev/shm/kvc_m1` and `/dev/shm/kvc_m2` were both still present 44 s after the signal. Earlier launches on the same pod had already left five segments (`ls -la /dev/shm`; none of these processes was alive):

```
-rw-r--r-- 1 root root 24 Aug 27 16:26 kvcached_vLLM_32878
-rw-r--r-- 1 root root 24 Aug 27 16:54 kvcached_vLLM_34028
-rw-r--r-- 1 root root 24 Aug 27 17:03 kvcached_vLLM_34618
[...]

PR #480, opened 2026-09-02
fix: remove TP worker socket dirs on shutdown and in kvctl delete

### LINKED ISSUE #476: TP worker socket dirs under /tmp are created at TP=1 and never removed; kvctl delete does not clean them
Every kvcached engine launch creates `/tmp/kvcached-tp-<ipc_name>-<uuid5[:8]>/w0.sock`, at TP=1 too, and nothing ever removes it. `kvctl delete <ipc>` removes the `/dev/shm` segment but not the socket dir, so the dirs accumulate across launches.

Env: vLLM 0.24.0, kvcached main 60cad94 (editable install + `kvcached_autopatch.pth` copied into site-packages per #470), 1x L40S, `ENABLE_KVCACHED=true KVCACHED_AUTOPATCH=1`, TP=1.

Code (upstream main 60cad94):

- `kvcached/tp_ipc_util.py:25-32`: `SOCKET_DIR = /tmp/kvcached-tp-{DEFAULT_IPC_NAME}-{uuid5(NAMESPACE_DNS, DEFAULT_IPC_NAME).hex[:8]}`. `DEFAULT_IPC_NAME` embeds the PGID (`kvcached/utils.py:80-98`), so every launch gets a fresh directory name.
- `kvcached/tp_ipc_util.py:102-114`: `start_worker_listener_thread()` does `os.makedirs(socket_dir, exist_ok=True)`, binds `w<rank>.sock`, and serves it from a daemon thread in a `while True` loop (lines 116-145). The socket is never closed or unlinked and the directory is never removed. The stale-file check at lines 106-110 only handles a file at the same path, which never exists because the dir name changes per PGID.
- `kvcached/integration/vllm/interfaces.py:90-93` starts the listener for every worker "regardless of TP size", and lines 65-67 start it in the same-process TP=1 case, so the dir exists at TP=1 even when no TP IPC ever happens.
- `shutdown_kvcached()` (`kvcached/integration/vllm/interfaces.py:96-106`) does not touch it.
- `kvctl delete` -> `cmd_delete` (`kvcached/cli/kvctl.py:311-318`) -> `delete_kv_cache_segment` (`kvcached/cli/utils.py:161-185`) unlinks the POSIX shm name and `/dev/shm/<name>` only. `kvctl list` (`_detect_kvcache_ipc_names`, `kvcached/cli/kvtop.py:13-27`) scans `/dev/shm` only, so the dirs are invisible to the CLI.

Repro:

1. `ENABLE_KVCACHED=true KVCACHED_AUTOPATCH=1 vllm serve Qwen/Qwen2.5-0.5B-Instruct` (TP=1). Log shows `Worker 0 IPC listener started at /tmp/kvcached-tp-kvcached_vLLM_<PGID>-<hash>/w0.sock` (`tp_ipc_util.py:117`).
2. Stop the server (SIGTERM or SIGKILL, same result).
3. `ls -d /tmp/kvcached-tp-*`
4. `kvctl delete kvcached_vLLM_<PGID>`, then `ls -d /tmp/kvcached-tp-*` again.

Observed after one day of launches on a single pod (`ls -d /tmp/kvcached-tp-*`; each dir contains `w0.sock`):

```
/tmp/kvcached-tp-kvc_m1-c77203ea
/tmp/kvcached-tp-kvc_m2-11edb5e5
/tmp/kvcached-tp-kvcached_vLLM_28422-6c1fd57b
/tmp/kvcached-tp-kvcached_vLLM_29279-4bd3d051
/tmp/kvcached-tp-kvcached_vLLM_35279-ca7a40ef
/tmp/kvcached-tp-kvcached_vLLM_35872-daf4f32d
/tmp/kvcached-tp-kvcached_vLLM_36657-611fb843
/tmp/kvcached-tp-kvcached_vLLM_37278-6253d9bf
```

Only the `kvc_m1`/`kvc_m2` pair (explicit `KVCACHED_IPC_NAME`) was running at that point; the six `kvcached_vLLM_*` dirs belong to launches from earlier in the day whose processes had been stopped. The suffixes are exactly `uuid5(NAMESPACE_DNS, <ipc_name>).hex[:8]` from `tp_ipc_util.py:25` (recomputed for all eight). `kvctl delete` on the leftover `/dev/shm/kvcached_vLLM_*` segments removes the `/dev/shm` entries (all that `delete_kv_cache_segment` touches) and leaves every dir in place.

Expected: the socket is closed and unlinked and the dir removed when the owning process shuts down (`shutdown_kvcached()` or an exit hook next to `start_worker_listener_thread`), and `kvctl delete <ipc>` also removes `/tmp/kvcached-tp-<ipc>-<uuid5[:8]>`, since the path is derivable from the ipc name.

Small and self-contained; happy to send a PR (unlink on shutdown + `kvctl delete` cleanup) if that shape works for you.

rishabhsinha17:
## Summary

Every engine launch creates `/tmp/kvcached-tp-<ipc>-<hash>/w<rank>.sock`, at TP=1 too, and nothing removed it: the listener thread served the socket until the process died, and `kvctl delete` only knew about `/dev/shm`. One directory accumulates per launch (listing in #476).

## Root cause

`start_worker_listener_thread()` binds the socket in a daemon thread with no stop path, `shutdown_kvcached()` never touched it, and `kvctl delete` (`delete_kv_cache_segment()`) unlinks only the shm segment.

## Changes

- `kvcached/tp_ipc_util.py`: track each listener. `stop_worker_listener_threads()` wakes the accept loop, closes and unlinks the socket, and removes the directory (and the `pp<N>` subdirectory) once no other worker's socket is left in it. Also registered with `atexit` for clean exits.
- `kvcached/integration/vllm/patches.py`: `GPUWorkerPatch` also wraps `Worker.shutdown()`, which vLLM runs on every exit path that reaches `EngineCore.shutdown()`: in-process via `UniProcExecutor.shutdown()` at TP=1, and in each `WorkerProc` from `worker_main()`'s `finally` block. Releases without `Worker.shutdown` are left alone.
- `shutdown_kvcached()` (vLLM and SGLang) stops the listeners too.
- `kvcached/utils.py`: the directory naming rule moves next to `DEFAULT_IPC_NAME` as `get_tp_socket_dir(ipc_name)` so `kvctl` can derive it without loading the compiled extension. `kvctl delete <ipc>` now removes the directory as well and says so.

## Validation

- `tests/test_tp_socket_cleanup.py` (new, in the CPU manifest, 14 tests): stop unlinks the socket and removes the dir and the pp subdir, keeps the dir while another rank's socket is still in it, restart replaces the previous listener, the `Worker.shutdown` patch runs after vLLM's own teardown (also when that raises), is inert when kvcached is disabled and skipped when `Worker` has no `shutdown`; `delete_tp_socket_dir` by name and by `/dev/shm/<name>` path; `kvctl delete` output.
- `tools/run_cpu_tests.sh`: 190 passed locally (the 10 errors are the pre-existing macOS ones). ruff, isort, mypy 3.10 clean (same 13 pre-existing mypy notes as main).
- Fork CI: CPU Tests 3.9-3.13 https://github.com/rishabhsinha17/kvcached/actions/runs/33645641599, Pre-commit/MyPy https://github.com/rishabhsinha17/kvcached/actions/runs/33645641540
- Not yet run end to end on a GPU (`vllm serve`, SIGTERM, `ls /tmp/kvcached-tp-*`). The hooked path is the one vLLM 0.24 takes on SIGTERM (`run_engine_core` finally -> `EngineCore.shutdown` -> executor -> `Worker.shutdown`); I will add the on-pod check when I am back on the L40S.

Found while testing kvcached for vllm-project/aibrix#2419. The shm segment half of the same shutdown gap is #477.

Fixes #476

### DIFF
--- a/kvcached/tp_ipc_util.py
+++ b/kvcached/tp_ipc_util.py
@@ -2,34 +2,21 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import asyncio
+import atexit
 import os
 import pickle
 import socket
 import threading
-import uuid
-from typing import Any, Dict, cast
+from typing import Any, Dict, Optional, Tuple, cast
 
-from kvcached.utils import DEFAULT_IPC_NAME
+from kvcached.utils import get_tp_socket_dir
 from kvcached.vmm_ops import kv_tensors_created, map_to_kv_tensors, unmap_from_kv_tensors
 
-
-def _get_socket_dir_name() -> str:
-    """
-    Build a human-readable, IPC-name-based directory with a short hash suffix.
-
-    This keeps the original text-based IPC name visible while adding a hash
-    for extra uniqueness. The hash is deterministic so all workers in the same
-    engine instance agree on the directory.
-    """
-    # Deterministic short hash derived from the base name
-    suffix = uuid.uuid5(uuid.NAMESPACE_DNS, DEFAULT_IPC_NAME).hex[:8]
-    return f"kvcached-tp-{DEFAULT_IPC_NAME}-{suffix}"
-
-
-# Socket directory for tensor parallel (TP) worker communication.
-# Unix domain socket paths are limited to 108 characters on Linux, so we keep
-# the directory name short and validate the final socket path length below.
-SOCKET_DIR = os.path.join("/tmp", _get_socket_dir_name())
+# Socket directory for tensor parallel (TP) worker communication:
+# /tmp/kvcached-tp-<ipc_name>-<hash>. Unix domain socket paths are limited to
+# 108 characters on Linux, so the name is kept short and the final socket path
+# length is validated below.
+SOCKET_DIR = get_tp_socket_dir()
 
 
 def get_worker_socket_path(rank: int, pp_rank: int = 0) -> str:
@@ -93,13 +80,90 @@ def recv_msg(sock: socket.socket) -> Message:
     return cast(Message, pickle.loads(data))
 
 
+class _WorkerListener:
+    """One worker's IPC listener: its bound socket, the directory holding it,
+    and the thread serving it."""
+
+    def __init__(self, rank: int, pp_rank: int, root_dir: str, socket_dir: str,
+                 socket_path: str, server_sock: socket.socket) -> None:
+        self.rank = rank
+        self.pp_rank = pp_rank
+        self.root_dir = root_dir
+        self.socket_dir = socket_dir
+        self.socket_path = socket_path
+        self.server_sock = server_sock
+        self.stop_event = threading.Event()
+        self.thread: Optional[threading.Thread] = None
+
+    def stop(self) -> None:
+        """Stop serving, unlink the socket, and remove the directory once no
+        other worker's socket is left in it."""
+        self.stop_event.set()
+        # accept() only returns on a connection, so make one to let the loop
+        # observe stop_event. If that fails the daemon thread simply dies
+        # with the process; the socket file is unlinked either way.
+        try:
+            with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as wake:
+                wake.settimeout(1.0)
+                wake.connect(self.socket_path)
+        except OSError:
+            pass
+        if self.thread is not None:
+            self.thread.join(timeout=1.0)
+        self.server_sock.close()
+        try:
+            os.unlink(self.socket_path)
+        except FileNotFoundError:
+            pass
+        _remove_dir_if_empty(self.socket_dir)
+        if self.socket_dir != self.root_dir:
+            _remove_dir_if_empty(self.root_dir)
+        print(f"Worker {self.rank} IPC listener stopped, removed {self.socket_path}")
+
+
+def _remove_dir_if_empty(path: str) -> None:
+    try:
+        os.rmdir(path)
+    except OSError:
+        # Still holds another worker's socket, or already gone.
+        pass
+
+
+_listeners: Dict[Tuple[int, int], _WorkerListener] = {}
+_listeners_lock = threading.Lock()
+_atexit_registered = False
+
+
+def stop_worker_listener_threads() -> None:
+    """Stop every worker IPC listener started in this process: close its
+    socket, unlink it, and remove the per-instance socket directory
+    (issue #476). Safe to call repeatedly and when nothing was started.
+    """
+    with _listeners_lock:
+        listeners = list(_listeners.values())
+        _listeners.clear()
+    for listener in listeners:
+        listener.stop()
+
+
 def start_worker_listener_thread(rank: int, pp_rank: int = 0):
     """
     Start a thread that listens for messages on the worker socket.
     pp_rank is used to create a PP-stage-specific subdirectory so that
     concurrent SGLang PP stages do not bind the same socket path.
+
+    The listener is registered so that stop_worker_listener_threads() (called
+    from the integrations' shutdown paths and at interpreter exit) can unlink
+    the socket and remove the directory again.
     """
-    socket_dir = os.path.join(SOCKET_DIR, f"pp{pp_rank}") if pp_rank > 0 else SOCKET_DIR
+    global _atexit_registered
+    with _listeners_lock:
+        previous = _listeners.pop((rank, pp_rank), None)
+    if previous is not None:
+        previous.stop()
+
+    root_dir = SOCKET_DIR
+    socket_dir = os.path.join(root_dir, f"pp{pp_rank}") if pp_rank > 0 else root_dir
     os.makedirs(socket_dir, exist_ok=True)
     socket_path = get_worker_socket_path(rank, pp_rank)
 
@@ -112,11 +176,19 @@ def start_worker_listener_thread(rank: int, pp_rank: int = 0):
     server_sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
     server_sock.bind(socket_path)
     server_sock.listen()
+    listener = _WorkerListener(rank, pp_rank, root_dir, socket_dir, socket_path,
+                               server_sock)
 
     def listen_loop():
         print(f"Worker {rank} IPC listener started at {socket_path}")
         while True:
-            conn, _ = server_sock.accept()
+            try:
+                conn, _ = server_sock.accept()
+            except OSError:
+                break  # socket closed by stop()
+            if listener.stop_event.is_set():
+                conn.close()
+                break
             try:
                 msg: Message = recv_msg(conn)
                 # print(f"Worker {rank} received message: {msg}")
@@ -142,7 +214,13 @@ def start_worker_listener_thread(rank: int, pp_rank: int = 0):
                 conn.close()
 
     t = threading.Thread(target=listen_loop, daemon=True)
+    listener.thread = t
     t.start()
+    with _listeners_lock:
+        _listeners[(rank, pp_rank)] = listener
+        if not _atexit_registered:
+            atexit.register(stop_worker_listener_threads)
+            _atexit_registered = True
 
 
 # How long one worker-IPC exchange may take before it is treated as a failure.
--- a/kvcached/cli/kvctl.py
+++ b/kvcached/cli/kvctl.py
@@ -74,7 +74,7 @@ Available commands:
   kvtop [ipc ...] [--refresh r]  Launch curses kvtop UI (q to quit)
   !<shell cmd>                 Run command in system shell
   help                         Show this help message
-  delete <ipc>                 Delete IPC segment and its limit entry
+  delete <ipc>                 Delete IPC segment and its worker socket dir
   exit | quit                  Exit the shell
 """
 
@@ -309,11 +309,15 @@ def cmd_top(ipcs: Optional[List[str]] = None, refresh: float = 1.0):
 
 
 def cmd_delete(ipc: str):
-    from kvcached.cli.utils import delete_kv_cache_segment
+    from kvcached.cli.utils import delete_kv_cache_segment, delete_tp_socket_dir
 
-    if delete_kv_cache_segment(ipc):
+    removed_segment = delete_kv_cache_segment(ipc)
+    removed_socket_dir = delete_tp_socket_dir(ipc)
+    if removed_segment:
         print(_clr(f"Deleted IPC '{ipc}'.", 'green'))
-    else:
+    if removed_socket_dir:
+        print(_clr(f"Removed worker socket dir for IPC '{ipc}'.", 'green'))
+    if not removed_segment and not removed_socket_dir:
         print(_clr(f"IPC '{ipc}' not found.", 'red', bold=True),
               file=sys.stderr)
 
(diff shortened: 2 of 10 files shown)


kvcached/tp_ipc_util.py
```diff
@@ -93,13 +80,90 @@ def recv_msg(sock: socket.socket) -> Message:
...
+        # observe stop_event. If that fails the daemon thread simply dies
+        # with the process; the socket file is unlinked either way.
+        try:
+            with socket.socket(socket.AF_UNIX, socket.SOCK_STREAM) as wake:
+                wake.settimeout(1.0)
+                wake.connect(self.socket_path)
+        except OSError:
+            pass
+        if self.thread is not None:
+            self.thread.join(timeout=1.0)
```
shipiyouniao: The timed join can return while the thread is blocked on an already accepted connection. Closing server_sock only closes the listening socket; it does not close conn or prevent dispatch after recv_msg completes. I reproduced map execution after stop returned by sending the body at that point. Please make successful stop mean that no handler can still access the allocator, and retain enough listener state to handle a failed/incomplete stop safely.

shipiyouniao (changes requested):
The 14 socket-cleanup tests pass on 50090d0, but I reproduced a shutdown race with an accepted connection on Linux:

1. Start the real listener and send a valid message-length header, withholding the body.
2. Wait until the listener enters recv_msg(), then call stop_worker_listener_threads().
3. stop returns after 1.000 s while the listener is still alive.
4. Send the remaining map request body. The listener executes map_to_kv_tensors() and replies with status=success after stop has returned.

I used stubbed VMM functions for this test; this demonstrates the ordering problem, not a reproduced CUDA crash. Both integration shutdown paths call _shutdown_kvcached_impl() after stop_worker_listener_threads(), so the late handler can race allocator teardown.

Please cover both an accepted connection stalled in recv_msg() and a handler already executing a backend operation. Shutdown needs to cancel pending reads/prevent late dispatch and drain active operations before releasing their resources; a timed join followed by unconditional teardown is not sufficient.

rishabhsinha17:
Thanks for the repro. I turned it into a regression test first; on 50090d0 it fails with "map_to_kv_tensors dispatched after stop returned", and a companion test that parks a handler inside a stubbed VMM call fails with "stop returned while a handler was still executing".

stop() no longer relies on the timed join. The listener tracks accepted connections, and stop() sets the stop flag under the same lock, then shuts each tracked connection down, which wakes a parked recv_msg() with a connection error. Its loop drops a request that finishes arriving after stop has begun and closes later-accepted connections without reading them. Teardown (unlink, rmdir, close) happens only once the thread has exited, so after a successful stop no handler is left in a VMM call. If the thread has not finished one within 5 s, stop tears nothing down, keeps the listener registered for a retry by a later call or the atexit hook, and stop_worker_listener_threads() returns False. Cleanup errors no longer escape the shutdown paths.

New tests pin the stalled read from your recipe, a handler that is already inside a backend operation when stop begins, and the retained listener retry after an incomplete stop. They fail on the previous head and pass on 300bb50 across ten consecutive local runs. I rebased the branch onto current main, resolving against #413 and #373 in tp_ipc_util.py; at the CI pins the local cpu manifest and mypy results are identical to main's, with lint clean.

rishabhsinha17:
CI caught a mypy error in the stalled-read test's stub on 300bb50, an Event.set() inside an or expression. Amended and repushed, so the head is now 7d32e5e. Nothing else changed, and the torch-free mypy run at the CI pin is clean locally on 3.9, 3.11 and 3.13.

kvcached/tp_ipc_util.py
shipiyouniao: Thanks for adding the repro tests! I reran all 17 on 7d32e5e and they pass. Cancelling the accepted reads and keeping a busy listener around addresses the listener-side part of this.

There's still a caller-side gap: both [vLLM shutdown](https://github.com/ovg-project/kvcached/blob/7d32e5e6bb268b8b599871c4e3b0825960240422/kvcached/integration/vllm/interfaces.py#L115-L116) and [SGLang shutdown](https://github.com/ovg-project/kvcached/blob/7d32e5e6bb268b8b599871c4e3b0825960240422/kvcached/integration/sglang/interfaces.py#L78-L79) ignore the False result and call _shutdown_kvcached_impl() anyway.

I checked this with the real listener and each integration's shutdown function, using a parked VMM stub. In both cases, after the default 5-second timeout:
- stop returned False and the listener was still registered/alive;
- allocator teardown was called while the backend operation was still running;
- _kvcached_initialized was reset to False.

Could you carry the incomplete-stop result through these callers, keeping allocator/initialization state intact until draining succeeds, and add the timeout test at the integration boundary? The [same-rank replacement path](https://github.com/ovg-project/kvcached/blob/7d32e5e6bb268b8b599871c4e3b0825960240422/kvcached/tp_ipc_util.py#L258-L262) also drops the old registry entry and ignores previous.stop(), so it needs the same check before replacing the listener. This is the remaining part of the original shutdown-order issue; the stalled-read fix itself looks good.

shipiyouniao: I added 26a4d68 on top of your commits to finish the caller-side fix. Both integrations now keep the allocator and initialization state intact when draining times out, and same-rank restart refuses to replace a live listener. Cleanup errors and concurrent stop calls also retain the state needed for a retry.

Before pushing, the full CPU suite passed on Python 3.9-3.13 (241 tests each), along with MyPy, pre-commit and the C++ tests. I also ran four T4 checks across both integrations and both layouts: pause after a real CUDA map, attempt shutdown, verify the retained pages are still readable/writable, then drain, clean up and reinitialize. All four passed without CUDA errors or container restarts. This covers single-GPU lifecycle handling, not full TP/MPS serving.

All 11 GitHub checks are green too. Could you take a look at the follow-up when convenient? I've left the PR open for your review.

shipiyouniao: Rebased onto main after #418 merged; the head is now 4ddbf76. Your two commits and authorship are preserved, with the compatibility fixes folded into my follow-up. The listener keeps #418's transactional map results, typed errors and unmap phase information. The drain tests now intercept the transactional entry point, with added checks for error replies and disconnected peers.

Before pushing, the full CPU suite passed on Python 3.9-3.13 (286 tests each), plus MyPy, pre-commit and C++ tests. T4 validation passed 95 native tests and eight shutdown-overlap checks across both integrations and both layouts: four paused after real map, and four between unmap prepare and commit. Deferred teardown, drain/retry and reinitialization all passed, with no CUDA errors or container restarts. This is single-GPU lifecycle coverage, not full TP/MPS serving validation.

All 11 GitHub checks are green and the conflict is gone. Could you take another look at the rebased follow-up when you have time?

rishabhsinha17:
Went through 26a4d68 and the rebased head. The caller-side gap was real in 7d32e5e. Both integrations ignored the False from stop_worker_listener_threads(), so a timed-out drain still ran _shutdown_kvcached_impl() against a live handler and reset _kvcached_initialized, which also made a retry impossible. Deferring teardown with the state intact closes it. Your registry rework also fixes two holes our version left, a cleanup exception dropped the entry with nothing left for the atexit retry to find, and a concurrent stop could see the cleared dict and report success while the first call was still draining. The current-thread return in stop() corrects another edge we got wrong, and the restart guard is what you asked for on the replacement path.

I compared both replayed commits line by line against the originals. Commit 2 has one difference, the OSError guard now wrapping the typed response, which is the right resolution. Commit 1 is identical, including a removal of import uuid that was correct before #418 and stale after it. At 4dd7e71 and a2169fe the transactional unmap broadcast would raise NameError. Your commit restores the import, and the squash merge keeps the intermediates out of main.

Locally at 4ddbf76 on macOS py3.11, test_tp_socket_cleanup.py passes 26/26 and the cpu manifest gives 275 passed with the same 1 failure and 10 errors as c48be96 on this machine, all environmental (socket path length under the macOS tmp root, /dev/shm). ruff 0.11.7, isort 6.0.1 and codespell are clean, and CI-mode mypy 1.11.1 is clean on 3.9, 3.11 and 3.13 with output identical to main's plus the new file.

One non-blocking note. test_tp_ipc_worker_failures.py reloads tp_ipc_util per test and leaves six fake listeners parked in a stubbed accept, so the atexit hook now spends 6x5s draining them at interpreter exit, about 30s per CI job that pytest's own clock never shows. Happy to push a small teardown for that here or in a follow-up, whichever you prefer.

Ready from our side. Thanks for finishing the caller side and for the careful rebase.

shipiyouniao (dismissed):

shipiyouniao (commented):
I revisited my approval and want to narrow it: the cleanup is useful, and the previous T4 timeout/drain/retry tests support the teardown fix. They don't measure the new connection-tracking overhead, though. I've withdrawn the unconditional approval while we close that gap and rebase the current conflicts.

The global listener registry lock is mainly on lifecycle operations. The additional per-listener connection lock is taken for accepted IPC connections, so the relevant check is a matched IPC/serving comparison, not just shutdown tests. Could you include that comparison on the rebased head, covering the small frequent requests as well as map/unmap? I'm not claiming a measured slowdown or a new CUDA race here.

Please also refresh the PR's validation section: it still describes the earlier test count and pending GPU work, although later comments contain the real teardown fault results. One concise current summary would make the scope much easier to review.

rishabhsinha17:
measured the tracking overhead, about 1 to 2us per accepted connection, a 2 to 6 percent ceiling on a stub that strips the VMM work. the client mirrors _send_and_receive_message, one fresh connection per message, kv_tensors_created, 16-offset map, and the prepare/commit unmap pair, 9 interleaved windows per arm on an M1 Pro, per-op medians vs main. an arm with track_conn/untrack_conn patched out recovers most of the delta, and the registry lock stays on start/stop only. rebased and pushed, full numbers in the refreshed validation section, happy to rerun with a different request mix if you want.

shipiyouniao (approved):
LGTM at `4cf1090`. The missing Linux serving comparison is now covered, alongside the earlier eight real-T4 VMM shutdown/retry cases on this head.

I ran A-B-B-A with Qwen2.5-0.5B-Instruct, FP16, vLLM 0.28 V1, TP=1 with a separate worker process, eager/synchronous scheduling and identical warmup/request sequences. Both arms use the same current-main/public-compatibility baseline and native extension; the difference is this PR.

| Metric | Baseline | This PR |
| --- | ---: | ---: |
| Output throughput, total tokens / total measured time | 224.48 tok/s | 224.09 tok/s |
| Mean TTFT | 3325.01 ms | 3350.84 ms |
| Mean TPOT | 43.97 ms | 43.92 ms |
| Complete requests | 256/256 | 256/256 |

Each round exercised 64 map and 56 transactional-unmap IPC calls. All 65,536 measured output tokens arrived. The -0.17% aggregate throughput difference is within the observed baseline variation; this does not establish a universal overhead bound or multi-GPU qualification.

Both arms still emit identical FTensor cleanup errors at interpreter teardown. Those are recorded separately, not counted as clean-shutdown success. The candidate does stop and remove its worker listener as intended.

@RixinLiu, the listener cleanup change is ready for your final look from my side.

RixinLiu:
The potential overhead comes from tracking every accepted IPC connection: the listener acquires a lock to add the connection to a set, then acquires it again to remove the connection after handling the request.

I tested this on an A100 40GB with Qwen2.5-7B BF16, vLLM 0.24, and background preallocation enabled using 32 concurrent requests, each with 4,096 input and 128 output tokens, totaling 2,560 measured requests. Every wave exercised 27 map and 23 unmap IPC calls, with physical KV usage growing from 2.41 to 8.31 GiB and returning to 2.41 GiB. Throughput was 372.40 tok/s baseline versus 372.47 tok/s with this PR (+0.019%). No meaningful throughput regression was observed in this workload.

RixinLiu (approved):

PR #481, opened 2026-09-02
fix: ship kvcached_autopatch.pth in editable installs and warn when it is missing

### LINKED ISSUE #470: Editable install does not install kvcached_autopatch.pth; KVCACHED_AUTOPATCH silently no-ops

rishabhsinha17:
## Summary

`pip install -e .` never installed `kvcached_autopatch.pth`, so on editable installs `ENABLE_KVCACHED` / `KVCACHED_AUTOPATCH` did nothing and vLLM/SGLang ran vanilla with no error. The editable wheel now carries the `.pth`, and `import kvcached` warns when the knobs are set but the `.pth` is not in a site directory.

## Root cause

With `build-backend = "setuptools.build_meta"` and setuptools>=64, `pip install -e .` goes through the PEP 660 `build_editable` hook. `DevelopWithPth` in setup.py hooks the legacy `develop` command, which that path never runs. The `.pth` is the only thing that registers the engine import hooks, so nothing in kvcached executed at all (#470).

## Changes

- `setup.py`: `EditableWheelWithPth` runs setuptools' `editable_wheel` and appends `kvcached_autopatch.pth` (with a RECORD entry) to the root of the built wheel. Root-level wheel files land in site-packages, so pip installs and uninstalls it like any other file. It fails loudly if no wheel is found instead of silently skipping.
- `kvcached/__init__.py`: with `ENABLE_KVCACHED` or `KVCACHED_AUTOPATCH` set and no `kvcached_autopatch.pth` in `site.getsitepackages()` / `site.getusersitepackages()`, emit a `RuntimeWarning` that names the fix. This only helps processes that import kvcached (a vanilla engine never does), which is why the packaging change is the real fix.
- README: the manual `python tools/dev_copy_pth.py` step is no longer needed; the tool stays for `--check` / `--remove` and as a fallback.

`data_files` was tried first and cannot work: setuptools installs `data_files` under `<dist>.data/data`, which pip places relative to `sys.prefix`, never in site-packages.

## Validation

Scratch venv on this host (Python 3.11.2, pip 26.2.1, setuptools 84.0.0). The repo was copied from main with one deviation: the `get_extensions()` call in setup.py replaced by `[], {}`, since there is no CUDA toolchain here; the `.pth` handling does not depend on the extension.

- main: `pip install -e . --no-build-isolation` gives `__editable__.kvcached-0.1.5.pth` but no `kvcached_autopatch.pth` in site-packages. With a stub `vllm` package and `ENABLE_KVCACHED=true`, `import vllm` succeeds and `kvcached.integration.vllm.autopatch` is never imported.
- main + `data_files=[("", ["kvcached_autopatch.pth"])]`: the `.pth` is nowhere under the venv.
- this branch, same command: `kvcached_autopatch.pth` is in site-packages, `RECORD` has `kvcached_autopatch.pth,sha256=McWKfjYxcxl6VZKcXni_mACnLrjLywEJJczPJZdNVZU,492`, bytes identical to the repo copy. `import vllm` now runs the hook (`kvcached.integration.vllm.autopatch` imported, patch manager logs appear). `pip uninstall kvcached` removes the `.pth`.
- `.pth` removed by hand, `ENABLE_KVCACHED=true`: `import kvcached` prints the RuntimeWarning; copied back: `python -W error -c "import kvcached"` is silent.
- `tests/test_autopatch_pth.py` (new, CPU manifest, 9 tests): wheel injection (entries preserved, RECORD line, no temp file left, wheel without RECORD rejected), `pip install --target` of the rewritten wheel places the `.pth` at the install root, and the warning (either knob; quiet with the `.pth` in site or user site or with the knobs unset; still warns when `site.getsitepackages` raises).
- `tools/run_cpu_tests.sh`: 185 passed locally (10 pre-existing macOS errors). ruff, isort, mypy 3.10 clean.
- Fork CI: CPU Tests 3.9-3.13 https://github.com/rishabhsinha17/kvcached/actions/runs/33646859752, Pre-commit/MyPy https://github.com/rishabhsinha17/kvcached/actions/runs/33646859624

Not run: `pip install -e .` with the real CUDA extension build. The editable_wheel step runs after the build and only touches the produced wheel, so I do not expect a difference; I will confirm on the L40S when I am back on it.

Found while producing the MLA rows for #425 and testing kvcached for vllm-project/aibrix#2419.

Fixes #470

### DIFF
--- a/kvcached/__init__.py
+++ b/kvcached/__init__.py
@@ -1,6 +1,10 @@
 # SPDX-FileCopyrightText: Copyright contributors to the kvcached project
 # SPDX-License-Identifier: Apache-2.0
 
+import os
+import site
+import warnings
+
 # Version information
 from importlib.metadata import version
 
@@ -20,3 +24,47 @@ except ImportError as e:
             "  pip install torch>=2.6.0")
     else:
         raise
+
+AUTOPATCH_PTH = "kvcached_autopatch.pth"
+
+
+def _autopatch_requested() -> bool:
+    return any(
+        os.getenv(name, "false").lower() in ("true", "1")
+        for name in ("ENABLE_KVCACHED", "KVCACHED_AUTOPATCH"))
+
+
+def _autopatch_pth_installed() -> bool:
+    """True if the .pth that registers the engine import hooks is in a
+    directory whose .pth files the interpreter executes at startup."""
+    site_dirs = []
+    try:
+        site_dirs.extend(site.getsitepackages())
+    except Exception:
+        pass
+    try:
+        site_dirs.append(site.getusersitepackages())
+    except Exception:
+        pass
+    return any(os.path.isfile(os.path.join(d, AUTOPATCH_PTH)) for d in site_dirs)
+
+
+def _warn_if_autopatch_pth_missing() -> None:
+    """The engines are patched by import hooks that kvcached_autopatch.pth
+    registers at interpreter startup. Without it, ENABLE_KVCACHED and
+    KVCACHED_AUTOPATCH do nothing and vLLM/SGLang run vanilla without any
+    error (issue #470). Say so loudly whenever kvcached is imported with the
+    knobs set but the .pth absent."""
+    if not _autopatch_requested() or _autopatch_pth_installed():
+        return
+    warnings.warn(
+        "ENABLE_KVCACHED/KVCACHED_AUTOPATCH is set but kvcached_autopatch.pth "
+        "is not installed in site-packages, so vLLM/SGLang will not be patched "
+        "and will run without kvcached. Reinstall kvcached (pip install -e . or "
+        "a wheel) or run `python tools/dev_copy_pth.py` from a kvcached checkout.",
+        RuntimeWarning,
+        stacklevel=2,
+    )
+
+
+_warn_if_autopatch_pth_missing()
--- a/README.md
+++ b/README.md
@@ -144,8 +144,12 @@ pip install kvcached --no-build-isolation
 # under the project root folder
 
 pip install -e . --no-build-isolation --no-cache-dir
-python tools/dev_copy_pth.py
 ```
+
+Both install `kvcached_autopatch.pth` into site-packages. That file registers
+the vLLM/SGLang import hooks; without it the engines run without kvcached and
+`import kvcached` warns. `python tools/dev_copy_pth.py --check` shows whether
+it is present and `python tools/dev_copy_pth.py` copies it by hand if needed.
 
 ### Using Docker
 
--- a/setup.py
+++ b/setup.py
@@ -1,15 +1,19 @@
 # SPDX-FileCopyrightText: Copyright contributors to the kvcached project
 # SPDX-License-Identifier: Apache-2.0
 
+import base64
+import hashlib
 import os
 import shutil
 import sys
+import zipfile
 from pathlib import Path
 from typing import List
 
 from setuptools import find_packages, setup
 from setuptools.command.build_py import build_py
 from setuptools.command.develop import develop
+from setuptools.command.editable_wheel import editable_wheel
 from setuptools.command.install import install
 
 try:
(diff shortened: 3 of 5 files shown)


kvcached/__init__.py
```diff
@@ -20,3 +24,47 @@
...
+def _autopatch_pth_installed() -> bool:
+    """True if the .pth that registers the engine import hooks is in a
+    directory whose .pth files the interpreter executes at startup."""
+    site_dirs = []
+    try:
+        site_dirs.extend(site.getsitepackages())
+    except Exception:
+        pass
+    try:
+        site_dirs.append(site.getusersitepackages())
```
shipiyouniao: Could we only include the user site when `site.ENABLE_USER_SITE` is true? `getusersitepackages()` still returns a path under `python -s` / `PYTHONNOUSERSITE=1` and in a venv that disables user packages. If a previous install left `kvcached_autopatch.pth` there but the active site lacks it, this returns true and suppresses the warning even though the hooks cannot be loaded from that directory.

I reproduced this with no active-site file, a file only in the user site, and `site.ENABLE_USER_SITE=False`: `_warn_if_autopatch_pth_missing()` emitted no warning. Please add a regression test alongside the enabled-user-site case.

shipiyouniao (changes requested):
The editable-wheel fix looks sensible, and all 9 tests in `test_autopatch_pth.py` passed in my Linux CPU run. I found one small issue in the new warning: a leftover file in a disabled user site suppresses it, even though that interpreter will not execute the file. The added regression check fails with `DID NOT WARN`.

Could you fix that check and cover disabled user-site behavior? I have not repeated the real CUDA-extension editable build; the packaging results here are CPU-only.

kvcached/__init__.py
rishabhsinha17: fixed in fc7d8a2. the user site now only counts when site.ENABLE_USER_SITE is true, so a leftover .pth in a disabled user site no longer suppresses the warning. added your repro as a regression test, file only in the user site with ENABLE_USER_SITE off, covering both the False and None values. it failed DID NOT WARN before the gate and passes after, 11/11 in test_autopatch_pth.py on my cpu run.

shipiyouniao (approved):
LGTM at `fc7d8a2`. The disabled-user-site finding is fixed, and I have now covered the real editable CUDA build rather than only the CPU packaging tests.

On a T4 with CUDA 12.9, Python 3.12 and PyTorch 2.13, the exact PR built its native extension and passed map/write/readback/unmap. The installed `.pth` and RECORD hash/size matched; fresh interpreters verified automatic patching, explicit disable and the missing-hook warning. Uninstall removed the hook and editable registration. The packaging tests passed 11/11.

I also repeated the build on current main plus the public 0.28 compatibility changes, with this PR's packaging hooks verified unchanged by AST comparison. Qwen2.5-0.5B-Instruct completed four requests and all 64 requested output tokens on vLLM 0.28 V1. The same 11 packaging tests passed there.

One separate limitation: model teardown still logs FTensor cleanup errors. The matched control without this PR has the same errors and counts, so this is not a packaging regression or a claim that GPU teardown is fully clean.

@RixinLiu, the editable-install change is ready for your final look from my side.

shipiyouniao (approved):
Still LGTM at `05b141c`. I merged current main into the branch without rewriting your commits. The only manual conflict was the CPU test manifest; both test entries are retained.

The exact candidate passed full pre-commit, GPU-free C++, and CPU/MyPy checks on Python 3.9-3.13 (336 CPU tests each). I also repeated the real T4 editable build/install, native map/write/readback/unmap, fresh-interpreter enable/disable, missing-hook fault detection, all 11 packaging tests, and uninstall cleanup. All 11 GitHub checks are green too.

@rishabhsinha17, please take a quick look at the merge resolution. @RixinLiu, no remaining findings from my side. This rerun covers packaging and native VMM, not a new model-performance measurement.

PR #482, opened 2026-09-02
fix(vllm): unlink the /dev/shm segment when EngineCore shuts down

### LINKED ISSUE #477: Graceful SIGTERM leaks the /dev/shm segment on vLLM: the only cleanup is the C++ MemInfoTracker destructor, which the forked EngineCore never runs

rishabhsinha17:
## Summary

A graceful SIGTERM to a kvcached-enabled vLLM server left `/dev/shm/kvcached_vLLM_<PGID>` (or the explicit `KVCACHED_IPC_NAME`) behind after every process had exited; `kvctl list` / `kvtop` kept showing the dead pools and PGID reuse produced `_1`, `_2` suffixes. `EngineCore.shutdown()` now ends in `shutdown_kvcached()`, which unlinks the segment while the process is still alive, and `MPClient.shutdown()` removes whatever segment an engine killed mid-teardown left behind.

## Root cause

On the vLLM path the segment belongs to the C++ `MemInfoTracker` inside `PageAllocator`, whose only cleanup is its destructor. vLLM's forked EngineCore leaves through `os._exit` after `run_engine_core()`'s `finally` block calls `EngineCore.shutdown()`, so no destructor runs. Nothing in the integration called `shutdown_kvcached()`, and that function only cleared the FTensor allocators anyway, not the `PageAllocator` held by the Python `KVCacheManager`. Full trace in #477.

GPU validation on vLLM 0.24.0 exposed a second layer: a server-level SIGTERM does not let `EngineCore.shutdown()` finish. `run_engine_core()` restores SIGTERM to SIG_DFL before calling it, and the process manager under `MPClient.shutdown()` sends the engine a second SIGTERM during that teardown, then a SIGKILL because the default `--shutdown-timeout` is 0. With SIG_DFL restored, the second signal ends the engine before the unlink runs. SIGTERM to the engine process alone, where no second signal follows, runs the unlink and removes the segment, so the engine-side fix is correct but unreachable on the server kill path. The client is still running after the engines are gone, which makes it the one place that can always clean up.

## Changes

- `kvcached/kv_cache_manager.py`: `KVCacheManager.shutdown()` stops the prealloc thread (so it cannot recreate the file) and unlinks `SHM_DIR/<ipc_name>`, exactly what `MemInfoTracker::cleanup()` does. Idempotent; the destructor's later unlink becomes a no-op.
- `kvcached/integration/vllm/interfaces.py`: `shutdown_kvcached()` shuts down every registered vLLM pool before the allocator; a failing pool does not stop the rest.
- `kvcached/integration/vllm/patches.py`: `EngineCorePatch` also wraps `EngineCore.shutdown` to call `shutdown_kvcached()` after vLLM's own teardown (also when that raises; never masks vLLM's result; inert when kvcached is disabled; a class without `shutdown` is left alone with a warning). `MPClientPatch` wraps `MPClient.shutdown` with the same contract to remove the segment once the original has stopped the engines, covering the kill path above.
- `kvcached/utils.py`: `unlink_default_ipc_segment()` unlinks `SHM_DIR/DEFAULT_IPC_NAME` and reports whether it removed a file. It lives in `utils` so the client process does not import torch or the compiled extension for it.

## Validation

- `tests/test_vllm_engine_core_shutdown.py` (new, CPU manifest, 18 tests): the `EngineCore.shutdown` wrapper (order, vLLM teardown raising, kvcached failure not masking vLLM's result, disabled, idempotent, class without `shutdown`), the `MPClient.shutdown` wrapper (same matrix), `shutdown_kvcached()` pool-then-allocator order and registry clearing with a failing pool, `KVCacheManager.shutdown()` (stops prealloc and unlinks, idempotent, missing segment, prealloc stop failing), and `unlink_default_ipc_segment()` (removes the segment, silent when already gone, warns instead of raising on OSError).
- ruff (pinned 0.11.7), isort and mypy clean on the changed files. The hosted CPU matrix runs on this PR.
- GPU end to end on an L40S (vllm 0.24.0, torch 2.11.0+cu130, Qwen2.5-0.5B-Instruct, `ENABLE_KVCACHED=true`), the check from #477: start `vllm serve`, confirm `/dev/shm/kvcached_vLLM_<PGID>` exists and the server returns a completion, `kill -TERM` the server process group, wait for every process to exit, inspect `/dev/shm`. On upstream main (efc074e) the segment remains, at the default shutdown timeout and at `--shutdown-timeout 10`. On this branch the segment is gone in both modes, with `Unlinked KV cache limit segment ... left by the engine process` in the API server log, and SIGTERM to the engine process alone exercises the engine-side path (`Unlinked KV cache limit segment ...` from `KVCacheManager.shutdown()`). No EngineCore survivors and GPU memory back to 0 MiB in every run. Command transcript in the PR comments.

#480 handles the worker socket directories on the same shutdown path; the two `shutdown_kvcached()` edits touch adjacent lines and I will rebase whichever lands second.

Found while testing kvcached for vllm-project/aibrix#2419.

Fixes #477


rishabhsinha17:
Ran the GPU E2E and it caught something. The red arm behaved, upstream main at efc074e leaks the segment after a graceful TERM, but at d0f2a34 this branch leaked it too. On the server kill path the engine is gone before the unlink runs. run_engine_core() restores SIGTERM to SIG_DFL before calling EngineCore.shutdown(), and the process manager under MPClient.shutdown() sends the engine a second SIGTERM during that teardown, then a SIGKILL because the default shutdown timeout is 0. With SIG_DFL restored, the second signal ends the engine mid-teardown. Sending SIGTERM to the engine process alone, where no second signal follows, produced the Unlinked line and an empty /dev/shm, which pinned the failure on the kill path rather than the unlink itself.

3854d38 adds an MPClientPatch for that path. Because the client is still running after the engines are gone, MPClient.shutdown() now removes whatever segment they left behind, and the engine-side unlink stays for engines nothing kills a second time. Rerun on the L40S with vllm 0.24.0, torch 2.11.0+cu130 and Qwen2.5-0.5B-Instruct, one fresh server per arm, TERM to the server process group, at the default shutdown timeout:

```
# arm = upstream main @ efc074e (red) | this branch @ 3854d38 (green)
ENABLE_KVCACHED=true vllm serve Qwen/Qwen2.5-0.5B-Instruct --port 8123 \
  --gpu-memory-utilization 0.30 --max-model-len 2048 --enforce-eager
ls /dev/shm | grep kvcached_        # while serving, after one completion
  red:   kvcached_vLLM_13452
  green: kvcached_vLLM_14486
kill -TERM -<PGID>; wait for exit    # pgrep EngineCore -> none, nvidia-smi -> 0 MiB, both arms
ls /dev/shm | grep kvcached_
  red:   kvcached_vLLM_13452        # the #477 leak
  green: (empty)
green API server log:
  [kvcached][INFO][utils.py:202] Unlinked KV cache limit segment /dev/shm/kvcached_vLLM_14486 left by the engine process
```

Same procedure at a 10s shutdown timeout, same outcome, red leaks kvcached_vLLM_15568 and green removes kvcached_vLLM_16555. TERM to the engine process alone exercises the engine-side path, Unlinked KV cache limit segment /dev/shm/kvcached_vLLM_11482 from KVCacheManager.shutdown(). No EngineCore survivors and 0 MiB on the GPU after every run. Unit coverage mirrors the EngineCore wrapper, 18 tests in the file now. Taking this out of draft.

kvcached/integration/vllm/patches.py
```diff
@@ -991,6 +993,111 @@ def _patched_engine_init(self, vllm_config, *args: Any, **kwargs: Any):
...
+
+        def _patched_client_shutdown(self, *args: Any, **kwargs: Any):
+            try:
+                return original_shutdown(self, *args, **kwargs)
+            finally:
+                if enable_kvcached():
+                    try:
+                        from kvcached.utils import unlink_default_ipc_segment
+
+                        unlink_default_ipc_segment()
```
shipiyouniao: [P2] Keep cleanup scoped to the client's shutdown generation

This runs even when the original `MPClient.shutdown()` is already a no-op. I reproduced: shut client A down, recreate its IPC pathname for a replacement engine, then call A.shutdown() again. The original shutdown runs only once, but this wrapper unlinks the replacement file on the second call. vLLM's finalizer guard protects its own teardown, not this `finally` block. That removes the live replacement's control/monitoring segment. Please track cleanup per client/generation and avoid unlinking a pathname now owned by another engine; add a repeated-shutdown/name-reuse regression test.

kvcached/kv_cache_manager.py
```diff
@@ -747,6 +753,31 @@ def observability_snapshot_dict(self, *, integration=None):
...
+
+        The C++ MemInfoTracker unlinks its /dev/shm segment only from its
+        destructor, which never runs when the owning process leaves through
+        os._exit (vLLM's forked EngineCore after SIGTERM, issue #477). Stop
+        the prealloc thread, then unlink the segment here, exactly what the
+        destructor would do. Safe to call more than once.
+        """
+        if self._shut_down:
+            return
+        self._shut_down = True
```
shipiyouniao: [P2] Do not mark a failed cleanup permanently complete

`_shut_down` is set before stopping the allocator and unlinking the file, while errors from both operations are swallowed. Injecting a one-shot `PermissionError` from `os.unlink()` leaves the segment behind; a second shutdown returns here without attempting the now-successful unlink. The stop-error path has the same retry limitation. Please distinguish successful cleanup from an attempted shutdown, keeping failed steps retryable without repeating already completed teardown or deleting a replacement segment. A failure-then-retry test reproduces this at the current head.

shipiyouniao (changes requested):
Thanks for adding the real shutdown checks. I ran the 18 existing tests at `3854d38`; all pass. Two additional cleanup regressions fail, though: an old client's repeated shutdown deletes a replacement segment with the same name, and a failed manager unlink cannot be retried because shutdown was already marked complete.

These are deterministic Linux tests using the actual patch methods and real temporary files, with the upstream client/allocator mocked. They are not reports of a reproduced GPU crash. Could you add coverage for these ownership/retry cases before merging? Details inline.

kvcached/integration/vllm/patches.py
RixinLiu: fixed in c2e1d6d

kvcached/kv_cache_manager.py
RixinLiu: fixed in c2e1d6d

RixinLiu (approved):

shipiyouniao (dismissed):
Thanks @RixinLiu, both of my earlier findings are fixed at `68b7180`. I reran the original regressions and added real-GPU shutdown checks; this looks good to approve.

The standalone PR head passed the 322-test CPU manifest and all 28 shutdown/regression tests. Remote CI is also green.

For the GPU checks, I used a T4 with Qwen2.5-0.5B-Instruct, FP16, TP=1, vLLM 0.28.0 V1, eager execution and synchronous scheduling. KVCached preallocation was enabled with 2 MiB pages. Since this environment runs 0.28, I applied the compatibility changes from #492 (`e84a699`), #493 (`3235b73`) and #500 (`511027b`) on top. The shutdown implementations were unchanged, verified by AST comparison against this PR head; the C++ source tree was identical too.

| Check | Result |
| --- | --- |
| Normal shutdown | The engine removed its shared-memory segment. |
| Restart with the same IPC name, then repeat the old client's shutdown | The replacement segment survived and the new engine continued generating successfully. |
| Inject one preallocation-thread stop failure and one unlink failure inside the running EngineCore | Both faults were hit. Failed attempts retained pool registration and initialization state; the third attempt completed cleanup. The successful thread-stop step ran only once. |
| SIGTERM to the server process group, shutdown timeout 0 and 10 seconds | Both cases passed. The parent client removed the segment, with no live child process remaining. |
| SIGTERM to EngineCore only | EngineCore removed the segment and the API server subsequently exited, with no live child process remaining. |

The accepted cases completed 24 requests and all 384 requested output tokens before the relevant shutdowns, including generation after the name-reuse check. No CUDA error, GPU OOM, illegal memory access or EngineCore startup failure occurred. No test segment or GPU compute process remained afterward.

The deliberate termination cases did log `EngineDeadError`; the process-group cases also logged the Python resource tracker's semaphore-cleanup warning. I am not counting these as silent exits or claiming that every kind of IPC resource is leak-free. This validates the shutdown/ownership/retry paths above, not TP/MPS behavior or performance.

kvcached/integration/vllm/patches.py
```diff
@@ -1068,6 +1070,120 @@ def _patched_engine_init(self, vllm_config, *args: Any, **kwargs: Any):
...
+                        cleanup = IPCSegmentCleanup(os.path.join(SHM_DIR, DEFAULT_IPC_NAME))
+                        self._kvcached_ipc_cleanup = cleanup
+                except Exception as e:
+                    logger.warning("Failed to capture client shutdown segment: %s", e)
+            try:
+                return original_shutdown(self, *args, **kwargs)
+            finally:
+                if cleanup is not None:
+                    try:
+                        cleanup.unlink()
```
shipiyouniao: [P1] Only the engine-owning client should unlink this segment. With `--api-server-count 2`, each frontend MPClient connects to engines managed by the supervisor, so `resources.engine_manager` is None. Its native shutdown does not stop EngineCore, but this unconditional fallback removes the shared segment anyway. I reproduced the segment disappearing while EngineCore and the other frontend were still alive on a T4. Please check ownership before capturing/using the cleanup object, and cover both non-owning clients and the legacy owned-process-handle path.

shipiyouniao (changes requested):
Thanks for resolving the manifest conflict. Your new head has the same Git tree as the merge candidate I just tested, so there is no need for me to push another merge.

I need to withdraw my approval, though: the additional multi-API-server check found an ownership case my earlier single-client tests missed. The previous shutdown tests still pass; this is a separate case.

On a T4, Qwen2.5-0.5B-Instruct, FP16, TP=1, vLLM 0.28 V1, eager execution and synchronous scheduling, I started `vllm serve` with `--api-server-count 2`, an explicit `KVCACHED_IPC_NAME`, and KVCached preallocation enabled with 2 MiB pages. As in my earlier report, the GPU run included the public compatibility changes from #492, #493 and #500; the shutdown implementation was unchanged.

Reproduction: serve four requests, then send SIGTERM to just one API-server child, not the server process group or EngineCore.

| Observation | Result |
| --- | --- |
| Requests before the signal | 4 completed, all 64 requested output tokens |
| Shared segment before the signal | Present; capacity 5,218,762,752 bytes |
| About 20 ms after the signal | Segment missing while EngineCore and the other API process were still alive |
| After the selected API exited | EngineCore, the other API and the supervisor were still alive; segment still missing |

The next HTTP request disconnected, but I am not using that alone as proof of a model failure during frontend shutdown. The directly observed defect is removal of the live engine's shared segment.

Could you restrict the client-side fallback to clients that actually own and shut down the engines, and add a non-owner regression? In current vLLM, externally managed clients have `resources.engine_manager is None`; their shutdown only closes the client resources. Older supported versions use `resources.core_engines` with process handles, so please preserve that ownership path too. A multi-API check should verify that one client's shutdown preserves the segment, and that final owner shutdown still cleans it up.

kvcached/integration/vllm/patches.py
rishabhsinha17: fixed in 8e9c164. only the client that launched the engines captures the cleanup, checked before the original shutdown runs, engine_manager on current vllm and per-engine proc_handle entries on resources.core_engines for the older versions. a client matching neither unlinks nothing, a leaked segment is recoverable with kvctl delete and a removed live one is not. regressions cover your frontend case (engine_manager None, segment stays), both legacy arms, the frontend-then-owner sequence ending in cleanup, and a retry that still runs after teardown clears the markers. shutdown file 32 passed, full cpu manifest 341, i cannot reproduce the multi-api serve here so the T4 check stays yours.

shipiyouniao:
Thanks, the non-owner fix in `8e9c164` works in the T4 reproduction: shutting down one of two API-server children now preserves the live segment, including its inode and capacity. The older-process-handle path and retry coverage look right too. I reran the Linux CPU manifest: 352 passed.

There is still a separate final-cleanup gap in the multi-API case:

| Check | Result |
| --- | --- |
| One frontend exits while EngineCore is alive | Segment preserved, as intended |
| Single API server, SIGTERM to the process group, shutdown timeout 0 | Segment removed |
| Two API servers, SIGTERM to the process group, shutdown timeout 0 | Supervisor exited with code 0 and no GPU compute processes remained, but the segment was still present 15 seconds later |

Both process-group checks completed four requests and all 64 requested output tokens before the signal. Same T4/Qwen2.5-0.5B-Instruct, FP16, TP=1, vLLM 0.28 V1, eager/synchronous setup as before. The GPU source includes the public compatibility changes from #492/#493/#500; the shutdown implementations were verified unchanged against this PR by AST comparison. One earlier multi-API attempt failed during CUDA driver initialization before serving and is excluded from the cleanup result above.

The missing owner appears to be the supervisor itself. In [run_multi_api_server()](https://github.com/vllm-project/vllm/blob/2cf0a6915ce544dc493a0990f2ea38d81601128a/vllm/entrypoints/cli/serve.py#L329), it creates `local_engine_manager` directly and later [calls its shutdown directly](https://github.com/vllm-project/vllm/blob/2cf0a6915ce544dc493a0990f2ea38d81601128a/vllm/entrypoints/cli/serve.py#L405). There is no owning MPClient in that process to execute the fallback, and EngineCore can still be killed before its own unlink.

Could you cover that actual owner shutdown boundary as well, with a two-frontend final-exit regression? Please keep the new non-owner guard: it fixes the live-segment deletion correctly. This remaining case is a stale-segment leak after exit, not a recurrence of that live-state deletion. If the supervisor path needs a separate follow-up, we should make that scope explicit rather than claim multi-API cleanup is complete.

rishabhsinha17:
the supervisor boundary is coverable from kvcached's side and belongs in this pr, fixed in c87b393. the autopatch fires in the supervisor too, vllm serve imports vllm there. a new patch wraps CoreEngineProcManager.shutdown, the call run_multi_api_server makes on exit, with the same capture-then-unlink as the client patch and no ownership check, the manager spawned the engines and the original shutdown joins or kills them before the unlink runs, timeout 0 included. segment naming holds there, same process group. your two-frontend final exit is the new regression, frontends leave the segment, the manager shutdown removes it. 8 new tests, shutdown file 40 passed, cpu manifest 349. headless ends in the same call and gets the same cleanup, on 0.10 plus, older versions keep the client coverage. i did not touch the non-owner guard logic beyond a docstring pointer, and the T4 rerun stays yours.

RixinLiu:
Took a look. @shipiyouniao you can merge this PR by yourself after you think it's good to merge

shipiyouniao:
@rishabhsinha17 I've added a follow-up for the shutdown failure case. Could you review e5ad559?

Both the owning client and the supervisor now only unlink after confirming that all owned engines exited. Live or uninspectable children keep their segment for a later cleanup attempt. A one-second group-wide reap covers vLLM returning immediately after SIGKILL. The original shutdown result/error is preserved.

Validation on this commit:
- Full pre-commit and C++ checks, CPU suites and mypy on Python 3.9–3.13: green, 374 CPU tests per version.
- T4, vLLM 0.28 V1, Qwen2.5-0.5B, TP=1: repeated shutdown/replacement protection; prealloc-stop and unlink failure/retry; injected terminate failure; single/two-API-server shutdown; non-owning frontend exit. This used the public 0.28 compatibility stack, with the changed shutdown code verified identical to this commit.
- Injected terminate failure: the real engine stayed alive, and its segment survived both the exception and a repeated no-op shutdown. After engine exit, retry removed the segment. This tests segment safety, not continued serving from an explicitly shut-down client.

One native ownership issue still needs attention before approval: `MemInfoTracker` unlinks by pathname in its destructor. A compiled reproduction using the real header confirms that destroying an old tracker after explicit cleanup deletes a replacement segment under the same name. That C++ code is unchanged from main, but early cleanup exposes the delayed-destruction case during same-process reuse. Could you handle the native ownership/disarming side? I kept it separate from this process-exit fix.

shipiyouniao:
Update: the additional lifecycle issues below are addressed in `de87d2c`, now pushed with regression tests and a validation plan. The private CI and scoped T4 checks passed. Author review and real two-GPU TP/PP validation are the remaining requested checks; see the follow-up comment for the results and limitations.

The confirmed failure at `3631eef` was a delayed first shutdown: the old engine had exited, a new engine had reused the IPC name, and the old manager then captured and deleted the new segment. Earlier replacement tests covered repeated shutdown, not this ordering. Further regressions covered failed identity capture, concurrent cleanup, and multiple pools sharing one segment. The fix retains ownership during startup instead of claiming a shutdown-time pathname.

shipiyouniao:
@rishabhsinha17 I've pushed `de87d2c` as an additive fix on your branch. Could you review it and run the two-GPU checks below? I kept your commits intact.

The key change is retaining segment identity during startup, while ownership can still be established, rather than looking it up on the first shutdown. Native pools sharing a segment now retain it until the last local pool stops. The tests cover delayed first shutdown, replacement generations, failed identity/unlink operations, concurrent cleanup, and startup cancellation. The lifecycle locks do not wrap allocation or mapping.

Validation on this exact candidate:

| Check | Result |
| --- | --- |
| Full pre-commit, GPU-free C++, CPU/MyPy on Python 3.9-3.13 | Passed; 447 CPU tests per interpreter |
| Native cleanup tests, including six real syscall faults | 18 passed |
| T4 real-model lifecycle, one/two API frontends | Eight cases passed, including cleanup retries and same-name restart |
| Headless supervisor with a native allocator child | Four cases passed; not a full headless model deployment |
| Two ranks sharing the T4, isolated experimental NCCL 2.30.7 | TP/PP normal cleanup and restart, TP rank SIGTERM, and PP stage SIGKILL passed; 32 requests / 512 output tokens across those four cases |

The model runs used Qwen2.5-0.5B-Instruct, vLLM 0.28 V1, FP16 and eager execution. The 0.28 compatibility overlay was checked against the candidate's lifecycle code. Same-device ranks are useful additional coverage, but are not a substitute for two physical GPUs. All 11 upstream CI checks also passed on the new head.

The [test plan](https://github.com/ovg-project/kvcached/blob/de87d2c2508384ae37c6e492c62a587706da2b75/docs/SHARED_SEGMENT_LIFECYCLE_TESTING.md#tppp-tier) is in the commit. Please run TP=2/PP=1 and TP=1/PP=2 separately on two physical GPUs with the normal NCCL/runtime configuration:

- Exercise normal shutdown and partial rank/stage failure during startup, serving and teardown, including a delayed last worker. Include one and two API frontends.
- Record the actual rank/stage-to-segment ownership. A frontend or stale owner must not remove a live/replacement segment; completed teardown must remove the original segment.
- Restart with the same IPC name and check complete generated outputs, process/GPU leftovers and CUDA/NCCL errors. Include both SIGTERM and SIGKILL fault paths.

The CPU lifecycle and native test files are included, but running pytest alone does not exercise those real TP/PP model scenarios. Combined TP=2/PP=2 would need four GPUs and is not claimed here. If your review and the applicable two-GPU checks are clean, I'm happy for this to proceed to merge.

One separate observation: PP stage SIGTERM did not automatically finish teardown within 60 seconds in the same-device experiment. Native vLLM with KVCached disabled showed the same timeout; explicit shutdown cleaned up both. That is not evidence of a regression introduced by this PR, so I would not block this fix on that observation alone. It is still worth reproducing on a standard two-GPU setup: if it persists in native vLLM there, we can investigate and report it upstream separately rather than expand this PR.

rishabhsinha17:
reviewed the full c87b393..de87d2c range, no defects found, and startup identity retention belongs in this pr, the 3631eef failure showed shutdown-time capture is unsound at every owner boundary. the confirmed-exit gate plus the retained-inode samestat check is the right pair, every failure path keeps the segment rather than guessing, and the lifecycle locks stay off the allocation path.

verified here, the push fast-forwards onto my checkout, 126 passed in the shutdown file, the manifest matches your 447 collected minus my usual macos env set, CI 11/11 confirmed. reverting patches.py and kv_cache_manager.py to c87b393 turns the delayed-first-shutdown, capture-failure-retry and no-handshake regressions red across all three owner boundaries. restored and green.

the native fault tier and the T4 and same-device runs stay your evidence.

i am lining up a two-GPU box and will run TP=2/PP=1 and TP=1/PP=2 from your plan, one frontend and two. SIGTERM and SIGKILL fault paths are in that pass, with the ownership mapping and the same-name restart.

main picked up #480 and now conflicts with the branch, i will push the merge resolution separately.

rishabhsinha17:
Merge of main pushed as 331ee5c with two judgment calls. In shutdown_kvcached the listener stop from #480 runs before the pool loop, and the pool failure path now returns False to match the bool contract. The vllm_modules fixture also clears leftover tp_ipc_util listeners the way socket_root does, because test_tp_ipc_pp_fanout and test_tp_ipc_worker_failures leave a wedged listener registered and the gate deferred the pool tests on it.

rishabhsinha17:
Ran the requested two-GPU matrix at this exact head 331ee5c. TP=2/PP=1 and TP=1/PP=2 both pass all eight scenarios. I recorded ownership at every phase, at no point did a non-owner or stale owner remove a live segment, completed teardown removed the original segment every time, and all 16 same-name restarts served complete outputs. After every case the audit found the case's segment absent, the serve process group exited and both GPUs at 0 MiB, and /dev/shm on the box sits empty after the full matrix.

Hardware was 2x NVIDIA L40S (SM 8.9, 46068 MiB each) on separate NUMA nodes with driver 580.178.04. Serving used vLLM 0.28.0 V1 with the V1 model runner pinned via VLLM_USE_V2_MODEL_RUNNER=0, async scheduling turned off by its serve flag, FP16, eager execution, and kvcached installed non-editable from 331ee5c. Before any case counted, each serve had to show the autopatch engaged, all three patched shutdown wrappers reporting the patch module, a live segment under the explicit IPC name, the per-worker socket dir, and a kvctl listing of the pool. All runs set NCCL_P2P_DISABLE=1 because PCIe peer-to-peer between these two GPUs is broken, nvidia-smi topo reports SYS and a minimal default-config NCCL allreduce hangs until its 90 s timeout, while the same probe completes over the SHM transport. NCCL stays the stock 2.29.7 library with two rank processes on two physical GPUs, only the transport moves. Disabling vLLM's custom allreduce by its serve flag follows from the same root cause, that path is raw CUDA IPC peer access outside the NCCL switch. Both settings are limitations of this box rather than choices.

TP=2 ran Qwen2.5-7B-Instruct and PP=2 ran Qwen2.5-0.5B-Instruct. An alloc-path geometry constraint forces the split, covered in the last observation below.

| Check | Result |
|-|-|
| Normal shutdown, one and two API frontends, both topologies | 4 cases passed, completed teardown removed the segment, group exit clean |
| Rank/stage SIGKILL during startup, after segment creation and before READY | Both topologies passed, the group self-exited, the reap audit was clean and the same-name restart worked |
| Rank/stage SIGTERM and SIGKILL mid-generation | 4 cases passed with full cleanup; the PP SIGTERM case took an external group SIGTERM after 182 s, detail below |
| SIGKILL one rank mid-teardown, group SIGTERM then KILL at +0.5 s | Both topologies passed |
| Delayed last worker, SIGSTOP through group SIGTERM with shutdown-timeout 10, SIGCONT at +15 s | Both topologies passed, orphaned-worker detail below |
| SIGTERM one of two API frontends while the engine lives | Segment contract held in both topologies; vLLM tears down the whole serve on its own, detail below |
| Same-name IPC restart after every scenario | 16 of 16 came up and served the four fixed prompts at temperature 0 with ignore_eos; the matrix's 104 counted completions, restarts plus primary checks, all read 64 completion tokens, 6656 total |
| Leftover and error scan | Every post-case audit found the case's segment absent, the serve group's process table empty and idle GPUs; the CUDA and NCCL scan of every restart log came back empty, and fault-case primary logs contain only the injected fault's own fallout; the one leftover class is the worker socket dirs, detail below |

Ownership mapping came out the same in both topologies. One segment exists per deployment under the explicit name. EngineCore creates it and keeps a descriptor, the launching supervisor process, the MPClient owner side, keeps a second descriptor, and TP ranks or PP stages touch it only through the native allocator, rank 0 on GPU 0 and rank 1 on GPU 1. Snapshots cover segment creation, serving, a teardown sampler at 0.3 s resolution and post-exit on the first case of each topology, plus fault time in every fault case, and each injected signal hit a live pid whose rank or stage sits in the case record. The frontend-exit teardown shows the new gate end to end. While EngineCore lingered, the supervisor-side wrapper logged at patches.py 1185 that it kept the segment because it could not yet confirm engine exit, the process manager force-killed EngineCore three seconds later, the wrapper then removed the segment under its retained identity within about a second of the confirmed exit, and the same name restarted onto a fresh inode.

vLLM 0.28 shuts the whole deployment down when one API frontend exits. SIGTERM to the second API server made the supervisor raise a RuntimeError for the exited frontend process and stop everything, including the remaining frontend, within a second, in both topologies. So the serving-path-after-frontend-exit check from the test plan is not reachable on stock 0.28 multi-frontend serve. Instead I verified that the inode is identical before and after the fault and that the full teardown removes the segment once.

Your A800 PP stage SIGTERM timeout shows up here too, with kvcached present. vLLM never noticed the TERMed stage-1 worker mid-generation, the deployment idled for 182 s until my external group SIGTERM, teardown then completed in about a second and removed the segment, and the restart served 4x64. TP=2 self-exits on the same fault and SIGKILL self-exits on both topologies, so the gap is PP and SIGTERM specific, consistent with tracking it on the vLLM side. Related, in the delayed-last cases vLLM's process manager gives up on the SIGSTOPped worker at the shutdown timeout, and the supervisor completes cleanup and removes the original segment while the stopped worker is still alive. No harm follows from that ordering, the owners snapshots show that only EngineCore and the supervisor open the segment and the workers go through their sockets, the identity check rejects any file created later under the old path, and the name is free for the restart, which came up on a fresh inode both times. As soon as SIGCONT delivers the queued SIGTERM, the stopped worker exits.

One leftover class does show up, the per-deployment worker socket directories. A process-group SIGTERM ends the workers directly, before their atexit cleanup runs, so each case's final teardown left its /tmp/kvcached-tp-(name)-(hash) directory with the last sockets inside, 16 directories for 32 serves since the path is per name. Removal itself works on an executor-managed worker exit, the frontend-exit teardowns logged the listener stop and removed the socket files. Each same-name restart also came up over its stale directory without complaint, the new listeners rebind onto the old paths. Worth a look at whether the #480 dir removal should also run from an engine-side shutdown path, and kvctl delete already covers manual cleanup.

The alloc path constrains which models can run TP=2 on 0.28 at all. Every 0.28 CUDA attention backend returns get_kv_cache_shape as (num_blocks, num_kv_heads, block_size, 2 * head_size) with K and V packed into the last dim, while alloc_kv_cache in integration/vllm/interfaces.py accepts only the older layouts with a literal K/V axis of size 2 in the first or second position. Only a per-rank KV head count of 2 satisfies both. Qwen2.5-0.5B has that geometry at TP=1 and PP=2, so the T4 tier and my PP runs serve correct output, and the accepted tensor is byte-identical to what vLLM asked for even though the axis kvcached matched is the head axis. At TP=2 the same model drops to one KV head per rank and every serve fails at KV allocation with "Unsupported kv cache shape: (434457, 1, 16, 128)". So TP=2 ran Qwen2.5-7B, whose four KV heads give two per rank. That constraint predates this PR, current main has the same alloc code, and I would track it as its own issue rather than grow this PR.

This pass does not claim combined TP=2/PP=2, which needs four GPUs, does not rerun headless, which your run covered, and cannot reach 0.28 geometries beyond two KV heads per rank for the reason above. All the per-case evidence sits on disk here and I can attach whatever helps, the full results table included.

shipiyouniao (approved):
Thanks for running the two-GPU matrix and spelling out the limitations. This closes the remaining topology check I requested for the shared-segment cleanup. Restoring my approval at `331ee5c`.

I rechecked the merge with #480: listeners drain before pools stop, then the allocator is released; an incomplete step keeps the state available for retry. The Linux lifecycle, socket-cleanup and combined shutdown regressions passed again (154 tests), all 11 CI checks pass on this head, and the final independent code review found no actionable issue. Your TP=2 and PP=2 fault checks and 16 same-name restarts provide the physical two-GPU evidence that the earlier T4 runs could not supply. The P2P/custom-allreduce settings, synchronous scheduling and lack of combined TP=2/PP=2 coverage remain explicit limits.

The separate PP SIGTERM hang is already covered by the pending vLLM PR https://github.com/vllm-project/vllm/pull/53597. Its out-of-band worker monitor handles the EOF/worker-teardown case we reproduced. That PR is still under review, so this is a pointer to the fix being worked on, not a claim that vLLM 0.28 already contains it. The case requiring an external group SIGTERM should remain recorded as such.

I opened #510 for the remaining worker socket-directory leak after process-group termination. We can address that in a separate PR, with ownership, delayed-worker and same-name-restart tests, rather than expanding this shared-memory cleanup PR again. In particular, an engine-side fallback must not remove sockets belonging to a still-live worker or a replacement deployment.

PR #484, opened 2026-09-04
fix(controller): serialize model sleep and wake operations

### LINKED ISSUE #462: SleepManager has no per-model_name lock, so concurrent sleep/wakeup calls each fire the full upstream call sequence
`SleepManager.put_model_to_sleep()` and `wakeup_model()` (`controller/sleep_manager.py`) check whether a model is already sleeping/waking, then `await` the real upstream call, then write the tracking dict afterward. There is no lock anywhere in the class, so two concurrent calls for the same `model_name` both pass the membership check before either has written the dict, and both go on to execute the full upstream call sequence.

For an SGLang-backed model this means `wakeup_model()`'s recovery path calls `update_weights_from_disk` against the live engine twice for what should be a single logical wakeup.

Verified this session in a clean `python:3.11-slim` container against current HEAD (`e8e6406`), running the real `SleepManager` class with only the three network-boundary methods (`_call_vllm_sleep_api`, `_call_sglang_resume_api`, `_call_sglang_load_weights_api`) stubbed:

```python
await asyncio.gather(put_model_to_sleep(m), put_model_to_sleep(m))
# upstream /sleep calls = 2, expected 1

await asyncio.gather(wakeup_model(m), wakeup_model(m))
# resume calls = 2, load_weights calls = 2, expected 1 each
```

This is a no-GPU environment, so I could not test the real vLLM/SGLang servers receiving two concurrent calls, only that `SleepManager` itself issues them twice with the real control flow and only the network boundary mocked.

Related: #408 / PR #409 fix a narrower symptom of the same missing synchronization (the losing racer's local bookkeeping), but PR #409 does not add any locking and its own description says the duplicate upstream call itself is out of scope. This issue is about that underlying gap. Not offering a PR right now since this lane is already at its open-PR allowance on this repo (PR #409); happy to send one once there is room.

justavibedev:
## Summary
- serialize sleep and wake transitions with a lock per model name
- share the lock across both operations so state is rechecked after any in-flight transition
- add CPU-only regression tests proving concurrent callers issue one upstream request

Fixes #462.

## Testing
- `python -m pytest -q tests/test_sleep_manager.py -k concurrent` (2 passed)
- `python tools/check_test_classification.py`
- `ruff check controller/sleep_manager.py tests/test_sleep_manager.py`
- `isort --check-only controller/sleep_manager.py tests/test_sleep_manager.py`

The network-boundary methods are mocked; no GPU or live vLLM/SGLang server was used.


rishabhsinha17 (commented):
The lock design holds up against the race cluster this repo has on record, and I verified it rather than taking the diff at face value.

What I checked. Both public entry points route through the same per-model lock (same dict, same key), so sleep and wake serialize against each other and not just against themselves, which is what "share the lock across both operations" needs to mean here. The membership checks live inside the locked inner methods, so the recheck-after-transition claim holds. Every writer of `sleeping_models` is now under the lock; the only other touches on main are reads (frontend.py:452, router.py:145 via `is_model_sleeping`, `get_sleep_candidates`). The monitor loop and `handle_model_wakeup_on_request` call the public wrappers, and the inner methods never call back into a wrapper, so the non-reentrant asyncio.Lock cannot self-deadlock. `setdefault` has no await point, so lock creation cannot race on one loop. Ran tests/test_sleep_manager.py on a0faa18, 11 passed; then reverted controller/sleep_manager.py to main while keeping your tests, and both new tests fail there, the wakeup one through exactly the KeyError-swallowed-to-False path, so these are real regression tests. The file is already in tests/manifests/cpu.txt:17, so CI covers them.

For the record on linkage, since it spans two issues. #462 describes exactly this missing synchronization, both racers passing the membership check before the await and each firing the full upstream sequence, including double `update_weights_from_disk` for SGLang wakeups; this PR is the fix that issue asked for. #408 is the narrower symptom of the same cluster, the losing wake racer reaching `del self.sleeping_models[model_name]` second, raising KeyError, and getting converted to False by the blanket except at sleep_manager.py:213. Under this lock that interleaving cannot occur, because the loser re-checks membership before touching anything.

Two things to resolve before this merges.

1. It collides with open #409, which fixes the #408 symptom the other way, `pop(model_name, None)` so both racers return True, with a regression test asserting `results == [True, True]` for concurrent wakeups. Your test asserts `[True, False]` with one upstream call. Whichever lands second breaks the other's test, and the two PRs disagree about what the losing racer should report, so this needs an explicit reconciliation rather than a merge-order accident. I think serialization is the right base since one upstream call is the point of #462. @AmirF194 for visibility.

2. The loser's False is router-visible. router.py:149-157 drops the request (records failure, returns None) whenever `handle_model_wakeup_on_request` returns False. With the lock, every request arriving while a wake is in flight blocks on the lock, then returns False from the membership check even though the model is awake by the time it returns, so a burst of traffic to a sleeping model gets one winner and the rest dropped. Main behaves the same via the #408 KeyError, so this is not a regression, but since you are touching the seam anyway, `handle_model_wakeup_on_request` returning True when the model is no longer in `sleeping_models` after `wakeup_model` comes back False would fix the user-visible half of #408 and make #409 redundant instead of conflicting.

Minor, take or leave. asyncio_mode is "auto" (pyproject.toml:64-66) so the `@pytest.mark.asyncio` decorators are redundant and no other test in the file uses them. `setdefault` builds a throwaway `asyncio.Lock()` on every call once the entry exists. `_model_locks` entries are never dropped when `remove_vllm_model`/`remove_sglang_model` retire a model, which only matters if model names churn.

rishabhsinha17:
Converging the two PRs, for visibility. On #409 AmirF194 offered to withdraw that one in favor of this PR and asked which way to go, and I have recommended there that this PR is the base, with the point 2 refinement from my review folded in. That refinement is the piece that resolves the pair, handle_model_wakeup_on_request returning True once the model has left sleeping_models after wakeup_model comes back False under the lock. Both concurrent wake callers then get a successful request with one upstream sequence, which is what #409 was protecting, and #409 becomes unnecessary rather than conflicting. With that folded in, plus a concurrent test at handle_model_wakeup_on_request asserting both requests proceed and one upstream call, this lands #462's fix and the user-visible half of #408 in one change. AmirF194 may bring his regression expectations here as review input, which I suggested on the other thread.

AmirF194:
Following up from #409: handle_model_wakeup_on_request (sleep_manager.py:296) returns whatever wakeup_model gives it, and under this PR's lock the losing racer's _wakeup_model finds model_name already out of sleeping_models and returns False. router.py:149-155 treats that False as a real wakeup failure and drops the request, even though the winner already finished waking the model up.

Fix would be in handle_model_wakeup_on_request: after wakeup_model returns False, check whether model_name is still in sleeping_models. If it isn't, someone else woke it up while we waited on the lock, so return True instead of propagating the False.

rishabhsinha17 (approved):
Re-verified at 72f5016d. The refinement from the last round is in and resolves both blocking points. handle_model_wakeup_on_request returns True when wakeup_model comes back False but the model has left sleeping_models, and _wakeup_model keeps the model in sleeping_models on every failure path before the del, so that check separates the racer that lost the lock from a real upstream failure. The new test_request_wakeup_still_reports_upstream_failure pins the failure side.

Ran tests/test_sleep_manager.py at 72f5016d, 14 passed. Reverted controller/sleep_manager.py to main while keeping your tests, and all four concurrency tests fail there, the request-wakeup pair through the router-visible False this fixes. Since #479 touched this file after your base, I merged current main into your head locally as well, 15 passed including the query-param test, so the two compose.

With both concurrent requests now succeeding on one upstream sequence, #409 is redundant rather than conflicting, per the thread there. CI has never run on this branch, so it needs a maintainer to kick the workflows before merge.

PR #486, opened 2026-09-11
feat(controller): make vLLM sleep mode configurable

ai-jz:
## Summary

The controller currently uses vLLM's default `abort` policy when putting a model to sleep. Add an opt-in `wait` mode so deployments can let the engine drain running requests first. Builds on #479.

## Changes

- Add `sleep_manager.vllm_sleep_mode` and `vllm_sleep_timeout_seconds` to YAML, retaining the `abort` and 30-second defaults.
- Allow `frontend.py --vllm-sleep-mode wait` to override the YAML mode.
- Send the mode alongside the sleep level as query parameters. Both settings apply to all vLLM models in one controller.

`wait` requires vLLM ≥ 0.17.0 with a supported V1 multiprocessing engine; queued requests are not guaranteed to complete. The timeout includes draining and offload but does not cancel remote sleep. Setup requirements are documented in the controller README.

## Tests

`PYTHONPATH=$PWD python -m pytest tests/test_sleep_manager.py -q`: **20 passed** on macOS/Python 3.11. Coverage includes:

- Both modes and the sleep level reach the query string; the request body stays empty.
- YAML defaults and overrides apply to both options; CLI overrides missing, null, and explicit YAML configuration.
- Invalid startup settings are rejected, and a real HTTP deadline returns failure.

Applicable pre-commit hooks and the configured MyPy 3.9 target passed. CPU suite: 189 passed, with the same 8 macOS `/dev/shm` errors as base `923849c` (179 passed). No live vLLM/GPU validation.


ai-jz:
@RixinLiu, could you approve the CI runs when you have a chance? This follows #479; I’d also appreciate your review when you have time.

ai-jz:
Thanks for approving the CI runs! All checks are green now.

ai-jz:
@RixinLiu, could you review this PR at your convenience? Thanks!

rishabhsinha17 (approved):
First pass at 5447edd9, and I checked the vLLM side rather than taking the README's word. The /sleep route reads mode from the query string with an abort default and hands it to EngineClient.sleep, and the handler reads query_params directly without validating unknown parameters, so servers that predate the mode parameter ignore it. Ran tests/test_sleep_manager.py at head, 20 passed. Reverted the two controller files to main while keeping your tests, and 11 of the 20 fail there.

One question, not blocking. vLLM's PauseMode is Literal["abort", "wait", "keep"], and the config rejects keep, with test_invalid_sleep_config pinning that choice. If that is a deliberate scope cut, a line in the README or the docstring saying why would save the next person from re-deriving it. If not, allowing it is a one-line change.

For sequencing, this composes with #484 in either merge order, since the two touch disjoint regions of sleep_manager.py and of the test file.

ai-jz:
Thanks for the careful review, @rishabhsinha17! This is an intentional scope choice: add graceful draining with `wait` while preserving `abort`. We haven’t validated `keep` across sleep/wake through the controller. Added a short docstring explaining this in 8f09f84.

rishabhsinha17:
Thanks, that settles the keep question. Checked 8f09f84d, it touches only the SleepConfig docstring, no code change, so my approval at 5447edd9 carries to this head. One note, CI has not run on 8f09f84d, the 11 green checks are from 5447edd9, a fresh run needs workflow approval from someone with write. Marking ready for a maintainer pass.

RixinLiu:
LGTM

PR #491, opened 2026-09-15
Support vLLM 0.29 Model Runner V2

ai-jz:
## Summary

Support vLLM 0.29's default Model Runner V2 (MRV2) while retaining kvcached's earlier integration routes. The change adapts allocation, cache views and block metadata, preserves existing ownership/prefix/copy lifetimes, and adds focused regression coverage. It carries forward existing cache/state families; allocator/controller redesign and unrelated baseline fixes remain separate.

**Why now:** vLLM's [0.29 release notes](https://github.com/vllm-project/vllm/releases/tag/v0.29.0) deprecate Model Runner V1 and target its removal in **0.32**. At the [recent roughly two-week release cadence](https://github.com/vllm-project/vllm/releases), those three increments suggest about six weeks from 0.29, as a planning estimate rather than a committed deadline. Reviewing the migration now gives kvcached a path forward before the legacy runner is removed; this PR qualifies 0.29, not future releases.

H100 qualification covers representative GQA/MHA, sliding-window/heterogeneous attention, MLA, Mamba2/GDN and FP8 KV configurations. Historical evidence includes full GSM8K, TP2, physical release/regrowth and sustained traffic; details and revision boundaries are below.

## Review follow-up

**Current head:** `acb0cb3733b192e8726a0e00e4fefefc51f91224`, rebased onto `73fa57ce` to include [#497](https://github.com/ovg-project/kvcached/pull/497) (stable PyTorch ABI) and [#472](https://github.com/ovg-project/kvcached/pull/472) (legacy cross-layer binding). The three contribution patches remain equivalent; no new feature or integration fix was added, and attribution is preserved.

**Fresh validation on this head:** rebuilt the native extension; 288 CPU tests, 19 native-vLLM view/block tests, 19 GPU ABI/transaction/layout tests, C++ tests, pre-commit and MyPy 3.12 passed. A two-H100 Qwen2.5-0.5B-Instruct smoke completed all 25 requests; each rank released and regrew 336 MiB, with prefix reuse, CUDA Graph replay and exact before/after health-fixture outputs. No unexpected CUDA safety errors were observed. All 11 hosted CI checks passed on this head.

**Earlier review follow-ups, tested at `27b0d965`:** rebased onto `7553c9fa` with [#418](https://github.com/ovg-project/kvcached/pull/418) and [#452](https://github.com/ovg-project/kvcached/pull/452), using a rebuilt native extension. No additional integration fix was needed.

| Review concern | Result |
| --- | --- |
| [TP2 mapping rejection during active generation](https://github.com/ovg-project/kvcached/pull/491#issuecomment-5710859910) | Rank 1 failed before mutation; rank 0 mapped and rolled back. The allocator stayed `HEALTHY`; the same engine completed all 33 requests. |
| TP2 successful map with lost ACK | Rank 1 mapped, then its ACK was withheld beyond a five-second test timeout. The allocator entered `FAILED`; all seven affected streams reported error 500 and `/health` returned 503. An explicit shutdown/fresh restart completed all 21 recovery requests. The failed pool was not reused. |
| [7B numerical differences](https://github.com/ovg-project/kvcached/pull/491#issuecomment-5711532838) | All **16 GSM8K prompts** matched exactly between pure vLLM 0.29 and vLLM+kvcached, with APC on and separately with APC off, including repeat verification. Tokens, text and returned selected-token/top-five logprobs were identical; maximum score difference **0.0**. |

TP2 used Qwen2.5-0.5B-Instruct on two H100s with MRV2, APC and default CUDA Graphs. After rejection and after restart, each GPU released/regrew **336 MiB**, with prefix reuse, subsequent graph replay and exact health-fixture outputs. Both injections were confirmed in rank-level traces during active generation; no unexpected CUDA safety error was observed. Recovery from the lost ACK required an explicit restart.

The numerical comparison used Qwen2.5-7B-Instruct, H100/TP1, BF16/FLASH_ATTN/BLNHC, default MRV2/graphs, **concurrency=1 and temperature=0**. Outputs ended naturally. These used our fixtures, not the reviewer's original L40S workload. [#482](https://github.com/ovg-project/kvcached/pull/482) remains an independent shutdown-cleanup fix, with no dependency or required merge order.

[The reviewer independently retested `27b0d965` on L40S](https://github.com/ovg-project/kvcached/pull/491#issuecomment-5737100358): both 0.5B and 7B matched pure vLLM on all 16 prompts, and 7B outputs remained exact after shrink/regrow. This was a fresh prompt sample because the original files were lost. They reported no remaining blocker.

## Compatibility approach

The bounded adapter follows the project's version-aware precedent: [#113](https://github.com/ovg-project/kvcached/pull/113), [#305](https://github.com/ovg-project/kvcached/pull/305) and [#394](https://github.com/ovg-project/kvcached/pull/394). [RixinLiu confirmed this approach](https://github.com/ovg-project/kvcached/pull/491#issuecomment-5686877612).

`NativeBlockPoolMixin` is named for its responsibility; `ModelRunnerV2Patch` follows the upstream architecture name. `VLLM_MRV2_RANGE` selects `>=0.29.0,<0.30.0`; with kvcached enabled, 0.29 V1 and unqualified newer versions fail before allocator/executor initialization. Earlier routes and the existing lower bound remain. Validation targets 0.24.0 and 0.29.0, not every version selected by historical routing. Qualified full-attention, sliding-window, MLA and Mamba spec classes are accepted; specialized subclasses need separate qualification.

## A/B Testing of this PR

*Each quality pair compares pure vLLM 0.29 with the upgraded kvcached integration using the same pinned model, backend, layout and effective capacity. All runs use the full 1,319-example GSM8K task. These are measurements of the earlier GPU-tested candidate, preserved after the source simplification; they are not fresh measurements of this PR commit.*

| Configuration | Pure vLLM 0.29 | Upgraded integration | Interpretation |
| --- | ---: | ---: | --- |
| Qwen/Qwen2.5-0.5B-Instruct, BF16 GQA | 443/1319 | 456/1319 | Break-even |
| Qwen/Qwen2.5-0.5B-Instruct, per-token/head FP8 BL | 398/1319 | 413/1319 | Break-even |
| DeepSeek-V2-Lite, BF16 MLA BL | 511/1319 | 516/1319 | Break-even |

We interpret these aggregate quality scores as practically break-even for this upgrade. Statistical significance has not been established; this is not a claim of improvement or proven equivalence. One run per arm; no exact-output claim. Protocol: lm-eval 0.4.12, five-shot raw completions, flexible-extract scoring, temperature 0, concurrency 16, output limit 256, APC on and default CUDA graphs. MLA BL scores 509/1319 after physical recovery in the same worker, versus 516 at startup, with quota reduced from 1458 to 972 MiB.

Recorded native-writer and serving tests establish physical release/regrowth and live-data retention for MLA BL/LB and per-token/head FP8 BL. Dense APC-off performance normalizes each kvcached integration against its matching pure-vLLM version, then compares those ratios: adverse observed ratios are **0.9094 throughput, 1.0543 p95 TTFT and 1.1780 p95 TPOT**, within the agreed 0.90/1.20/1.20 thresholds. Evidence comprises three original-0.24 pairs, three historical upgraded-0.29 pairs and one fresh-0.29 bridge.

The throughput margin is only **0.94 percentage points**. This is an observed sensitivity bound across different GPU sessions, not a confidence interval. A separate APC-off trace did not exercise either removed shortcut, supporting bounded source applicability; it did not instrument every formal request. Earlier APC-on timing and generated scores remain tied to their measured revision. Removing the shortcuts restores upstream query/scanning costs and native query error exposure.

<details>
<summary>Representative model, layout and backend coverage</summary>

Text-only H100 80GB runs used default MRV2 and `FULL_AND_PIECEWISE` CUDA graphs. BL is contiguous `BLNHC`; LB is noncontiguous `LBNHC`. Unless stated, runs used TP1/PP1 and BF16. Page sizes are per physical pool.

| Model | Distinct cache/state geometry | Recorded coverage |
| --- | --- | --- |
| `Qwen/Qwen2.5-0.5B-Instruct` | GQA, `FLASH_ATTN`, 2 MiB pages | BL/LB serving and recovery; BL additionally TP2, repeated recovery and sustained traffic. |
| `HuggingFaceTB/SmolLM2-1.7B-Instruct` | MHA, `FLASH_ATTN`, 2 MiB pages | BL/LB serving/recovery; BL bounded co-residence. |
| `openai/gpt-oss-20b` | Sliding window 128, native MXFP4 weights, `FLASH_ATTN` | BL/LB final-channel output, window-boundary, checkpoint-eligible prefix and recovery checks. |
| `google/gemma-4-12B-it` | Heterogeneous head/page geometry, window 1024, `FLASH_ATTN`/FA4 | BL/LB APC off and LB APC on. BL/APC-on wording differences also occurred in native vLLM. BL and LB used different effective capacities: 20 and 4 GiB. |
| `Qwen/Qwen3.5-9B` | Gated DeltaNet (GDN), BF16 convolution/FP32 recurrent state, aligned 66 MiB pages | BL/LB serving/recovery; BL repeated recovery and sustained traffic. |
| `Zyphra/Zamba2-1.2B-instruct` | Mamba2, BF16 state tensors, aligned 10 MiB pages | BL/LB serving/recovery with the observed minimum shrink footprint. |
| `deepseek-ai/DeepSeek-V2-Lite` | MLA latent/RoPE storage, `FLASH_ATTN_MLA`, 18 MiB pages | BL/LB native-writer/remap and serving/recovery; BL full GSM8K before and after recovery. |
| Qwen GQA above, quantized KV | Scalar `fp8_e4m3`/`FLASH_ATTN`; `fp8_per_token_head`/`TRITON_ATTN` with inline FP32 scales | Both layouts serving/recovery; per-token/head BL also full GSM8K and exact native-writer/remap. |

These are selected configurations, not every model/backend combination or default-page-size compatibility. GDN/Mamba2/MLA allocation bundles were 528/60/486 MiB, using release native wheels. Per-token/head FP8 capacity was matched after rounding: 8,176 retained IDs including the null ID, rather than comparing raw configured block counts.

</details>

<details>
<summary>GPU qualification results beyond aggregate GSM8K</summary>

| Area | Executed evidence | What it establishes |
| --- | --- | --- |
| Two-GPU tensor parallelism | GQA: 29 candidate checks and 16 recovery requests; each rank released and regrew 336 MiB | Both workers used MRV2/default graphs; rank-local IPC mapping, HBM changes and graph replay observed. Earlier candidate, before the upstream IPC rebase. |
| Repeated recovery and sustained traffic | GQA and GDN each completed 20 physical cycles and 320 exact fixture outputs, then separate runs with at least 3,600 active seconds | 33,358 GQA and 7,306 GDN load requests, plus 26 warmups each, at concurrency 1/2/6 with streaming and nonstreaming. No errors, cancellations or fixture mismatches; settled pool accounting returned to baseline. 44 GDN requests took over five seconds. |
| Native cache preservation | MLA BL/LB each released/remapped 972 MiB; per-token/head FP8 BL released/remapped 96 MiB | Known native-written data and FP32 scales, live-data retention, stable virtual pointers and two writer-graph replays per case. These standalone writer tests are distinct from model-serving graph checks. |
| Serving after physical recovery | MLA BL/LB: 29 responses each; FP8 BL: 29 responses | MLA released 972 MiB/regrew 486 MiB; FP8 released 480 MiB/regrew 336 MiB. MLA uses 3 arithmetic + 26 workload checks; FP8 uses 11 content + 18 workload checks. Explicit reset and fitting-request recovery boundaries remain. |
| Lifecycle and bounded sharing | 29 pre-sleep + 29 post-wake candidate checks; separate two-engine sharing test with 38 exact outputs | Sleep/wake serving continuity and donor-release/recipient-use/donor-recovery under controlled steps. Sleep alone retained native KV; neither test proves saturated-HBM admission or higher serving density. |

These GPU results are from manual qualification, with GSM8K as the only benchmark dataset and no vision/VLM coverage. Native-vLLM view/hash tests remain integration tests, outside the hosted CPU manifest; dependency-free lifetime and startup guards run in CPU CI.

</details>

<details>
<summary>Historical performance details</summary>

“Pure vLLM” means the matching engine with kvcached disabled. The original and upgraded integrations both use kvcached. Automatic prefix caching (APC) is supported; APC-off isolates the no-prefix-reuse workload, while APC-on characterizes a repeated-prefix workload.

| Dense APC off | Integration/pure-engine throughput | Observed overhead |
| --- | ---: | ---: |
| Original kvcached + vLLM 0.24, three pairs | 0.9474–0.9613 | 3.9%–5.3% |
| Earlier upgraded candidate + vLLM 0.29, three pairs | 0.8742–0.8929 | 10.7%–12.6% |
| Same upgraded candidate, additional 0.29 pair in a later GPU session | 0.9035 | 9.6% |

The upgrade comparison divides the 0.29 integration/pure-engine ratio by the 0.24 ratio. It retains all adverse observations; the total overhead is not attributed entirely to an inherited issue. Workload: six fixed 3,072-input/128-output token cases, closed-loop concurrency six, Qwen2.5-0.5B-Instruct BF16/FLASH_ATTN, BL, block size 16, matched 1.5 GiB effective KV capacity and default graphs. Each arm has 1,620 formal requests after 870 conditioning requests; APC-off observations report zero cached tokens.

<details>
<summary>Historical APC-on characterization — three pairs per profile</summary>

Ratios below are upgraded candidate / pure vLLM 0.29. Higher throughput and lower latency are better. TTFT is time to first token; TPOT is time per subsequent output token. These measurements precede removal of the LRU optimization and are not current-head timing acceptance.

| Profile | Throughput ratio | p95 TTFT ratio | p95 TPOT ratio | Formal requests, all six arms |
| --- | ---: | ---: | ---: | ---: |
| Dense Qwen2.5-0.5B-Instruct | 0.9136–0.9390 | 1.0388–1.0700 | 1.0374–1.0846 | 10,620 |
| GDN Qwen3.5-9B | 0.9975–1.0018 | 0.9912–1.0075 | 0.9981–1.0042 | 2,052 |

Both series met their frozen throughput/latency and baseline-variability gates. They use the same six 3,072/128-token cases and concurrency six, fresh scored processes and identical conditioning within each profile. Dense reused 3,056 cached tokens per formal request at 1.5 GiB effective KV capacity; GDN reused 2,640 at 4.125 GiB, with block size 528 and aligned state pages. This is reuse-heavy traffic, not an open-loop SLO or competing-model capacity benchmark. Sampled resident HBM was 3,637/3,685 MiB for pure/integrated Dense and 24,305/24,833 MiB for GDN; these results do not establish a memory-saving claim.

</details>

</details>

## Remaining checks and limitations

The broader family, quality and timing results are historical. No fresh performance benchmark was run after #418; unchanged kernels and equivalent contribution patches do not establish unchanged end-to-end performance. The latest rebase received the bounded validation above, rather than a new full model-family or performance run.

Remaining boundaries: GDN APC-off timing is inconclusive; MLA early-context code retrieval and MLA LB full GSM8K are unestablished. Target PP2, cross-layer sharing, GPU INT8 and debug-build aligned pages are not qualified. Existing 0.24 PP2/Mamba2-BL limits remain. Stronger saturated-HBM pressure qualification did not reach its required headroom band, so it establishes neither admission recovery nor a regression. Recovery still uses the recorded drain/reset/fitting-request steps; transparent queue recovery is not claimed.

Rollback: revert this integration and use a previously qualified engine image. The 0.29 legacy-runner fallback is intentionally rejected with kvcached enabled.

<details>
<summary>Runtime identities and shipped regression commands</summary>

| Runtime | Engine commit | PyTorch |
| --- | --- | --- |
| vLLM 0.29.0 | `98dff2a81d747d1dba01a47f939f48c3526d4206` | `2.13.0+cu130` |
| vLLM 0.24.0 | `ee0da84ab9e04ac7610e28580af62c365e898389` | `2.11.0+cu130` |

The historical TP2 and GQA/GDN cycle/soak runs used the earlier candidate004 wheel (SHA-256 `f850471beb59da483b519e88dd6f289028ad13571dfaad18fb7ea234abb7e498`). The later performance/quality/storage candidate007 source archive has SHA-256 `b10ebc9abed0143855f11d535962b72a7e110fde6b4722ef21c77cd9b6ed60f4`. These identify historical artifacts, not commits or fresh checks of the rebased source.

The post-rebase TP1/TP2 run used the freshly built `da18b957` wheel, SHA-256 `014aacf0deb58d1c843d95120a8c680e1baf032eda0e0bb7d81e8349cdd36a42`.

The `27b0d965` follow-up uses wheel SHA-256 `50959893a8e3dd94486c143f17295bcb0e5ea4f3843ac304e21e133489380cd4`, with installed Python/native bytes checked against that wheel. Qwen2.5-7B-Instruct is pinned to `a09a35458c702b33eeacc393d103063234e8bc28`; Qwen2.5-0.5B-Instruct to `7ae557604adf67be50417f59c2c2f167def9a775`. Both use the vLLM 0.29 image digest `sha256:082ca6f035279109041ffd3fe0695cb568b29bc580b35c4f297a66a08b216c1b`. The fault test shortens the existing IPC deadline to five seconds and injects one rank-1 fault per scenario; it does not change the product default.

The `acb0cb37` rebase smoke uses rebuilt wheel SHA-256 `abaee5f0b50922ed94a87c1d864f5014ed102a279f38e414b436951e4f64e681`, the same pinned 0.5B model and 0.29 image above, and verifies all 31 installed Python/native files against the wheel. This is a fault-free serving/recovery smoke; the earlier fault injections were not repeated.

Build the native extension separately for each matching engine/PyTorch environment. On Linux/Python 3.12, with that environment and the repository's test dependencies installed:

```bash
ENABLE_KVCACHED=false KVCACHED_AUTOPATCH=0 bash tools/run_cpu_tests.sh -q
bash tests/cpp/run_tests.sh
pre-commit run --all-files

# Run separately in the vLLM 0.29 environment: older tests install engine stubs.
ENABLE_KVCACHED=false KVCACHED_AUTOPATCH=0 python -m pytest -q \
  tests/test_vllm_model_runner_v2.py tests/test_vllm_native_block_pool.py
```

These commands cover shipped regressions. The one-off GPU qualification drivers are not part of this patch; the commands do not reproduce the physical instrumentation or performance experiments.

</details>


ai-jz:
Agent-authored response:

Human input: keep this contribution focused on the vLLM upgrade and upgrade-induced regressions, and publish it as a Draft for early feedback; unrelated inherited issues remain outside this scope.

Local review for `7302cab`: correctness/simplification review completed, confirmed findings fixed and reviewed. The 212 CPU tests, 19 native-vLLM tests and local repository checks passed. GPU evidence retains its measured revisions in the description. Both hosted workflows currently report `action_required` with no job results; CI and maintainer review remain pending.

RixinLiu:
> Maintainer feedback requested: Does this bounded adapter fit the intended compatibility strategy? Please correct our reading of the history or point us to an existing minimum-version/deprecation policy. The existing lower bound is unchanged.

Yes, this fits our intended compatibility strategy.

RixinLiu:
LGTM, and i also tested **some** experiments on my A100 40GB GPU below:

- Dense correctness: Qwen2.5-0.5B passed three shrink/regrow cycles with prefix caching both off and on. Outputs matched native vLLM.
- Dense performance: Median throughput was 1.85% lower with 2 MiB pages and 1.12% lower with 4 MiB pages versus native vLLM. Setup is same with yours, except i use less requests (180 for warmup, 360 for formal)
- GDN correctness: Qwen3.5-9B, text-only with prefix caching enabled, passed three shrink/regrow cycles at both 4 MiB and 66 MiB. Outputs matched native vLLM.

About the throughput overhead part, i think the reason that my overhead number is smaller than yours may be gpu execution on H100 is faster than A100. But i think this should not block this pr, the overhead problem should be a different topic including comprehensive benchmark on different GPUs, and then we can consider optimization.

@rishabhsinha17 can you also test on your end, share your thoughts?

shipiyouniao:
@ai-jz I reviewed `7302cab` and reran the CPU manifest on Linux: 212 passed. I did not find a blocking code issue in the adapter, though I have not rerun the 0.29 GPU tests myself.

Could you add a small real-GPU fault-injection test before we finish review? The recorded TP2 release/regrowth tests exercise successful unmap/remap, but report zero mapping failures. They do not cover a partial or uncertain mapping result.

Using an already-qualified small model with TP2, MRV2, APC and the existing CUDA Graph settings, I suggest two separate cases during active requests:

1. Make one rank reject a map before changing its mapping, while allowing the other rank to proceed.
2. Let one rank complete the map, then drop or delay its success response past the IPC timeout.

Please confirm that each injection actually hits, and exercise subsequent prefix reuse/eviction and physical release/regrowth. The useful evidence would be the tested SHA, rank-level outcome, complete responses/token counts and output checks after clearing the fault, plus CUDA errors and worker/EngineCore restarts. HTTP 200 alone would miss truncated output.

This is a targeted check of the new runner's block/copy lifetimes against failure handling, not a request to rerun the whole model matrix or redesign IPC in this PR. If a case exposes an inherited limitation, please distinguish it from an MRV2 regression and document the recovery boundary. If these tests already exist, a link to the commands and logs is enough.

rishabhsinha17:
Ran your protocol on an L40S 48GB (Ada, driver 580.159.03) at head `7302cab`, vllm 0.29.0 with torch 2.13.0+cu130, the same runtime identity as the PR body. Its native 0.29 regression suite, the two new test files run with kvcached disabled, passed all 19 on this machine. MRV2 confirmed default-on for 0.29 on CUDA and engaged in both arms of every comparison, checked through the kvcached MRV2 persistent allocation log line and the patched `allocate_kv_cache` code path.

Dense correctness on Qwen2.5-0.5B matched native 16/16 greedy in both APC modes at a matched 1.5GiB KV capacity, 8192 blocks with 2MiB pages, kvcached on BLNHC while native kept its default LBNHC as a pure vLLM control. Elasticity was 6/6 clean cycles, three natural-drain with APC off and three forced with APC on through `VLLM_SERVER_DEV_MODE` plus `POST /reset_prefix_cache`, cutting the limit to 256MB and restoring, with byte-identical outputs after every cycle.

Throughput on the 0.5B at 2MiB pages came in at -6.30% median for kvcached, 1610.4 native vs 1508.9 kvcached out tok/s over 3 formal windows, same workload and counts as yours, six 3072/128 cases at closed-loop 6 with 180 warmup and 360 formal. Larger than your -1.85% on A100. Qwen2.5-7B-Instruct measured +0.06%, 156.49 native vs 156.59 kvcached, with shrink/regrow 3/3 clean. On Ada the overhead is small-model-bound, which fits your read that it belongs in a separate benchmarking effort and should not block this PR.

One question for @ai-jz. At 7B the greedy comparison matched 11/16, and the 5 divergences are tail tie-flips after a long shared prefix, coherent alternate continuations rather than garbage. A layout control isolated the cause. Native forced to BLNHC vs native LBNHC matched 16/16, while native BLNHC vs kvcached BLNHC diverged on the same 5 prompts. That points at kvcached allocation numerics rather than the layout choice, ordinary tie-break drift by the look of it, but is it expected?

Forcing V1 with kvcached enabled, `VLLM_USE_V2_MODEL_RUNNER=0`, dies fast at startup with `KVCachedConfigError` and does not leave a segment behind. One operational note, a graceful SIGTERM on 0.29 with this PR still leaves `/dev/shm/kvcached_vLLM_*` behind, the #477 behavior, and #482 covers that on main.

All of this is single-GPU TP1 so it does not speak to the TP2 fault injection cases above. Within that scope the PR holds up on Ada under your protocol.

ai-jz:
@shipiyouniao @rishabhsinha17 Thanks for the checks. I pushed the validated rebase as `27b0d965` (includes #418/#452) and updated the description with two focused follow-ups.

- **TP2 faults:** confirmed both injections during active requests. A known rejection rolled back and serving continued in the same engine. A lost ACK caused explicit request errors and unhealthy status; a fresh restart restored serving. Prefix reuse, CUDA Graph replay and 336 MiB release/regrowth per GPU passed after both recovery paths.
- **7B numerics:** on H100, Qwen2.5-7B-Instruct at concurrency=1 and temperature=0 matched pure vLLM exactly for all 16 GSM8K prompts, with APC on and separately off, including returned selected-token/top-five logprobs.

According to past experience with native vLLM, long greedy outputs can diverge later where the top two candidates have equal or nearly equal returned logprobs.

@rishabhsinha17, would you be willing to cross-check with your original prompts/settings? We're happy to run them here too if you share them.

shipiyouniao (approved):
Thanks for adding the focused tests. I checked the rebase at `27b0d965`; the adapter changes remain equivalent to the version I reviewed, and all 11 CI checks pass.

Your TP2 GPU follow-up addresses my earlier request: a confirmed mapping rejection recovers in the same engine, while a lost ACK fails closed and requires a fresh restart. The subsequent prefix reuse, graph replay and physical release/regrowth checks cover both recovery paths. I reviewed the reported GPU results; I have not independently rerun those tests on this head.

No remaining blockers from me for this scoped 0.29 MRV2 adapter. The original L40S output differences still need the matching prompts/settings to compare, and the historical performance results are not a fresh post-rebase benchmark.

@RixinLiu ready for your final review.

RixinLiu:
LGTM, after @rishabhsinha17 re-takes a look at the output divergence problem, i will merge this

RixinLiu (approved):

rishabhsinha17:
Reran the comparison at `27b0d965` on the same L40S 48GB class, vllm 0.29.0 with torch 2.13.0+cu130, matched KV capacity in both arms, 4096 blocks at block size 16 for the 7B with 2MiB pages, kvcached on BLNHC, native at its default LBNHC, FLASH_ATTN, BF16, APC off, greedy at temperature 0 with 128 max tokens, sequential single requests, and this time with logprobs=5 captured on every token in both arms.

My original prompt files were lost with that first pod, so I rebuilt a deterministic 16-prompt set from a fixed template pool with a recorded seed and generation script, matching the original run's settings and counts, and saved it durably this time. Happy to share the JSON if you want to run it on the H100.

At this head Qwen2.5-0.5B matched native 16/16 and Qwen2.5-7B-Instruct matched 16/16, token for token, so the 11/16 I measured at `7302cab` does not reproduce here, though the rebuilt set makes this a fresh sample rather than a strict rerun of the lost prompts. Captured logprobs back your near-tie explanation with numbers. Across the 2048 generated positions on the 7B, the largest difference between the two arms in any returned logprob, selected token or shared top-5, is about 4e-6, while 43 positions return equal logprobs for the top two candidates, at least one such position in every prompt. kvcached's allocation perturbs the scores at the 1e-6 scale. A flip needs a top-2 gap at or below that scale, and agreement everywhere else follows. My earlier divergences were tail flips after long shared prefixes, consistent with that mechanism, and the layout control had already isolated the perturbation to allocation rather than layout. I agree this is the known near-tie greedy behavior of native vLLM and not a correctness problem in this PR.

A shrink and regrow cycle at 7B on this head was also clean. I cut the segment limit to 256MB, restored it, and the recaptured outputs were byte-identical 16/16 against the same server's baseline. Allocation identity matches the earlier run, pools=28 page_bytes=32768 blocks=4096 layout=BLNHC, with the V2 runner engaged in both arms.

Nothing blocking from my side. @RixinLiu

ai-jz:
Thanks @rishabhsinha17 for the rerun and @shipiyouniao for the review.

Rebased onto `73fa57ce` in `acb0cb3`, including #497 and #472, with no additional integration changes. The rebuilt extension passed CPU/native checks and a two-H100 serving/shrink-regrow smoke; all 11 CI checks are green. The PR description now separates these fresh checks from the earlier qualification results.

@RixinLiu, both review follow-ups are addressed. Ready to merge from our side whenever convenient. Thanks!

shipiyouniao (commented):
Rechecked `acb0cb3`. Its source tree exactly matches a clean merge of the previously reviewed `27b0d965` with `73fa57ce`; there are no extra adapter changes hidden in the rebase. I also checked the stable-ABI call signatures from #497 and the legacy/MRV2 routing around #472. No new blockers found, and all 11 CI checks pass.

Thanks for rebuilding the extension and adding the fresh TP2 H100 smoke results. Those cover release/regrowth, prefix reuse, graph replay and output checks on this head. I reviewed the reported results, rather than rerunning them myself. The fault-injection evidence is still from `27b0d965`, and this rebase check is not a fresh performance qualification.

@RixinLiu my approval still stands for this rebase.

Issue #490, opened 2026-09-15
vLLM 0.28 compatibility


shipiyouniao:
This issue tracks **KVCached compatibility with vLLM 0.28**: adapter changes, remaining gaps, and the configurations we have actually validated. Project-wide priorities are tracked in #419; the support matrix is tracked in #425.

## Adapter PRs

Status checked on 2026-09-19. The four adapter PRs are still open; an approval or an earlier passing test run does not mean the stack has merged.

| Area | PR | Status | Dependency |
| --- | --- | --- | --- |
| Startup profiling and warmup bookkeeping | #492 | Approved review; CI passing; not merged | None within this stack |
| Packed attention KV storage and backend strides | #493 | Open; conflicts with current main, needs rebase | #492 |
| Model Runner V1 hybrid/Mamba page binding | #500 | Open; conflicts with current main, needs rebase; implementation by @CMQY | #493 |
| Model Runner V2, attention-only | #494 | Open; conflicts with current main, needs rebase | #493 |

There are two branches after the shared fixes:

- **V1 hybrid/Mamba:** #492 -> #493 -> #500.
- **V2 attention-only:** #492 -> #493 -> #494.

#500 does not depend on #494, and #494 does not provide V2 Mamba support. Each child will be rebased after its parent merges. The initial 0.28 hybrid work targets V1; V2 hybrid/Mamba remains a separate follow-up.

## Related Compatibility Work

| Area | Tracking | Boundary |
| --- | --- | --- |
| Padded KV pages and strides | #399 / #426, open | Coordinate with this existing PR. #493 explicitly rejects unsupported padded layouts; it does not implement this work. |
| Cross-layer KV sharing | #417 / #472, open | Code review found no confirmed new functional defect. Actual allocation/reshape entry-point checks passed, but model-output comparison with native vLLM and checked-in entry-point regression tests are still requested. Validate the combined path on 0.28; #494 currently rejects this configuration. |
| Model Runner V2 migration on 0.29 | #491, open; no longer draft | Approved reviews and passing CI, but not merged. Coordinate overlapping runner and block-lifecycle changes. Its 0.29 results do not qualify 0.28 or replace its acceptance tests. This is not an extra dependency of the V1 stack. |
| PyTorch stable ABI | #306 / #497, merged | The ABI target, default-layout regression and minimum-version source-build issue have been addressed. Build/load checks passed on PyTorch 2.10 and 2.13; see below. This is separate from model-level 0.28 acceptance. |
[...]

PR #492, opened 2026-09-15
fix(vllm): preserve profiling state for 0.28 warmup

shipiyouniao:
## Summary

Part of #490 (worker profiling contract).

vLLM 0.28 warmup reads `Worker.total_consumed` and distinguishes persistent allocations from transient activation headroom. Current main initializes `total_consumed` to weights only and charges all process-local Torch growth as peak activation memory. It also runs graph-memory profiling inside the normal memory-profiling context, even though graph profiling creates temporary KV caches and has its own memory accounting. This rebased change completes that bookkeeping rather than merely adding a missing field.

This change completes the normal profiling context before calling `profile_cudagraph_memory()`. Graph profiling still runs in inference mode, its estimate remains available for logging, and real CUDA failures still propagate. The estimate is not deducted from the logical KV budget.

For the newer profiling result, the existing process-local Torch charge is split into persistent growth and transient headroom. `total_consumed` receives weights plus persistent growth; `peak_activation_memory` receives the remainder. Their sum and the resulting KV budget are unchanged. Older profiling interfaces retain the existing activation charge. Explicit user KV budgets still delegate to vLLM.

The patch deliberately does not copy `profile_result.total_consumed`: that value includes whole-device free-memory deltas and can count allocations made by other processes sharing the GPU.

The execution order follows [vLLM 0.28 Worker.determine_available_memory](https://github.com/vllm-project/vllm/blob/2cf0a6915ce544dc493a0990f2ea38d81601128a/vllm/v1/worker/gpu_worker.py#L515); the distinction between consumed memory and transient headroom is defined in [memory_profiling](https://github.com/vllm-project/vllm/blob/2cf0a6915ce544dc493a0990f2ea38d81601128a/vllm/utils/mem_utils.py#L317). Only process-local Torch values are used by this patch, not the device-wide consumed-memory delta.

## Validation

Tested commit: `e84a69936da15c85aa281ebdb297cb96a9b5534d`.

- Full pre-commit checks, GPU-free C++ tests, and MyPy on Python 3.9-3.13 passed.
- CPU suite: 303 passed on each Python version from 3.9 through 3.13.
- Rebuilt and imported the native extension with PyTorch 2.13.0+cu129 and vLLM 0.28.0 on a Tesla T4.
- Three real-CUDA profiling tests passed using vLLM's actual `MemorySnapshot` and `memory_profiling`, a synthetic runner, and CUDA Graph capture/replay.

| Profiling probe | Result |
| --- | --- |
| 2 MiB weights, 4 MiB persistent growth, 16 MiB transient peak, 512 MiB budget | KV budget 490 MiB; consumed memory 6 MiB; activation headroom 16 MiB |
| Subsequent graph profiling with a 64 MiB estimate | Graph replay succeeded; estimate recorded without changing the KV budget |
| Oversized CUDA allocation during normal profiling | Actual CUDA OOM propagated; no available-capacity result published; subsequent CUDA operation succeeded |
| Oversized CUDA allocation during graph profiling | Actual CUDA OOM propagated; no available-capacity result published; subsequent CUDA operation succeeded |

## Scope

This fixes the profiling/warmup bookkeeping contract. It is not a complete vLLM 0.28 compatibility claim: packed KV layouts, the actual V2 runner hooks, startup-plan and multimodal reservations, and hybrid prefix-cache behavior remain separate items in #490. The fault tests above are single-GPU tests, not TP/MPS validation.


rishabhsinha17 (approved):
Reviewed at 631c26f against v0.28.0's worker and mem_utils, plus a control run of the new tests against main's patches.py. The failure the summary describes reproduces in the 0.28 source. compile_or_warm_up_model guards on peak_activation_memory at gpu_worker.py:750 and then reads self.total_consumed unguarded at :767. Our override never set that field, and on 0.28 the auto-budget path stops there with AttributeError. Moving profile_cudagraph_memory after the context exit matches upstream 0.28's own ordering, and the before_torch_peak subtraction it replaces measured the same quantity, leaving pre-0.28 budgets where they were. The accounting split mirrors upstream's decomposition restricted to process-local terms, persistent clamps into the 0 to torch_peak_increase range so peak_activation cannot go negative, and the parametrized test pins the identity that requested minus total_consumed minus peak_activation equals available. Declining profile_result.total_consumed is right for this class, that field is a whole-device free-memory delta. Inference mode still covers the graph call, torch.inference_mode wraps the entire replacement function, and the failure test pins the no-swallow behavior. Against main's patches.py the new tests fail 9, and at this head all 24 pass, so they pin the change rather than the mock. One note for #490 rather than this PR, upstream 0.28 now includes ROCm in graph profiling via is_cuda_alike while _should_profile_cudagraph_memory still excludes it.

rishabhsinha17 (approved):
still LGTM after the force-push. it keeps the core fix, non_torch_memory stays 0 and nothing copies the device-wide total_consumed, and the new warmup split only runs on profile results that have transient_peak_headroom, which 0.28 adds, with the persistent and transient parts summing to the same budget as before. reran test_vllm_virtual_kv_capacity.py on cpu at e84a6993, 24/24 with the two new warmup and cudagraph-failure tests.

shipiyouniao:
Thanks for checking the updated head and rerunning the tests, Rishabh. `e84a699` is conflict-free with CI green. @RixinLiu, could you take a final look at this profiling fix? It is the first step in the 0.28 stack; after it lands, I can remove the inherited diff from #493, followed by #500. This PR alone is not the complete 0.28 adaptation.

RixinLiu (approved):

RixinLiu:
Merged, but i found this fix is for vllm 0.27+, not 0.28 @shipiyouniao I changed the commit message, and also the pr title

PR #493, opened 2026-09-15
feat(vllm): support 0.28 packed attention KV storage

shipiyouniao:
## Summary

Part of #490 (packed KV storage). Depends on #492: merge that PR first, then I will rebase this PR to leave only the packed-storage commit. The isolated storage diff is [e84a699...3235b73](https://github.com/shipiyouniao/kvcached/compare/e84a69936da15c85aa281ebdb297cb96a9b5534d...3235b73777105823e7b2782facc78c031410d709).

vLLM 0.28 attention backends expose a four-dimensional `(blocks, heads, tokens, K+V)` cache instead of the five-dimensional layout with a separate K/V axis. The old adapter rejects that shape during graph-profiler temporary-cache initialization.

This change allocates one combined K/V backing pool per layer and builds views using the backend's stride order. It handles NHD and HND, contiguous and non-contiguous layer storage, and differing logical/kernel block sizes. EngineCore and worker use the same physical block geometry. Unsupported layouts and padded page geometries are rejected before an incompatible view is used; padded-page support remains tracked by #426.

Capability discovery now includes HND for packed K/V, while legacy split K/V remains NHD-only. The rebase preserves the main-branch cross-layer-sharing and heterogeneous-view paths.

The native vLLM 0.28 Triton scale-cache implementation already respects tensor strides and storage offsets. The older five-dimensional KVCached override is therefore not installed on that version.

## Validation

Tested commit: `3235b73777105823e7b2782facc78c031410d709`.

- Full pre-commit, GPU-free C++ tests, and MyPy on Python 3.9-3.13 passed.
- CPU suite: 311 passed on each Python version from 3.9 through 3.13.
- Packed KV and heterogeneous-view regression tests with the compiled extension and vLLM 0.28 installed: 43 passed.
- Tesla T4, PyTorch 2.13.0+cu129, vLLM 0.28.0, Qwen2.5-0.5B-Instruct, FP16, TP=1, Triton attention, V1 explicitly selected. CUDA Graphs and prefix caching were enabled; asynchronous scheduling was disabled for this scoped layout comparison.

| Serving mode | Startup | Repeated functional prompts | Concurrent requests | Generated output tokens | Final health |
| --- | --- | --- | --- | --- | --- |
| Native vLLM | Passed | Stable | 16/16 | 1,024/1,024 | HTTP 200 |
| KVCached, non-contiguous | Passed | Matched native | 16/16 | 1,024/1,024 | HTTP 200 |
| KVCached, contiguous | Passed | Matched native | 16/16 | 1,024/1,024 | HTTP 200 |

The functional prompts covered a fixed reply, simple arithmetic, a factual question, and Chinese text. Full responses and token counts were checked, not just HTTP status.

All eight real CUDA VMM probes passed, covering both layer layouts, NHD/HND, and logical/kernel block-size ratios 1 and 2. Each probe maps 600 blocks across physical-page boundaries, writes and checks distinct K/V values for three layers, releases the pages, forces a physical-headroom rejection using an actual CUDA allocation, releases that pressure, and verifies all 600 blocks can be allocated and used again.

## Scope

This is a V1 packed-storage step, not complete vLLM 0.28 support. HND view correctness is tested independently; full-model serving here covers Triton NHD only. FlashAttention/FlashInfer serving, quantized/padded layouts, V2, hybrid prefix-cache lifecycle, and distributed failure ordering still need their own validation. These single-GPU results do not establish TP/MPS correctness or a performance comparison.


shipiyouniao:
@codex review

RixinLiu:
Added FlashAttention 2 coverage on A100. On vLLM 0.28 at 5b74b3a, both KVCached layouts matched native on all 48 complete outputs per layout. Setup: Qwen2.5-0.5B-Instruct BF16, CUDA graphs and APC enabled.

I also reproduced a startup failure on vLLM 0.27: the backend returns packed KV, but the `>=0.28.0` gate selects split geometry. Fixed in `054179b` by moving the boundary to 0.26 and updating the version tests. After the fix, both layouts on 0.27 also matched native on all 48 complete outputs. The CPU suite passed 501 tests; the focused shape/packed-view suite passed 56.

PR #495, opened 2026-09-15
tools: add bounded Codex engine compatibility repair

shipiyouniao:
## Summary

Add a bounded Codex repair runner for the compatibility work in #490, complementing #423. The sync job updates the engine fork; this runner repairs KVCached integration code against the selected engine source.

- Start with operator-owned behavior probes, not just an interface scan.
- Give Codex the failing checks, upstream source and an exact file allowlist, with attempt and time limits.
- Re-run independent checks before exporting a candidate patch and changed files. Record failures without publishing anything.
- Reject changes to protected inputs, runtime dependencies, Git refs and the index. Treat infrastructure failures separately from compatibility failures.

## Scope

This PR adds the local runner, CPU regression tests and usage documentation. It does not change inference code or #423's workflow, and does not commit, push, open PRs or merge. Workflow integration should keep repair, validation and publishing in separate jobs, without giving the agent publishing credentials. The integrity checks are not a security sandbox.

## Validation

The final runner was exercised against the vLLM 0.28 profiling/warmup regression using a clean pre-fix candidate and separate, fixed validation inputs. The generated repair is a test artifact, not another implementation of #492 in this PR.

| Check | Result |
|---|---|
| Baseline behavior probes | 9 failed, 15 passed |
| After one Codex repair attempt | 24 passed |
| Independent T4 CUDA checks | 3 passed |
| Runner regression tests | 42 passed |
| Python 3.9-3.13 CPU suites and MyPy | Passed |
| Pre-commit and GPU-free C++ tests | Passed |

The T4 checks use native vLLM 0.28 memory profiling, actual CUDA allocation and graph capture/replay. They also force real allocation OOM in the forward and graph profiling phases, check that it propagates, and verify that subsequent CUDA arithmetic still works. No manual edits were made to the generated repair before those checks.

This validates the repair loop on a known regression, not unattended compatibility with arbitrary upstream changes. It is not a full serving benchmark or TP/PP, MPS or hybrid-cache certification; those checks still belong to each resulting compatibility PR.


shipiyouniao:
Merged current main and kept both sets of CPU tests in the manifest.

The review also caught two runner issues, fixed in `163b360`:

- On Windows, a failed `taskkill` could leave an unbounded wait. Termination is now bounded; if descendant termination cannot be confirmed, the controller stops instead of starting another repair attempt.
- A check named `agent` could overwrite the agent transcript. Check logs now have a separate namespace, with a regression test that verifies the transcript survives.

Validation on this head: full pre-commit and Python 3.9-3.13 CPU/MyPy matrix passed before pushing, with 523 CPU tests per version. The Windows runner suite passed all 45 tests. A dedicated Codex review found no remaining actionable issues in this diff. All 11 GitHub checks are green too.

PR #496, opened 2026-09-15
ci: automate bounded vLLM release compatibility checks

shipiyouniao:
## Summary

This adds the release-driven workflow after #495: detect a stable vLLM release, pin its source, run a bounded Codex repair on a CPU host, and validate the result on a separate GPU runner. Codex credentials stay off the GPU machine. A verified candidate can produce a PR for human review; nothing is merged automatically.

**Depends on #495.** The first commit is that PR's repair runner. The second commit adds this workflow. I will rebase after #495 merges. #423's engine-fork synchronization is complementary, not a prerequisite for testing a pinned official release. Tracked in #490.

## Workflow

- Poll for unseen stable vLLM releases and record claims/results in a dedicated tracker issue. Interrupted attempts require an explicit retry.
- Permit at most two CPU/GPU rounds. Feed a candidate-bound GPU failure back to Codex; infrastructure failures stop the attempt.
- Reconstruct the same candidate commit in each job. Keep the controller and checks separate from generated changes, and restrict changes to the vLLM adapter and its profiling tests.
- Run the full CPU suite, Python 3.9-3.13 MyPy, pre-commit and C++ tests before publishing. Recheck the target base, remote SHA, PR identity and target CI afterward. Never overwrite a changed branch or silently rebase an already tested candidate.
- Keep the workflow disabled until an operator configures a dedicated GPU runner, tracker and credentials.

## Validation

Tested revision: `3eb5a418153ace0466299a5981be0f008784bc50`. The complete applicable CI ran privately before the first push:

| Check | Result |
| --- | --- |
| Full CPU suite, Python 3.9 / 3.10 / 3.11 / 3.12 / 3.13 | 650 passed on each version |
| MyPy, Python 3.9-3.13 | Passed |
| All pre-commit hooks, including workflow lint | Passed |
| GPU-free C++ tests | Passed |
| Dedicated local code review | No actionable findings |

The real-device probe from this revision uses a tiny locally generated Llama on one T4 with vLLM 0.28.0 (CUDA 12.9 build). The adapter under test was the existing #494 head, `cbb81bde4ddbdec64bef3c0d75c61ff35f4747db`; this PR does not change the serving adapter. Native vLLM and KVCached each completed 16 requests and 256 output tokens with identical token IDs. The KVCached run verified an active adapter and a partially mapped virtual KV pool. An intentionally oversized CUDA allocation failed as expected, and a subsequent GPU operation succeeded. Both engines shut down cleanly.

A separate local replay exercised the CPU-to-GPU repair feedback with real Codex execution and T4 profiling probes. The first GPU run had two failures; one repair reduced this to one failure. The bounded attempt stopped without publishing a PR. This is evidence of the failure gate, not a claim that the agent solved that regression completely.

## Remaining Deployment Check

The local replay and direct GPU probe do not validate GitHub-hosted authentication, runner registration, image provisioning or a live scheduled Action. Those still need an operator-controlled end-to-end run before enabling the schedule. No credentials or runners have been registered by this change.

The tiny single-GPU probe does not certify TP/PP, MPS, hybrid/Mamba, multimodal models or multi-instance performance. Changes involving those paths still need their corresponding real-device acceptance tests.

Setup and trust boundaries are documented in `docs/VLLM_RELEASE_COMPAT.md`.


shipiyouniao:
Updated this branch with current main and the runner fixes from #495. The existing task profiles, prior-GPU-failure input, and credential filtering are preserved. This remains a draft and still depends on #495.

Validation on `8602371`:

- Full pre-commit and Python 3.9-3.13 CPU/MyPy matrix passed before pushing: 968 CPU tests per version.
- The Windows timeout and transcript regressions passed.
- On a T4, the updated controller passed the vLLM 0.28 V1 and V2 attention probes in both contiguous and non-contiguous layouts. Each case checked native/patched token equality, intentional CUDA allocation failure and recovery, and explicit shutdown.
- Dedicated Codex review completed without actionable regressions; all 11 GitHub checks are green.

The GPU replay used an already validated compatibility candidate with its source fingerprint checked before and after execution. This is validation of the updated controller, not a claim that current main supports every release profile. It does not cover hybrid/Mamba, TP/PP, MPS, or multi-instance qualification.

shipiyouniao:
I also used this workflow on an actual release bug, rather than only replaying an already working candidate.

An independent regression reproduced a capacity drop between scheduler preflight and `get_new_blocks()`. The early guard raised an ordinary `ValueError`, which bypassed allocation-miss handling. One local Codex repair attempt changed only the allowed implementation and test files, and passed the independent witness and focused regressions. I subsequently fixed three test-only MyPy assignments before final validation.

The final candidate was then tested on a remote T4 with vLLM 0.28 V1 and 0.29 MRV2. Each run caught all three injected post-preflight capacity drops, completed all 16 prompts and 256 output tokens, and matched the native control token for token. Separate CUDA allocation-failure/recovery checks passed too. The exact final candidate passed full pre-commit/C++ checks and 624 CPU tests plus MyPy on each Python version from 3.9 through 3.13. The repair is still local and unpublished.

This validates the local-controller/remote-GPU repair-and-regression path, with operator involvement. It does **not** claim release-triggered Actions, automatic publication, or an unattended deployment has been accepted. The serving checks are single-GPU, eager, tiny attention-only model checks; they do not cover hybrid/Mamba, TP/PP, MPS or full-model performance.

This branch is now at `739fc80fcb09cfd97cb723ca493920ee6e5a0c29`. The profiling fixture now measures peak changes across the context boundary, so it correctly distinguishes graph capture inside versus after profiling. Private CI passed with 969 CPU tests per Python version, dedicated review found no actionable issue, and all 11 GitHub checks passed. The PR remains draft and still depends on #495.

PR #497, opened 2026-09-16
Port csrc tensor path to PyTorch stable ABI

### LINKED ISSUE #306: [RFE] Port kvcached to libtorch stable ABI
There is an ongoing effort in the PyTorch community to port extension to Torch's stable ABI. May I suggest that you port the kvcached C++ extension to PyTorch's [libtorch stable ABI](https://docs.pytorch.org/docs/stable/notes/libtorch_stable_abi.html), too? This would allow shipping one wheel that works across PyTorch versions instead of rebuilding for each release.

kvcached currently has to be rebuilt for every PyTorch version because `torch/extension.h` and pybind11 are tied to the internal C++ ABI. The stable ABI avoids this. Flash Attention 3 already ships stable ABI wheels, and vLLM is working on the same https://github.com/vllm-project/vllm/issues/26946).

With the stable ABI, one wheel per platform covers all PyTorch versions, which means less CI, less publishing, and less testing. Users can switch between PyTorch versions (including nightlies) without rebuilding kvcached. The stable and unstable APIs can coexist, so the migration can happen without breaking anything.

## Implementation

Your C++ extension is small and only touches a few PyTorch APIs:

- `torch::Tensor`, `torch::Device`, `torch::Dtype`
- `torch::from_blob`

All of these have stable equivalents in PyTorch 2.10+. My proposal would bump the minimum supported Torch version to 2.10.0, though.

I have created a quick PoC and pushed it to my fork https://github.com/tiran/kvcached/commits/torch-stable-abi/ . The commit is 80% written by Claude and then edited by hand to address a few problems. The new code builds and imports fine, but I have not verified that it actually works. I can turn my branch into a PR if you like to play with it.

tiran:
Register the KV tensor ops via STABLE_TORCH_LIBRARY and switch the ftensor/allocator tensor path to torch::stable::Tensor/Device, keeping PageAllocator/InternalPage on pybind11. Drop the dead pybind11 dtype-cast .ipp.

The ops now live under torch.ops.kvcached.*, so the extension is renamed kvcached._C and a thin kvcached/vmm_ops.py wrapper re-exports them alongside the pybind classes. This mirrors PyTorch's own layout of a private compiled module (_C) plus a Python wrapper re-exporting torch.ops.*, and keeps the existing kvcached.vmm_ops.* API unchanged.

Add tests/test_stable_abi.py to guard the boundary: it scans the built _C extension's undefined symbols with nm and fails on any unstable at::/c10::/non-stable torch:: reference.

Fixes: #306


tiran:
I have fixed the MyPy issues

setup.py
```diff
@@ -64,6 +64,9 @@ def get_extensions():
         "-std=c++17",
         f"-D_GLIBCXX_USE_CXX11_ABI={int(cxx_abi)}",
         backend_define,
+        # Target the PyTorch 2.10 stable ABI for the KV tensor ops registered
+        # via STABLE_TORCH_LIBRARY in csrc/torch_bindings.cpp.
+        "-DTORCH_TARGET_VERSION=0x0210000000000000",
```
RixinLiu: Here we should use 020a to represent 2.10?

csrc/torch_bindings.cpp
```diff
@@ -193,24 +207,35 @@ page_allocator_group_indices_by_page(std::shared_ptr<PageAllocator> allocator,
 
 } // namespace kvcached
 
+// Register the KV tensor ops in the "kvcached" dispatcher namespace.
+STABLE_TORCH_LIBRARY(kvcached, m) {
+  m.def("init_kvcached(str dev_str, int page_size=0, bool "
+        "contiguous_layout=False) -> ()");
```
RixinLiu: The previous Python binding defaulted contiguous_layout to True. Could we preserve that default here for callers that omit this argument?

setup.py
shipiyouniao: [P1] Encode the intended 2.10 ABI target correctly

The minor-version byte is hexadecimal: 0x0210000000000000 selects 2.16, not 2.10. The intended value is 0x020a000000000000, as documented in the [PyTorch stable-ABI example](https://docs.pytorch.org/tutorials/advanced/cpp_custom_ops.html).

This has an observable effect when building against newer headers. My PyTorch 2.13 build succeeds, but nm -D -u shows imports of torch_exception_get_what and torch_exception_get_what_without_backtrace. Those APIs are gated to 2.13 in the headers and are absent from the 2.10 stable shim. The current symbol test still passes because it checks namespaces, not the minimum version of imported C APIs.

Please correct the target and add a build/load check against the minimum supported PyTorch version, so a binary built with newer headers does not silently require newer symbols.

csrc/torch_bindings.cpp
shipiyouniao: [P2] Preserve the existing Python default for contiguous_layout

The parent commit exposes py::arg("contiguous_layout") = true. This schema changes that public default to False, even though PageAllocator still defaults to True. Callers that omit the argument now get a different tensor layout and count.

Reproduced on the T4 with init_kvcached("cuda:0", 2 * 1024 \** 2) followed by create_kv_tensors(16 * 1024 \** 2, 2, "cuda:0", 2): the new binding returns two tensors with 8,388,608 elements each, instead of the single contiguous tensor required by the old default. Explicit True/False calls both pass the native mapping checks.

Please keep contiguous_layout=True in the schema and add a regression test that omits the argument.

shipiyouniao (changes requested):
I built 79370d198b9b1544fd0785f759e9dbc79e69fbdb on a Tesla T4 with PyTorch 2.13.0+cu129. The native build and both ABI tests passed, as did 58 existing layout/snapshot tests and 17 additional native checks covering explicit layouts, element widths, group IDs, repeated map/write/unmap/remap, and recovery after an invalid dtype argument.

The additional default-layout compatibility test failed: omitting contiguous_layout now returns two per-layer tensors where the previous Python binding returned one contiguous tensor. I also found that the ABI target constant selects 2.16 rather than the intended 2.10; the resulting binary imports newer symbols despite passing the namespace-based ABI test. Details are in the two inline comments.

Please address these compatibility regressions before merging. These checks are limited to a single GPU; I have not validated loading the same binary across PyTorch versions or multi-rank fault injection.

setup.py
tiran: You are absolutely right, my bad. I missed this bug, because I tested the extension with Torch 2.13 and 2.14. I should have tested it with 2.10, too.

I'll rewrite the code to create the `TORCH_TARGET_VERSION` define from a `tuple[int, int]`. That makes it easier to understand for humans, too. I'll also file a ticket with PyTorch. It should fail to compile with a user requests a newer stable ABI version.

csrc/torch_bindings.cpp
tiran: ack! My conversion tool got confused by the fact that some code paths default to false. For example `FTensorAllocator` has

```
  static void init(const std::string &dev_str, size_t page_size = 0,
                   bool contiguous_layout = false);
```

setup.py
tiran: Torch added a compile time guard a few weeks ago. The changes https://github.com/pytorch/pytorch/pull/193962 didn't make it into 2.14.0 release.

shipiyouniao (commented):
Thanks, both code fixes are confirmed on 70c2e12.

I rebuilt the extension with PyTorch 2.13.0 and the corrected 2.10 ABI target, then loaded the same binary under PyTorch 2.10.0 without rebuilding. Its SHA-256 matched in both environments, and the 2.10 process loaded its own libtorch libraries. Both runtimes passed 20 tests on a T4: the two ABI checks plus 18 native regression cases covering layouts, dtypes, group IDs, repeated map/write/readback/unmap, invalid-dtype error recovery, and the omitted contiguous_layout argument.

The 11 CI checks are also green, but they do not build the CUDA extension or exercise this minimum-runtime load check. Please include regression coverage for the two original findings in the repository: newer-header build/minimum-version load, and the omitted-argument layout default. The source fixes themselves no longer reproduce either issue in this validation.

This was single-GPU binding/ABI validation, not a TP/MPS serving or distributed fault-injection run.

tiran:
I found another problem during testing. Torch <2.13 does not support pybind11 headers in stable ABI. My changes break compilation on PyTorch 2.10, 2.11, and 2.12:

```
torch/include/pybind11/functional.h:151:2: error: #error "This file should not be included when either TORCH_STABLE_ONLY or TORCH_TARGET_VERSION is defined."
```

It wasn't a problem with my previous PR. Back then, kvcached did not use pybind11 bindings.

csrc/torch_bindings.cpp
RixinLiu, line 16: Could we use the same workaround as TorchCodec?

  #pragma push_macro("TORCH_TARGET_VERSION")
  #undef TORCH_TARGET_VERSION

  #include <pybind11/functional.h>
  #include <pybind11/pybind11.h>
  #include <pybind11/stl.h>

  #pragma pop_macro("TORCH_TARGET_VERSION")

  #include <torch/csrc/stable/library.h>
  #include <torch/csrc/stable/tensor.h>

  This restores the ABI target before including the Torch headers.

RixinLiu (changes requested):

csrc/torch_bindings.cpp
tiran: ah nice, this trick is safe and even simpler than my initial approach.

setup.py
tiran: The PR now includes a compile time check to ensure that `TORCH_TARGET_VERSION` is not higher than the current Torch version. It was easier to implement than a test and it catches build issues right away. Builds also fail when `TORCH_TARGET_VERSION` is not set.

csrc/torch_bindings.cpp
tiran: I have implemented your suggestion and included a note that the workaround can be remove in the future.

shipiyouniao (approved):
Thanks, the remaining fixes check out on `8c3b778`, and all 11 CI checks are green.

I rebuilt and tested this revision on a T4:
- Built with PyTorch 2.13: 110 ABI, layout, native fault-path and transaction tests passed.
- Loaded the same binary under PyTorch 2.10, without rebuilding and with matching SHA-256: the same 110 tests passed.
- Built separately against the 2.10 headers: compilation succeeded, and all four ABI/default-layout checks passed.

The ABI target and default-layout regressions I reported are fixed. The checks also cover the native transaction/error paths from #418; this is single-GPU validation, not a TP/MPS serving benchmark.

@RixinLiu, looks good from my side. Leaving the final merge to you.

RixinLiu (approved):

PR #505, opened 2026-09-21
build: define TORCH_STABLE_ONLY explicitly to enforce stable ABI

tiran:
Makes any at::/c10:: usage a compile error.

See: #497, #306


shipiyouniao (approved):
Reviewed and tested commit 7ba26a08f8467fafc34621d7316ccef3fc153d17 on a Tesla T4 with Python 3.12.

| Build PyTorch | Runtime PyTorch | Result |
| --- | --- | --- |
| 2.13.0+cu129 | 2.13.0+cu129 | Build/import succeeded; 18 GPU mapping lifecycle tests passed |
| 2.13.0+cu129 | 2.10.0+cu128 | Same binary imported successfully; 18 GPU tests passed |
| 2.10.0+cu128 | 2.10.0+cu128 | Build/import succeeded; 18 GPU tests passed |
| 2.10.0+cu128 | 2.10.0+cu128 and 2.13.0+cu129 | 15 transaction failure/rollback tests passed in each runtime |

No tests skipped. Both builds included TORCH_STABLE_ONLY. Coverage included mapped-tensor writes/readback, unmap/remap, invalid-input recovery, preservation of existing mappings on failure, and prepared-unmap abort/commit.

The macro save/undef/restore around the pybind11 includes looks correctly scoped. No blocking issues found. HIP was not tested; these results cover the CUDA native extension, not multi-rank/MPS serving.

@RixinLiu this looks ready from my review and CUDA validation. Please double-check and merge when appropriate.

tiran:
Thanks for the exhaustive tests. I don't have easy access to CUDA hardware for running these tests on my own.

This change should not affect the build at all. The new macro triggers a compile error if-and-only-if a contributor includes a header file from unstable API in the future.

rishabhsinha17 (approved):
LGTM. csrc greps clean of at:: and c10:: and every torch include is already torch/csrc/stable or torch/headeronly, so the new define only enforces what #497 established, and the pybind11 push/undef/pop mirrors the TORCH_TARGET_VERSION workaround already in that include block. setup.py is the only build path that needs the flag. i cannot rebuild the CUDA extension on this machine, so build coverage is shipiyouniao's T4 matrix.

RixinLiu (approved):

